Issue #3, opened 2021-08-03
Use http.MaxBytesReader instead of io.LimitReader


akshayjshah:
We should implement the `MaxReadBytes` option with `http.MaxBytesReader` instead of the plain `io.LimitReader` - that's what it's for!

severb:
I can take a stab at this if you're not in a hurry to land it.

akshayjshah:
Sever! It's good to hear from you, though I'm surprised you discovered my side hack :)

I've got a bunch of local commits piled up that may obsolete this issue, so hold off a bit. If you're interested in helping out, though, I'll make more of an effort to create issues for potential fixes & improvements.

severb:
I'm looking for a side project and this caught my attention. I like the idea and I'd love to help out. I think I can put a few hours a week into it.

akshayjshah:
Turns out that this logic is now being handled further away from the HTTP layer, so this isn't a good idea. I think we'd also prefer to impose these limits on a per-message basis rather than for the whole stream, so we probably don't want to slap a limit on the whole request body.

PR #5, opened 2021-08-08
Test Handler's ReadMaxSize

severb:
Check that the `Hander` short-circuits when the request's payload exceeds `ReadMaxBytes`. These tests set the scene for #3. I'm not sure if the plan is to grow the more end-to-end suite in `rerpc_test`. If so, I can move the tests there.

I also added a new variable length field in `ping` to make it easier to test larger payloads—especially protobuf-marshaled ones.


akshayjshah:
This is fantastic - thanks, @severb! I'll take a more detailed look tonight.

akshayjshah:
I made some __very__ large changes to the internals of handlers and clients - sorry, @severb! I'll fix up these changes so they apply to the current state of the code and land them.

handler_test.go
```diff
@@ -0,0 +1,126 @@
...
+	const readMaxBytes = 32
+	ping := rerpc.NewHandler(
+		"internal.ping.v1test.PingService.Ping", // fully-qualified protobuf method
+		"internal.ping.v1test.PingService",      // fully-qualified protobuf service
+		"internal.ping.v1test",                  // fully-qualified protobuf package
+		rerpc.Func(func(ctx context.Context, req proto.Message) (proto.Message, error) {
+			return &pingpb.PingResponse{}, nil
+		}),
+		rerpc.ReadMaxBytes(readMaxBytes),
+	)
```
akshayjshah, line 34: Since we're testing through the HTTP layer, I think we can use the generated code instead of calling `NewHandler` directly.

handler_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+		t.Run("grpc", func(t *testing.T) {
+			padding := "padding                      "
+			probeBytes, err := proto.Marshal(&pingpb.PingRequest{Number: 42, Msg: padding})
+			assert.Nil(t, err, "marshal request")
+			assert.Equal(t, len(probeBytes), readMaxBytes+1, "probe size")
+			r, err := http.NewRequest(
+				http.MethodPost,
+				"rerpc.io/internal.ping.v1test.PingService/Ping",
+				bytes.NewReader(probeBytes),
```
akshayjshah, line 102: We should use the generated client here - this isn't actually a valid gRPC request. gRPC request bodies are (potentially) a series of messages, and each one has a 5-byte prefix. This request is missing the prefix, and there's no exported function to add one.

The test passes, but it shouldn't.

akshayjshah (commented):
@severb, this is awesome. These tests caught a handful of bugs I introduced late last night :)

You don't need to do anything more - I'll push a commit to get everything building and passing and then land the PR. Thanks so much for helping out!

akshayjshah:
I don't have push access to your branch, so I pulled these commits down, added a few on top, and pushed them to `main`.

Your commits are 9655eb144fcdcc35850897101afa23615eb79440, 02a0bce311054706c22952ae788b66f709af851b, and 19f4ea4dde1af46c6c7666726ce6115a2ab3acdb :tada:

handler_test.go
severb: This makes a lot of sense.

handler_test.go
severb: I saw the change in https://github.com/rerpc/rerpc/commit/b8e87af950a4f95cb02e8acdbbc48629c9f0aae6 — glad I caught that one. I completely forgot about gRPC's own framing.

Issue #8, opened 2021-08-12
Validate metadata keys and values more thoroughly


akshayjshah:
HTTP and gRPC restrict the characters allowable in headers. We should extend `IsReservedHeader` to validate more thoroughly (and add a corresponding check for the header value).

See https://github.com/grpc/grpc-go/issues/468 for details.

PR #14, opened 2021-08-15
Expand IsReservedHeader checks

severb:
Expand IsReservedHeader to include the following reserved header keys:

  * The empty key -- per RFC 7230, Section 3.2
  * Keys starting with ":" -- per RFC 7540, Section 8.1.2.1
  * Keys with characters outside of this this alphabet: a-z, 0-9, "-", "_",
    and "." -- per gRPC spec, Header-Name rule

NB: Since reRPC treats header keys as case insensitive and changes their
casing based on the transport, we extend the gRPC Header-Name rule to
include A-Z.

Fixes the keys part of #8

### DIFF
--- a/header.go
+++ b/header.go
@@ -36,7 +36,7 @@ func IsReservedHeader(key string) error {
 		return errors.New("empty header key")
 	}
 	if key[0] == ':' {
-		return fmt.Errorf("%q is a reserved HTTP2 proto-header", key)
+		return fmt.Errorf("%q is a reserved HTTP2 pseudo-header", key)
 	}
 	for _, c := range key {
 		if !(('a' <= c && c <= 'z') ||
--- a/header.go
+++ b/header.go
@@ -2,6 +2,7 @@ package rerpc
 
 import (
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/textproto"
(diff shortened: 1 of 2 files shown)


header.go
```diff
@@ -24,9 +25,28 @@ import (
...
+// reserved.
+//
+// Unreserved keys can only contain the following ASCII characters:
+// a-z, A-Z, 0-9, "-" (hyphen-minus), "_" (underscore), and "." (period).
 func IsReservedHeader(key string) error {
+	if key == "" {
+		return errors.New("empty header key")
+	}
+	if key[0] == ':' {
+		return fmt.Errorf("%q is a reserved HTTP2 proto-header", key)
```
akshayjshah, line 39: Should this be "...HTTP2 *pseudo*-header"?

header.go
```diff
@@ -24,9 +25,28 @@ import (
...
 //
 // Currently, the following keys are reserved: Accept, Accept-Encoding,
-// Accept-Post, Allow, Content-Encoding, Content-Type, Te, and Trailer. Keys
-// prefixed with "Grpc-", "Rerpc-", and "Twirp-" are also reserved.
+// Accept-Post, Allow, Content-Encoding, Content-Type, Te, and Trailer. Empty
+// keys, or keys prefixed with ":", "Grpc-", "Rerpc-", and "Twirp-" are also
+// reserved.
+//
+// Unreserved keys can only contain the following ASCII characters:
+// a-z, A-Z, 0-9, "-" (hyphen-minus), "_" (underscore), and "." (period).
```
akshayjshah, line 33: :heart: Love it.

akshayjshah (approved):
Great just as it is! I've been wondering about how we should handle trailers, @severb - maybe you've got some ideas.

gRPC uses trailers a lot, and they're exposed in the grpc-go APIs. They're definitely useful for servers to send some metadata part-way through a stream. In `net/http`, you can set trailers mid-stream by setting a *header* whose key starts with `Trailer: `. This is a little weird, but it's the Go API. Should we allow the same thing here, so reRPC behaves like `net/http`, or should we add a separate API?

@amckinney too in case you've got ideas and time to think about this right now :)

header.go
severb: Good catch!

PR #15, opened 2021-08-15
Test bad routes

severb:


rerpc_test.go
```diff
@@ -423,6 +475,19 @@ func TestServerProtoGRPC(t *testing.T) {
 			assert.Zero(t, rerr.Details(), "error details")
 		})
 	}
+	testBadRoute := func(t *testing.T, client pingpb.PingServiceClientReRPC) {
+		t.Run("bad_route", func(t *testing.T) {
+			req := &pingpb.PingRequest{}
+			res, err := client.Ping(context.Background(), req, rerpc.OverrideProtobufPackage("test.badroute"))
```
akshayjshah, line 481: Awesome, I hadn't thought of using `OverrideProtobufPackage` this way but it's a good idea!

rerpc_test.go
```diff
@@ -307,6 +332,32 @@ func TestHandlerTwirp(t *testing.T) {
...
+			expected := &twirp.Status{
+				Code:    "bad_route",
+				Message: "no handler for path /internal.ping.v1test.PingService/Foo",
+			}
+			got := &twirp.Status{}
+			contents, err := io.ReadAll(response.Body)
+			assert.Nil(t, err, "read response body")
+			assert.Nil(t, json.Unmarshal(contents, got), "unmarshal JSON")
+			assert.Equal(t, got, expected, "unmarshaled Twirp status")
+		})
```
akshayjshah, line 360: These two blocks look identical to me - can we pull them out into a shared function? I know the tests I wrote aren't great about this right now :/

akshayjshah (changes requested):
Very much appreciate the test improvements! If you don't have the time/energy to pull the duplicated code out into a function, I can do it - just let me know.

rerpc_test.go
severb: Good idea. I can factor out a helper to test Twirp statuses, I think it will simplify a few other tests too.

akshayjshah (approved):
Thanks for the cleanup!

PR #16, opened 2021-08-15
Add bad route fallback in README.md

severb:
I was playing around with a poorly configured client and was hitting an
endpoint for another package. The default response returned by
http.NotFound doesn't work well with gRPC/Twirp clients. I think it's
prudent to include the fallback in the example and let expert users
remove it in case they want to do things differently.

If this is a good idea, we should update the docs to match.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ func main() {
   ping := &PingServer{}
   mux := http.NewServeMux()
   mux.Handle(pingpb.NewPingHandlerReRPC(ping))
+  mux.Handle("/", rerpc.NewBadRouteHandler())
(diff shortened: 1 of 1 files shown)


severb:
I'm not 100% sure it's worth complicating the example unless we think users are going to use it as a starting point.

akshayjshah:
I wish nobody used these tiny readme examples as real starting points, but they definitely do - this change makes sense to me. The package-level example (in the GoDoc) includes the bad route handler, but I'm happy to merge documentation PRs too!

akshayjshah (approved):

PR #17, opened 2021-08-22
Explore call headers API

severb:
We're experimenting with two API flavors: one option-based and one
context-based.

Both pass a MutableHeader struct to the call, which seems better than
yarpc's one option per header approach because we have a straightforward
way to signal and handle errors if the key or value is invalid.

I find the option-based approach pretty nice, but one thing bothers me:
what would it mean to pass a rerpc.ResponseHeaders option to the
client?! It doesn't seem to have well-defined semantics, so we should
probably disallow it. We can do so by extending the sealed CallOption
interface with a marker method to statically differentiate between
options that affect calls vs. those that affect both calls and clients.

Alternatively, I've been toying with a pure context-based API. This
approach is a departure from mainstream implementations, but it solves
the problem described above. It also mirrors the handler API and
reinforces the idea that headers are similar a context, just some
request-scoped values that pass the API boundary.

Do you have any other ideas we should explore? And what do you think
about the two variants?


rerpc_test.go
```diff
@@ -760,3 +760,101 @@ func TestCallMetadataIntegration(t *testing.T) {
...
+		requestMeta := callerMD.Request()
+		v := requestMeta.Get("Test-Ping") + "bar"
+		requestMeta.Set("Test-Ping", v)
+		return next(ctx, msg)
+	})
+}
+
+func (i headerPingInterceptor) WrapStream(next rerpc.StreamFunc) rerpc.StreamFunc {
+	return next
+}
```
akshayjshah, line 791: Unless we anticipate more complexity here, this interceptor might be more concise if we use `UnaryInterceptorFunc`.

call.go
```diff
@@ -30,6 +30,22 @@ type CallOption interface {
 	applyToCall(*callCfg)
 }
 
+type requestHeadersOption struct{}
+
+func (requestHeadersOption) applyToCall(*callCfg) {}
+
+func RequestHeaders(MutableHeader) CallOption {
```
akshayjshah, line 37: I've been following `net/http`, which calls the `map[string][]string` type `Header` (note the singular). I think we should be consistent, so these options would be `RequestHeader` and `ResponseHeader`. I'm also fine with changing the other types (Mutable/Immutable/etc) to be plural.

akshayjshah (commented):
Seeing both options in use is really helpful - thanks for putting in the extra work, Sever. After looking at the two approaches, do you have a preference? (It sounds like you prefer the context-based approach, but I want to make sure.)

To me, the option-based approach is simpler and avoids action at a distance.

* It's _impossible_ to set headers far from the client constructor or call site. This seems good to me - you can set common headers when you construct the client and one-off headers at the call site, but you can't insert random headers anywhere you've got access to a context.
* It's easy to set client-scoped headers. Where would we do that with a context-based approach?
* There's less danger of mistakenly leaking headers intended for one downstream service to another.
* We don't need to introduce functions to remove all or some headers from a context, which I think we'd need if we went with a context-based approach. (Sort of like `WithoutMeta`.)
* It's clearer that these APIs apply only to client calls: if you're writing a handler, you don't use `ResponseHeaders` to set response headers.

I found the context-based header/trailer APIs in grpc-go and Twirp to be pretty confusing. That may just be my fault, though :)

I do like that the context-based API doesn't require any special handling of the `ResponseHeader` option. It's a little unpleasant, but I don't mind special-casing that option and documenting that it has no effect when passed to a client. (We could also add error return values to the client constructors, but that's a more intrusive change.)

severb:
Yeah, I think an options-based approach is the way to go. I hope to update the patch with the implementation this week.

akshayjshah:
Awesome. Thanks again for showing both approaches in practice :)

severb:
Sorry for the pause. I took another look at this and we pass client options directly into generated code. I couldn't think of a neat way to filter `ResponseHeader` from the option list without changing the generated code and exporting another symbol (maybe we have more options when using generics because we encapsulate more generated code).

If we have to export something, an alternative is to introduce a `ClientOption`  as a subset of `CallOption`. I updated the PR to exemplify the idea. Unfortunately, this comes with the extra mental tax on complexity, but it gives us a mechanism to statically solve the problem. I'm not sure I like `applyToClient` for the sealed method, but that's an easy fix. What do you think about this—is it too much for too little?

severb:
I forgot to mention, the generated code needs to change to match this:
```
diff --git a/internal/ping/v1test/ping_rerpc.pb.go b/internal/ping/v1test/ping_rerpc.pb.go
index f7b8cb4..be9feb1 100644
--- a/internal/ping/v1test/ping_rerpc.pb.go
+++ b/internal/ping/v1test/ping_rerpc.pb.go
@@ -34,7 +34,7 @@ type PingServiceClientReRPC interface {
 type pingServiceClientReRPC struct {
        doer    rerpc.Doer
        baseURL string
-       options []rerpc.CallOption
+       options []rerpc.ClientOption
 }

 // NewPingServiceClientReRPC constructs a client for the
@@ -43,7 +43,7 @@ type pingServiceClientReRPC struct {
 //
 // The URL supplied here should be the base URL for the gRPC server (e.g.,
 // https://api.acme.com or https://acme.com/grpc).
-func NewPingServiceClientReRPC(baseURL string, doer rerpc.Doer, opts ...rerpc.CallOption) PingServiceClientReRPC {
+func NewPingServiceClientReRPC(baseURL string, doer rerpc.Doer, opts ...rerpc.ClientOption) PingServiceClientReRPC {
        return &pingServiceClientReRPC{
                baseURL: strings.TrimRight(baseURL, "/"),
                doer:    doer,
```

akshayjshah:
Thanks, Sever! Excited to take a look in more detail - hopefully tonight, but it might take me a day or two.

PR #21, opened 2021-08-24
Add ErrorDetail interface

amckinney:
In the interest of making the API a little less oriented around `proto.Message`, this explores an `ErrorDetail` interface that can only ever be implemented by the `rerpc` library (via an un-exported method).

There are some obvious warts with this approach, including:

1. The `ErrorDetail` encoding option must be explicitly defined in the `rerpc` library (i.e. users cannot define their own pluggable error detail types).
2. More allocations are required to transform to and from gRPC error details by default.
3. It's less ergonomic for clients to retrieve error details because they must manually type-assert them as a `*anypb.Any`. This gets worse when we support multiple types of details (if ever).

However, this interface makes the `Error` API less dependent on `*anypb.Any` error details, which is a gRPC-specific implementation detail. I'm not yet convinced idea just yet, but I think it's worth exploring as we audit the API as a whole.

### DIFF
--- a/error.go
+++ b/error.go
@@ -35,13 +35,25 @@ import (
 type Error struct {
 	code    Code
 	err     error
-	details []*anypb.Any
+	details []interface{}
+}
+
+// ErrorDetail represents an error detail sent between reRPC servers.
+type ErrorDetail interface {
+	isDetail()
+}
+
+// NewProtoErrorDetail constructs an ErrorDetail from the given proto.Message.
+func NewProtoErrorDetail(detail proto.Message) ErrorDetail {
+	return &protoErrorDetail{
+		detail: detail,
+	}
 }
 
 // Wrap annotates any error with a status code and error details. If the code
 // is CodeOK, the returned error is nil. Otherwise, the returned error will be
 // an *Error.
-func Wrap(c Code, err error, details ...proto.Message) error {
+func Wrap(c Code, err error, details ...ErrorDetail) error {
 	if e := wrap(c, err); e != nil {
 		e.SetDetails(details...)
 		return e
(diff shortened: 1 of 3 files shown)


akshayjshah:
## It's good! 😍 

Not surprisingly, the code looks good :) As ABG would say, you computer well. If we need to make this change, the API you're proposing is about as nice as it gets without using `interface{}`. Of the concerns you raise, I think most are okay:

> The ErrorDetail encoding option must be explicitly defined in the rerpc library (i.e. users cannot define their own pluggable error detail types).

This is a feature! The concrete detail type is tied to the protocol, and we're not doing a YARPC and making the protocol itself pluggable.

> More allocations are required to transform to and from gRPC error details by default.

Eh. Not ideal, but not a deal-breaker either.

> It's less ergonomic for clients to retrieve error details because they must manually type-assert them as a *anypb.Any. This gets worse when we support multiple types of details (if ever).

Definitely true. We could mitigate this by adding

```go
func OnlyProtoDetails(deets []ErrorDetail) []*any.Any` {
  // filter down to just the proto details and unwrap them
}
```
and similar helpers for any other detail types we introduce later. To me, the bigger problem is that handlers get harder to write (more below).

## But...

I'm not sure what we're accomplishing here. We make error details _look_ like they're not tied to protobuf, but they _are_ - the gRPC spec doesn't permit any other encoding, and other gRPC clients won't know what to do with them. (FWIW, this concern is unrelated to whether or not we support pluggable encodings: even if the in-stream messages are Avro, flatbuffers, or whatever, error details must still be plain protobuf.)

Let's put that aside for the moment, though. Does this get us closer to supporting error details for other protocols? The first protocol to consider is Twirp, since handlers already support it. Say we added a Twirp-focused detail type: it'd probably be a `[2]string` representing a key-value pair, and we'd assemble the slice of details into a `map[string]string` before putting them on the wire. What would handler authors do with that? They'd need to model their errors as _both_ Twirp KV pairs _and_ gRPC `*any.Any`s to accommodate both types of client. If we added support for another protocol later on, they'd need to add a third representation of their error details. This seems pretty awful to me. It'd be much more user-friendly if we found a way to represent `[]*any.Any` in non-gRPC protocols, even if that process is lossy.

At the highest level, _this is a gRPC implementation_. I can understand wanting to support other encodings for the in-stream messages, but I don't see why we should shy away from the inescapable realities of the wire protocol. This change _might_ let us invent a protocol that's very similar to gRPC but allows other encodings for details, but I don't think it's worth sacrificing usability for that. (And if our experience with TChannel, gRPC, HTTP/1.1, and YARPC has taught us anything, it should be that this sort of premature abstraction almost never works out well.)

PR #22, opened 2021-08-24
Add a CodecProvider and refactor Stream

amckinney:
In a similar spirit to #21, this proposes a new `CodecProvider` type that could be used to support multiple encodings (not just `proto.Message`). The `CodecProvider` acts upon the `Content-Type` header value, and currently switches between Protobuf's JSON serialization (for Twirp), and the default Protobuf serialization (for gRPC). As a result of this, we can more generically represent the `Stream` interface so that it acts upon `interface{}` instead of `proto.Message`.

This is currently structured so that the `CodecProvider` is a concrete type, and must be instantiated with `rerpc.NewCodecProvider`. We might actually want to write this as an `interface` so that users can plug in their own encoding support, but they wouldn't actually be able to support a _new_ `Content-Type` in practice because `rerpc` acts specifically on the `Content-Type` values encoded by the following:

```go
// ReRPC's supported HTTP Content-Types. Servers decide whether to use the gRPC
// or Twirp protocol based on the request's Content-Type. See the protocol
// documentation at https://rerpc.github.io for more information.
const (
	TypeDefaultGRPC = "application/grpc"
	TypeProtoGRPC   = "application/grpc+proto"
	TypeProtoTwirp  = "application/protobuf"
	TypeJSON        = "application/json"
)
```

A pluggable interface would let users customize how each message is [un]marshled for each of the above `Content-Type` values, but if the user encodes something different than is what is expected between the client and server for any given `Content-Type`, then it's useless in practice.

There's a fair amount to unpack here, so let's discuss whether or not this is the right direction to take things.

### DIFF
--- a/bad_route.go
+++ b/bad_route.go
@@ -18,6 +18,7 @@ func NewBadRouteHandler(opts ...HandlerOption) []*Handler {
 	}
 	h := NewHandler(
 		StreamTypeUnary,
+		nil,        // The CodecProvider is never called for this handler.
 		"", "", "", // protobuf package, service, method names
 		func(ctx context.Context, sf StreamFunc) {
 			stream := sf(ctx)
--- a/call.go
+++ b/call.go
@@ -40,6 +40,7 @@ type CallOption interface {
 // internal/ping/v1test package.
 func NewCall(
 	ctx context.Context,
+	codecProvider *CodecProvider,
 	doer Doer,
 	stype StreamType,
 	baseURL, pkg, service, method string,
@@ -82,7 +83,7 @@ func NewCall(
 	reqHeader["Te"] = teTrailersSlice
 	ctx = NewCallContext(ctx, spec, reqHeader, make(http.Header))
 	sf := StreamFunc(func(ctx context.Context) Stream {
-		return newClientStream(ctx, doer, methodURL, spec.ReadMaxBytes, cfg.EnableGzipRequest)
+		return newClientStream(ctx, codecProvider, doer, methodURL, spec.ReadMaxBytes, cfg.EnableGzipRequest)
 	})
 	return ctx, sf
 }
--- a/cmd/protoc-gen-go-rerpc/rerpc.go
+++ b/cmd/protoc-gen-go-rerpc/rerpc.go
@@ -151,6 +151,7 @@ func clientSignature(g *protogen.GeneratedFile, cname string, method *protogen.M
 func clientImplementation(g *protogen.GeneratedFile, service *protogen.Service, name string) {
 	// Client struct.
 	g.P("type ", unexport(name), " struct {")
+	g.P("codecProvider *", rerpcPackage.Ident("CodecProvider"))
 	g.P("doer ", rerpcPackage.Ident("Doer"))
 	g.P("baseURL string")
 	g.P("options []", rerpcPackage.Ident("CallOption"))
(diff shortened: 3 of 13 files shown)


amckinney:
> A pluggable interface would let users customize how each message is [un]marshled for each of the above Content-Type values, but if the user encodes something different than is what is expected between the client and server for any given Content-Type, then it's useless in practice.

For a (potentially) future consideration, I stumbled upon some prior art in a dRPC's [Protocol](https://github.com/storj/drpc/blob/ebd70c186e6c2c5e18ace6d1db4f5696399b3102/drpchttp/options.go#L47) abstraction, which encapsulates the `read` and `write` logic specific to each `Content-Type` (rather than just the [un]marshaling logic). We might be able to do something similar here.

akshayjshah:
> For a (potentially) future consideration, I stumbled upon some prior art in a dRPC's Protocol abstraction, which encapsulates the read and write logic specific to each Content-Type (rather than just the [un]marshaling logic). We might be able to do something similar here.

I like dRPC a lot! As a wire protocol, it's nice - super fast and simple. They're a little cavalier about the cost of dropping HTTP, though - they make it sound like pure overhead, but using dRPC throws away the last decade of features in proxies and LBs. No Envoy, no service mesh, no tracing/metrics/etc from sidecars, and no L7 load balancing. That's great for their problem domain, but it doesn't seem like a good tradeoff for most projects.

> There's a fair amount to unpack here, so let's discuss whether or not this is the right direction to take things.

I like the idea of pluggable encodings, but I think that this approach ends up in an awkward middle ground. I'd prefer to do less or do quite a bit more.

## Do Less

One approach is to do less and remove a few layers from your prototype. In this world, I think we could drop the `CodecProvider` and instead expose

```go
func JSON(c Codec) Option { ... }
func Protobuf(c Codec) Option { ... }
```
We'd add fields to `callCfg` and `handlerCfg`, but `NewCall` and `NewHandler` wouldn't need to change. If the user doesn't supply one or both codec, we'd just use the implementations we've got today. Nice and simple! Of course, this doesn't let users bring new codecs - it just lets them choose new implementations for the two we've got.

## Do More

In this option, we extend the Codec interface to include two new methods:

```go
func TwirpSubtype() string // e.g., "protobuf"
func GRPCSubtype() string // e.g., "proto"
```
Users could then bring totally new codecs. I'd still get rid of the `CodecProvider` and add two `map[string]Codec`s to `callCfg` and `handlerCfg` (one for Twirp and one for gRPC). We'd need to special-case `application/grpc` to use whatever codec's registered for `application/grpc+proto`, but I'm pretty sure that's actually part of the spec (can't find the reference right now). If we're going to do this, I'd really like to actually use it for something - maybe we also ship a flatbuffer or capnp plugin in a separate module?

## Do Lots More

The other option is to do quite a bit more. In that world, I think we'd go for something closer to dRPC's `Protocol` and let users bring their own HTTP-based protocols. We'd need to figure out how our protocol interface works with compression (dRPC doesn't support compression AFAICT, so they dodge the issue) and options like `MaxReceiveBytes` - ideally compression, protocol, and options would be orthogonal. Something like this might make sense?

```
type Compressor interface {
  Name() string // e.g., "gzip" or "snappy"

  // method set should make pooling really easy
  GetReader(io.Reader) io.Reader
  PutReader(io.Reader)
  
  GetWriter(io.Writer) io.Writer
  PutWriter(io.Writer)
}

type Compressors struct { ... }
func (cs *Compressors) Get(name string) (Compressor, bool) { ... }

type Protocol {
  ContentType() string
  Stream(w http.ResponseWriter, r *http.Request, *Compressors) Stream
}
```

I have no idea how we'd wire options into the protocols without exporting the `callCfg` and `handlerCfg` structs. Maybe all options are protocol-specific? I'm also not sure what the generated clients should look like - do they accept protocols as options? If so, what happens if I pass a flatbuffer-based protocol to a protobuf-generated stub? I worry that this approach leads to a package structure like YARPC or dRPC: everything's nicely decomposed, but the packages are so small that it's difficult for users to figure out how to put it all together.

## Overall

I lean towards the "do less" option, of course - from what I can see, there's virtually no demand for other codecs. I suspect that won't meet your goals, though :)

PR #25, opened 2021-09-01
ajs/generics

akshayjshah:
- Make client & server streams generic
- Make streams and interceptors work on interface{}
- Simplify type params on generic streams


internal/crosstest/v1test/cross_rerpc.pb.go
```diff
@@ -27,9 +28,9 @@ const _ = rerpc.SupportsCodeGenV0 // requires reRPC v0.0.1 or later
 type CrossServiceClientReRPC interface {
 	Ping(ctx context.Context, req *PingRequest, opts ...rerpc.CallOption) (*PingResponse, error)
 	Fail(ctx context.Context, req *FailRequest, opts ...rerpc.CallOption) (*FailResponse, error)
-	Sum(ctx context.Context, opts ...rerpc.CallOption) *CrossServiceClientReRPC_Sum
-	CountUp(ctx context.Context, req *CountUpRequest, opts ...rerpc.CallOption) (*CrossServiceClientReRPC_CountUp, error)
-	CumSum(ctx context.Context, opts ...rerpc.CallOption) *CrossServiceClientReRPC_CumSum
+	Sum(ctx context.Context, opts ...rerpc.CallOption) *callstream.Client[SumRequest, SumResponse]
+	CountUp(ctx context.Context, req *CountUpRequest, opts ...rerpc.CallOption) (*callstream.Server[CountUpResponse], error)
+	CumSum(ctx context.Context, opts ...rerpc.CallOption) *callstream.Bidirectional[CumSumRequest, CumSumResponse]
 }
```
akshayjshah, line 34: @amckinney Recommend expanding this file to look at the diff in the generated code :)

internal/crosstest/cross_test.go
```diff
@@ -69,7 +70,10 @@ func (c crossServerReRPC) Fail(ctx context.Context, req *crosspb.FailRequest) (*
 	return nil, rerpc.Errorf(rerpc.CodeResourceExhausted, errMsg)
 }
 
-func (c crossServerReRPC) Sum(ctx context.Context, stream *crosspb.CrossServiceReRPC_Sum) error {
+func (c crossServerReRPC) Sum(
+	ctx context.Context,
+	stream *handlerstream.Client[crosspb.SumRequest, crosspb.SumResponse],
+) error {
```
akshayjshah, line 76: This is a good example of how user's code changes. On the client side, the stream type is nearly always inferred (it's usually returned from a client method, so the user doesn't type it out). Users *do* need to write the stream's type in their service implementations, though.

To me, this is a little wordier than the generated types, but also clearer - it's obvious that the different client streams are all the same type, and that they'll have the same method set. The names are also (maybe?) better than the weirdo generated names with underscores: `*crosspb.CrossServiceReRPC_Sum` vs `*handlerstream.Client[crosspb.SumRequest, crosspb.SumResponse]`.

akshayjshah:
Note that if you've installed tip as `gotip`, you'll need to run like so to get everything building:

```bash
make clean
GOEXPERIMENT=unified GO=gotip make test
```

akshayjshah:
One other thought: would the stream types be clearer if we made one package per RPC type, with server- and client-side implementations side-by-side? We'd have `clientstream.Call` (or `.Client`), `bidistream.Server` (or `.Handler`), etc.

PR #69, opened 2022-02-09
Make error-wrapping uniform & reliable

akshayjshah:
This PR begins by introducing `ErrorInterceptor`, a high-level interceptor
similar to `HeaderInterceptor`. Using this new type, users can more easily map
their custom Go error implementations to and from `*Error`. This lets handler
implementations and clients act as though these custom Go errors are
transparently sent over the network. See `error_interceptor_ext_test.go` for an
example of what this looks like.

Then, we use the underlying machinery from `ErrorInterceptor` to make our own
error wrapping more uniform. We can remove nearly all our special-casing of
`context.Canceled` and `context.DeadlineExceeded`, instead handling that in a
thin wrapper around our gRPC `Sender` and `Receiver` implementations. In the
same wrapper, we ensure that all user-visible errors are `*Error`s with
attached error codes.


amckinney (approved):
This is such a cool quality of life improvement - great thinking.

zepatrik:
Anyone looking for this, the commits got rewritten to https://github.com/connectrpc/connect-go/compare/0330d93...aa9c1de
Later, it was removed again in #120

emcfarlane:
Hey @zepatrik , this was before v1 was finalized. I think the reason for removal is that its easy for users to write custom error interceptors to suite their needs. Let me know, or can raise an issue if theres is some functionality you are missing here.

PR #201, opened 2022-04-18
Use CodeInternal when user-supplied Codecs can't marshal Status

### LINKED ISSUE #197: Custom codec status marshal are swallowed and cause invalid grpc responses
When returning an error from a custom codec, the error does not surface anywhere and and connect won't send any grpc-trailers required for a valid grpc response.

Repro case:
```go
package connect_test

import (
	"bytes"
	"context"
	"errors"
	"fmt"
	"io"
	"net/http"
	"net/http/httptest"
	"testing"

	"github.com/bufbuild/connect-go"
	"github.com/bufbuild/connect-go/internal/assert"
)

type bufferCodec struct {
	name string
}

var _ connect.Codec = (*bufferCodec)(nil)

func (b *bufferCodec) Name() string { return b.name }
func (b *bufferCodec) Marshal(src any) ([]byte, error) {
	buffer, ok := src.(*bytes.Buffer)
	if !ok {
		return nil, fmt.Errorf("marshal expected *bytes.Buffer, got %T", src)
	}
	return buffer.Bytes(), nil
}
func (b *bufferCodec) Unmarshal(src []byte, dst any) error {
	buffer, ok := dst.(*bytes.Buffer)
	if !ok {
		return fmt.Errorf("unmarshal expected *bytes.Buffer, got %T", src)
	}
	buffer.Reset()
	_, err := io.Copy(buffer, bytes.NewReader(src))
	return err
}

func TestBadCodec(t *testing.T) {
	// codec := &bufferCodec{name: "json"}
	codec := &bufferCodec{name: "proto"}

	connectHandler := connect.NewUnaryHandler(
		"pkg.Service/Method",
		func(ctx context.Context, r *connect.Request[bytes.Buffer]) (*connect.Response[bytes.Buffer], error) {
			return nil, connect.NewError(connect.CodeFailedPrecondition, errors.New(r.Msg.String()))
		},
		connect.WithCodec(codec),
	)
	server := httptest.NewTLSServer(
		http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
			connectHandler.ServeHTTP(w, r)
			t.Log("sent headers: ", w.Header())
		}),
	)
	defer server.Close()

	request := connect.NewRequest(bytes.NewBufferString("foo"))
	client, err := connect.NewClient[bytes.Buffer, bytes.Buffer](
		server.Client(),
		server.URL+"/pkg.Service/Method",
		connect.WithGRPC(),
		connect.WithCodec(codec),
	)
	assert.Nil(t, err)
	response, err := client.CallUnary(context.Background(), request)
	if err != nil {
		if connectErr := new(connect.Error); errors.As(err, &connectErr) {
			t.Log(connectErr.Error()) // prints "Unknown: EOF"
			t.FailNow()
		}
		t.Error("expected connect code")
	}
	assert.NotNil(t, response)
	t.Fail()
}
```

With the buffer codec named `proto`, connect will feed my codec its internal `proto.Message` type to encode/decode status messages for the `Grpc-Status-Details-Bin` trailer. As the codec doesn't implement this but returns an unexpected type error I expected that to show up somewhere but instead the client will get a `Unknown: EOF` error.

Digging in, it seems to be that codec errors for error status messages cause connect to not send any `Grpc-{message,status}` trailers. I think it is understandable that we don't propagate the codec error message to the client, but do think that connect needs to return an internal error status code here.

akshayjshah:
Fixes #197.

### DIFF
--- a/connect_ext_test.go
+++ b/connect_ext_test.go
@@ -29,6 +29,7 @@ import (
 	"github.com/bufbuild/connect-go/internal/assert"
 	"github.com/bufbuild/connect-go/internal/gen/connect/connect/ping/v1/pingv1connect"
 	pingv1 "github.com/bufbuild/connect-go/internal/gen/go/connect/ping/v1"
+	"google.golang.org/protobuf/proto"
 )
 
 const errorMessage = "oh no"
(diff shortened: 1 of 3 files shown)


doriable (approved):
Oh wow, this is a nice find, @robbertvanginkel :D
This fix seems quite reasonable to me :stamp:

PR #204, opened 2022-04-20
Do not return error from Client constructor

bufdev:
Per discussion. While returning an error from `NewClient` does make a lot of sense to do some preflight validation, it's painful to use in a lot of cases, and we can just effectively do the validation before the first call. All errors that come up in practice are things you find within seconds of using the `Client` for the first time, and you fix once, and never hit again. On balance, this seems like a good tradeoff.

I am not in love with the propagation of the `err` field.


mfridman (approved):

akshayjshah (approved):
Looks good to me!

PR #210, opened 2022-05-12
Fix server-side timeout handling

akshayjshah:
Now that we've defined the Connect protocol, we can simplify the protocol
interfaces: content-type negotiation is simpler, so the interface gets smaller.
This is also an easy moment to fix a bug in server-side timeout handling.

### DIFF
--- a/protocol.go
+++ b/protocol.go
@@ -159,20 +159,15 @@ func (r *errorTranslatingReceiver) Close() error {
 }
 
 func sortedAcceptPostValue(handlers []protocolHandler) string {
-	var max int
-	for _, handler := range handlers {
-		max += len(handler.ContentTypes())
-	}
-	seen := make(map[string]struct{}, max)
-	accept := make([]string, 0, max)
+	contentTypes := make(map[string]struct{})
 	for _, handler := range handlers {
 		for contentType := range handler.ContentTypes() {
-			if _, ok := seen[contentType]; ok {
-				continue
-			}
-			seen[contentType] = struct{}{}
-			accept = append(accept, contentType)
+			contentTypes[contentType] = struct{}{}
 		}
+	}
+	accept := make([]string, 0, len(contentTypes))
+	for ct := range contentTypes {
+		accept = append(accept, ct)
 	}
 	sort.Strings(accept)
 	return strings.Join(accept, ", ")
--- a/protocol.go
+++ b/protocol.go
@@ -22,6 +22,8 @@ import (
 	"strings"
 )
 
+const discardLimit = 1024 * 1024 * 4 // 4MiB
+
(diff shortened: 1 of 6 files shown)


protocol.go
```diff
@@ -155,3 +157,39 @@ func (r *errorTranslatingReceiver) Receive(msg any) error {
 func (r *errorTranslatingReceiver) Close() error {
 	return r.fromWire(r.Receiver.Close())
 }
+
+func sortedAcceptPostValue(handlers []protocolHandler) string {
+	var max int
```
bufdev, line 162: nit: should this be sum?

protocol.go
```diff
@@ -155,3 +157,39 @@ func (r *errorTranslatingReceiver) Receive(msg any) error {
...
+	for _, handler := range handlers {
+		max += len(handler.ContentTypes())
```
bufdev, line 164: or is this supposed to be `maxOf(max, len(...))`?

protocol.go
```diff
@@ -155,3 +157,39 @@ func (r *errorTranslatingReceiver) Receive(msg any) error {
...
+	}
+	seen := make(map[string]struct{}, max)
```
bufdev, line 166: why not just add all content types to `seen`, then convert to a slice, and sort the slice?

protocol.go
```diff
@@ -155,3 +157,39 @@ func (r *errorTranslatingReceiver) Receive(msg any) error {
...
+}
+
+func discard(reader io.Reader) error {
+	if lr, ok := reader.(*io.LimitedReader); ok {
+		_, err := io.Copy(io.Discard, lr)
+		return err
+	}
+	// We don't want to get stuck throwing data away forever, so limit how much
+	// we're willing to do here: at most, we'll copy 4 MiB.
+	lr := &io.LimitedReader{R: reader, N: 1024 * 1024 * 4}
```
bufdev, line 192: nit: i know this was pre-existing, but in a perfect would, this discard limit would be a const at the top of this file IMO

bufdev (approved):

protocol.go
akshayjshah: It's the maximum possible size of the union, which is the sum of the individual set sizes.

protocol.go
akshayjshah: I've never understood this impulse for consts that are only referenced once (why add more indirection?), but I'm in the minority. I'll change it before pushing.

protocol.go
akshayjshah: It's about the same amount of code, since there's no convenient set type with an `Elements() [T]` method, and it makes more passes over the data. Don't mind changing if you'd prefer it that way.

PR #211, opened 2022-05-17
Return an Unavailable error on failed requests

smaye81:
This returns a more intuitive `Unavailable` error when issuing a request to an unreachable host.


mfridman (approved):
lgtm. A code "Unavailable" makes more sense here instead of a code "Unknown". 

The nice property is the caller can catch this with a `CodeOf == CodeUnavailable` and assert for net errors.

akshayjshah (changes requested):
Can we add a test too?

akshayjshah (approved):
Actually, never mind! I'm moving all this code around anyways, let's land this ASAP.

PR #212, opened 2022-05-18
Refactor gRPC implementation

akshayjshah:
This is a _very_ long PR - sorry! It might be more digestible commit-by-commit.

Many of the commits are just moving deck chairs around. There are a few
substantive changes:

- The complicated bit of code that uses `net/http` and `io.Pipe` to let us make
  streaming HTTP requests is factored out of gRPC, so we can re-use it in
  Connect.
- The Sender and Receiver interfaces change a bit to make it impossible for
  clients to send trailers or handlers to receive them. (This matches the
  protocol specification for gRPC, gRPC-Web, and Connect.) This _really_
  clarifies protocol implementations and has almost no user-visible effect.
- All gRPC-related code gets moved to `protocol_grpc.go`, and identifiers
  include `grpc`. In a few places, we change error codes from Unknown to
  something more specific (often Internal).

All in, our gRPC implementation ends up being just 767 lines of Go (excluding
comments). Once this is in, we can also land the Connect protocol
implementation :)


akshayjshah:
Okay! We can always change things as we notice issues - it's mostly unexported code.

PR #219, opened 2022-05-19
fix context error got wrapped by CodeUnavailable

cyinma:
seems like there is a regression around the recent `CodeUnavailable` changes

### DIFF
--- a/connect_ext_test.go
+++ b/connect_ext_test.go
@@ -90,6 +90,14 @@ func TestServer(t *testing.T) {
 			)
 			assert.Equal(t, connect.CodeOf(err), connect.CodeInvalidArgument)
 		})
+		t.Run("ping_timout", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


smaye81 (approved):

akshayjshah:
Awesome! Thank you for catching this, Tommy :)

Can we use a smaller timeout in the test - perhaps a nanosecond (in practice that's 15us on OS X)? A millisecond is a fairly long time with local networking, and I don't want the test to be flaky.

cyinma:
> Awesome! Thank you for catching this, Tommy :)
> 
> Can we use a smaller timeout in the test - perhaps a nanosecond (in practice that's 15us on OS X)? A millisecond is a fairly long time with local networking, and I don't want the test to be flaky.

done :)

akshayjshah (approved):

PR #227, opened 2022-05-26
Improve client streaming ergonomics

akshayjshah:
Rather than having handlers implementing client streaming methods call a
method to send the response, make things more like unary: they should
just return `(response, error)`.


akshayjshah:
This came up while I was writing docs for streaming: it was hard to understand why implementing a client streaming handler wasn't a little more like unary.

doriable (approved):
I'm a fan of this, I think this is a more intuitive implementation/behaviour (`SendAndClose` basically didn't offer any actual abstraction).

bufdev (approved):

akshayjshah:
> behaviour

I suppose regression was inevitable once you returned home.

PR #229, opened 2022-05-27
Special-case errors where h2c is not configured and where connect.WithGRPC is not used

bufdev:
This is pretty hacky, but it works. A few notes:

- Adding a test for this is a little difficult I think since it requires grpc-go to be reproduced, but we can talk about it.
- We will need to update the docs link.
- I prefixed the error instead of suffixed so I could use `%w`, let me know your thoughts.

### DIFF
--- a/duplex_http_call.go
+++ b/duplex_http_call.go
@@ -236,7 +236,7 @@ func (d *duplexHTTPCall) makeRequest() {
 	response, err := d.httpClient.Do(d.request)
 	if err != nil {
 		err = wrapIfContextError(err)
-		err = wrapIfLikelyH2CNotConfiguredError(err)
+		err = wrapIfLikelyH2CNotConfiguredError(d.request, err)
 		err = wrapIfLikelyWithGRPCNotUsedError(err)
 		if _, ok := asError(err); !ok {
 			err = NewError(CodeUnavailable, err)
--- a/duplex_http_call.go
+++ b/duplex_http_call.go
@@ -236,6 +236,7 @@ func (d *duplexHTTPCall) makeRequest() {
 	response, err := d.httpClient.Do(d.request)
 	if err != nil {
 		err = wrapIfContextError(err)
+		err = wrapIfLikelyH2CNotConfiguredError(err)
 		if _, ok := asError(err); !ok {
 			err = NewError(CodeUnavailable, err)
 		}
--- a/duplex_http_call.go
+++ b/duplex_http_call.go
@@ -237,6 +237,7 @@ func (d *duplexHTTPCall) makeRequest() {
 	if err != nil {
 		err = wrapIfContextError(err)
 		err = wrapIfLikelyH2CNotConfiguredError(err)
+		err = wrapIfLikelyWithGRPCNotUsedError(err)
 		if _, ok := asError(err); !ok {
 			err = NewError(CodeUnavailable, err)
 		}
--- a/error.go
+++ b/error.go
@@ -218,8 +218,7 @@ func wrapIfLikelyH2CNotConfiguredError(err error) error {
 	if errString := err.Error(); strings.HasPrefix(errString, `Post "`) &&
 		(strings.Contains(errString, `net/http: HTTP/1.x transport connection broken: malformed HTTP response`) ||
 			strings.HasSuffix(errString, `write: broken pipe`)) {
-		return fmt.Errorf("You likely have not configured h2c, see %s: %w", commonErrorsURL, err)
-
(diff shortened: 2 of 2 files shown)


duplex_http_call.go
```diff
@@ -236,6 +236,8 @@ func (d *duplexHTTPCall) makeRequest() {
 	response, err := d.httpClient.Do(d.request)
 	if err != nil {
 		err = wrapIfContextError(err)
+		err = wrapIfLikelyH2CNotConfiguredError(err)
```
akshayjshah, line 239: Could we only do this check if the URL starts with "http://"? If it's HTTPS then we definitely don't have h2c problems.

error.go
```diff
@@ -194,3 +200,49 @@ func wrapIfContextError(err error) error {
...
+	if _, ok := asError(err); ok {
+		return err
+	}
+	// net/http code has been investigated and there is no typing of any of these errors
+	// they are all created with fmt.Errorf
+	// grpc-go returns the first error 2/3-3/4 of the time, and the second error 1/4-1/3 of the time
+	if errString := err.Error(); strings.HasPrefix(errString, `Post "`) &&
+		(strings.Contains(errString, `net/http: HTTP/1.x transport connection broken: malformed HTTP response`) ||
+			strings.HasSuffix(errString, `write: broken pipe`)) {
+		return fmt.Errorf("you likely have not configured h2c, see %s: %w", commonErrorsURL, err)
```
akshayjshah, line 222: Can we make this less sure-sounding? The broken pipe error in particular has many causes. Maybe "possible h2c config error, see %s: %w"?

error.go
```diff
@@ -194,3 +200,49 @@ func wrapIfContextError(err error) error {
...
+	if _, ok := asError(err); ok {
+		return err
+	}
+	// golang.org/x/net code has been investigated and there is no typing of this error
+	// it is created with fmt.Errorf
+	// http2/transport.go:573:	return nil, fmt.Errorf("http2: Transport: cannot retry err [%v] after Request.Body was written; define Request.GetBody to avoid this error", err)
+	if errString := err.Error(); strings.HasPrefix(errString, `Post "`) &&
+		strings.Contains(errString, `http2: Transport: cannot retry err`) &&
+		strings.HasSuffix(errString, `after Request.Body was written; define Request.GetBody to avoid this error`) {
+		return fmt.Errorf("you likely are talking to a gRPC server without using the connect.WithGRPC() client option, see %s: %w", commonErrorsURL, err)
```
akshayjshah, line 245: Maybe "possibly missing WithGRPC client config, see %s: %w"?

akshayjshah (approved):
Some nits, feel free to land when addressed. Really
Appreciate you doing this!

error.go
```diff
@@ -19,12 +19,15 @@ import (
...
 	"fmt"
 	"net/http"
+	"strings"
 
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/reflect/protoreflect"
 	"google.golang.org/protobuf/types/known/anypb"
 )
 
+const commonErrorsURL = "https://connect.build/docs/common-errors"
```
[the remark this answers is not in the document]
bufdev: Done

duplex_http_call.go
bufdev: Sent the request to the function and check it there, done

error.go
bufdev: Updated it along those lines, check it out and let me know what you think

error.go
bufdev: Updated it along those lines, check it out and let me know what you think

bufdev:
Updated - let me know what you think of the error wording now, will hold off on merging until you give the OK

error.go
```diff
@@ -194,3 +200,53 @@ func wrapIfContextError(err error) error {
...
+// This happens when running a gRPC-only server.
+// This is fragile and may break over time, and this should be considered a best-effort.
+func wrapIfLikelyH2CNotConfiguredError(request *http.Request, err error) error {
+	if err == nil {
+		return nil
+	}
+	if _, ok := asError(err); ok {
+		return err
+	}
+	if url := request.URL; url != nil && url.Scheme == "https" {
```
akshayjshah, line 216: Can we bail out if the scheme is anything other than "http"? As written, we're still returning our special h2c error strings for file:// and unix:// URLs.

akshayjshah (approved):
Error strings look good! One request to change == to !=.

error.go
bufdev: Done

PR #246, opened 2022-06-01
WithInterceptors: Improve ASCII diagram

abhinav:
I felt that the diagram demonstrating the flow of information across
server interceptors, the network, and client interceptors could be more
readable.

This makes the following changes to it:

- Use ":" as a vertical ellipsis for "A B ... Y Z"
- Turn the network entries into a single delimited layer that the
  requests pass through
- Curve the lines at edges with `'` instead of `|` and `-`
- Increase distance between left and right sides by two more spaces
- Add arrow end/start around "handler logic"

### DIFF
--- a/option.go
+++ b/option.go
@@ -177,30 +177,32 @@ func WithCompressMinBytes(min int) Option {
 //
 // Applied to client and handler, WithInterceptors(A, B, ..., Y, Z) produces:
 //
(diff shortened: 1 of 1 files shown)


akshayjshah (approved):
I love it 😻 Thanks, ABG!

PR #248, opened 2022-06-02
Fixed README examples; Set autogeneration of them using mdox tool.

bwplotka:
Hey!

Huge fan here, this project is amazing! 💪🏽  It has lots of features (3 protocols in one), so I think examples has to be clear. I found the existing ones obsolete so I recreated them and committed with tooling that autogenerates them in README once changed. 

`make test` also checks if the examples are buildable, which will keep them up-to-date! 

Hopefully that will help make this project more accessible, cheers. Keep the good work!

I hope you don't mind putting auto-formatter for README. It ensures consistency and e.g puts all on single line (all IDEs handle that just fine, so no point in trying to manually adjust width of text)

Signed-off-by: Bartlomiej Plotka <bwplotka@gmail.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,5 +4,8 @@ go 1.18
 
 require (
 	github.com/google/go-cmp v0.5.8
+	golang.org/x/net v0.0.0-20220531201128-c960675eff93
 	google.golang.org/protobuf v1.28.0
 )
+
+require golang.org/x/text v0.3.7 // indirect
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ lintfix: $(BIN)/golangci-lint $(BIN)/buf ## Automatically fix some lint errors
 	$(BIN)/buf format -w .
 
 .PHONY: generate
-generate: $(BIN)/buf $(BIN)/protoc-gen-go $(BIN)/protoc-gen-connect-go $(BIN)/license-header ## Regenerate code and licenses
+generate: $(BIN)/buf $(BIN)/protoc-gen-go $(BIN)/protoc-gen-connect-go $(BIN)/license-header $(BIN)/mdox ## Regenerate code, examples and licenses
 	rm -rf internal/gen
 	PATH=$(BIN) $(BIN)/buf generate
 	@# We want to operate on a list of modified and new files, excluding
@@ -67,6 +67,8 @@ generate: $(BIN)/buf $(BIN)/protoc-gen-go $(BIN)/protoc-gen-connect-go $(BIN)/li
 			--license-type apache \
 			--copyright-holder "Buf Technologies, Inc." \
 			--year-range "$(COPYRIGHT_YEARS)"
+	@# README auto-generates examples
+	$(BIN)/mdox fmt -l *.md
 
 .PHONY: upgrade
 upgrade: ## Upgrade dependencies
@@ -94,6 +96,10 @@ $(BIN)/license-header: Makefile
 $(BIN)/golangci-lint: Makefile
 	@mkdir -p $(@D)
 	GOBIN=$(abspath $(@D)) $(GO) install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.46.2
+
+$(BIN)/mdox: Makefile
+	@mkdir -p $(@D)
+	GOBIN=$(abspath $(@D)) $(GO) install github.com/bwplotka/mdox@v0.9.0
 
 $(BIN)/protoc-gen-go: Makefile go.mod
 	@mkdir -p $(@D)
--- a/README.md
+++ b/README.md
@@ -1,22 +1,10 @@
-Connect
-=======
+# Connect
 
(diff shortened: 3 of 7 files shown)


bwplotka:
Some early feedback: I think as a project, we might need to think about better `Insecure` / nonTLS options - mangling with tricky http2/http client server is quite confusing. E.g by using `http.DefaultClient` it does not set any TLS by default. Yet it because it switched (I think?) to HTTP 2, suddenly it tries HTTPS and fails. This inconsistency is the same on server side. Thus you need to manually create HTTP2  client/server handler with explicit TLS off. Just tricky (:

bufdev:
Thanks for the input! A few notes here:

- The formatter changes the README significantly - it puts badges on the same line in the file, it takes out the link redirects organized at the bottom of the file, and it takes out our line wrapping (which we do want to keep, most of us here use vim for Markdown) - so I don't think this is something we can consider in its current state, especially until it is stable.
- We could consider a PR updating the examples on its own, we did make some careful design decisions though - for example, on the insecure h2c, we felt it better to keep the example not using `newInsecureClient`, and instead we detect this condition as best we can (https://github.com/bufbuild/connect-go/pull/229) and follow up with documentation https://connect.build/docs/go/common-errors as this is generally a one-time setup issue.

go.mod
```diff
@@ -4,5 +4,8 @@ go 1.18
 
 require (
 	github.com/google/go-cmp v0.5.8
+	golang.org/x/net v0.0.0-20220531201128-c960675eff93
 	google.golang.org/protobuf v1.28.0
 )
+
+require golang.org/x/text v0.3.7 // indirect
```
akshayjshah, line 11: I'd really prefer to keep `x/net` and `x/text` out of `go.mod`.

akshayjshah (changes requested):
Thanks for the kind words, and for taking the time to open a PR 😍 We appreciate it!

I really like that you've included an example of using cURL in the README. I'd be happy to accept a PR with just that change.

Unfortunately, I prefer the README wrapping and indentation as it is. I'd also prefer to keep `x/net/http2` out of our dependency graph, so converting the code blocks to testable examples doesn't work. I do like the direction that you're taking your Markdown tool, though - I can see why it's useful!

Finally, I understand that h2c is complicated with the standard library. I agree that it'd be nice if we could have a simple `WithInsecure` option. I don't see a clean way for us to do this, though: the `http2.Server` passed to `h2c.NewHandler` and the `http2.Transport` used on the client have many options, and I'd like users to see all of them. If `connect-go` gets some traction, perhaps we can reopen this discussion with the Go team and add a bit of friendlier API to `x/net/http2`.

bwplotka:
Ack, thanks for considering this PR 👍🏽 

As the format goes, sounds like the blocker is **line wrapping** and **taking links out of the bottom**, anything else? (what indentation do you have in mind @akshayjshah?) (cc @saswatamcode). Sounds like an easy addition to https://github.com/bwplotka/mdox then. I really believe in automation here, the examples and docs can get out of control easily. 

As per committing code examples here, I had to use the same module (thus extra dep), otherwise, I would need to duplicate ping proto hidden by `internal` (unless there is some go mod replace trickery I could do to make it work).

Happy to manually improve examples then, but I am missing the point here @bufdev: 

> We could consider a PR updating the examples on its own, we did make some careful design decisions though - for example, on the insecure h2c, we felt it better to keep the example not using newInsecureClient, and instead we detect this condition as best we can (https://github.com/bufbuild/connect-go/pull/229) and follow up with documentation https://connect.build/docs/go/common-errors as this is generally a one-time setup issue.

Hm, I assume you mean to keep the example client secure by default for security best practices, but I feel putting non-working examples to main docs is rather confusing for users. It looks like a mistake not a design decision to put an insecure server and client expecting a secure connection. 🤔  Perhaps if we add BIG warning and info to docs in `newInsecureClient` that would make sense? Or try to put secure server example? WDYT @bufdev ?

> Finally, I understand that h2c is complicated with the standard library. I agree that it'd be nice if we could have a simple WithInsecure option. I don't see a clean way for us to do this, though: the http2.Server passed to h2c.NewHandler and the http2.Transport used on the client have many options, and I'd like users to see all of them. If connect-go gets some traction, perhaps we can reopen this discussion with the Go team and add a bit of friendlier API to x/net/http2.

Makes sense 👍🏽 thanks for the explanation. Perhaps some docs or simple helper in the common `connect` package could be helpful.

saswatamcode:
mdox has a `--soft-wraps` option which preserves the soft line breaks i.e any custom line wrapping. It is an option that we use for [prometheus-operator](https://github.com/prometheus-operator/prometheus-operator/blob/main/Makefile#L303) docs too! 
Preserving links at the bottom would be a nice feature to have! 🙂

bufdev:
Closing the loop here:

1. We've discussed internally and we don't think we can adopt mdox at this time - not just the formatting, but we have concerns about the i.e. `sed -n '15,69p'` added into the Markdwon.
2. On balance, want to keep the examples as-is using secure HTTP.

Closing this PR as I think any changes to examples will be significantly different than on here, but feel free to discuss.

PR #253, opened 2022-06-03
Stencil fewer functions

akshayjshah:
We're currently using two generic functions to unmarshal unary requests
from Receivers. Because the two functions must use structs as their type
parameters (and structs have different gcshapes), we're stenciling a new
version of each function for each unary RPC.

These functions aren't complex enough to warrant this binary bloat. It's
simple enough to inline the logic where we use it instead.

(I don't think we need to be paranoid about this category of problems,
though - the Go generics implementation will almost certainly get better
over the next few releases, so it's not worth contorting our code to
work around imperfections in 1.18.)

### DIFF
--- a/connect.go
+++ b/connect.go
@@ -225,28 +225,6 @@ type Spec struct {
 	IsClient   bool   // otherwise we're in a handler
 }
 
-// receiveUnaryRequest unmarshals a message from a Receiver, then envelopes
-// the message and attaches the Receiver's headers and RPC specification.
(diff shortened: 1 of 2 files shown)


bufdev (approved):
Seems fine to me - it might even be more readable

akshayjshah:
FYI @pkwarren @smaye81 - as you get more into `connect-go`, might be worth reading the three generics implementation proposals to get a sense of what's happening under the hood here.

PR #265, opened 2022-06-06
guard against panics in client_stream

pkwarren:
If a client fails initialization, it is expected that all calls on the
underlying streams short circuit and return the initialization error.
Additionally, calls to any other methods (i.e. RequestHeader) should
work without a panic.

Update ClientStreamForClient, ServerStreamForClient, and
BidiStreamForClient to be safe from panics when created with an
initialization error.

### DIFF
--- a/client_stream.go
+++ b/client_stream.go
@@ -81,9 +81,9 @@ type ServerStreamForClient[Res any] struct {
 	receiver Receiver
 	msg      Res
 	// Error from client construction. If non-nil, return for all calls.
-	err error
+	constructErr error
 	// Error from Receive().
-	recvErr error
+	receiveErr error
 }
 
 // Receive advances the stream to the next message, which will then be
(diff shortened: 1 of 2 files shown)


client_stream.go
```diff
@@ -77,7 +80,10 @@ func (c *ClientStreamForClient[Req, Res]) CloseAndReceive() (*Response[Res], err
 type ServerStreamForClient[Res any] struct {
 	receiver Receiver
 	msg      Res
-	err      error
+	// Error from client construction. If non-nil, return for all calls.
+	err error
+	// Error from Receive().
+	recvErr error
```
pkwarren, line 86: Felt cleaner to separate this into a different error so we can disambiguate a client initialization error from an error returned by Receive.

akshayjshah: House style discourages abbreviated names like this, even though they're common in Go. Can we rename these two fields to `constructErr` and `receiveErr`?

client_stream_test.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/bufbuild/connect-go/internal/assert"
+	pingv1 "github.com/bufbuild/connect-go/internal/gen/connect/ping/v1"
+)
+
+func TestClientStreamForClient_NoPanics(t *testing.T) {
+	t.Parallel()
+	initErr := errors.New("client init failure")
+	cs := &ClientStreamForClient[pingv1.PingRequest, pingv1.PingResponse]{err: initErr}
+	assert.ErrorIs(t, cs.Send(&pingv1.PingRequest{}), initErr)
+	assert.Equal(t, len(cs.RequestHeader()), 0)
```
akshayjshah, line 31: Can you also make sure that writes to the map don't panic? (Same for all the request header tests below.)

client_stream_test.go
```diff
@@ -0,0 +1,62 @@
...
+	res, err := cs.CloseAndReceive()
+	assert.Nil(t, res)
+	assert.ErrorIs(t, err, initErr)
```
akshayjshah, line 34: A+ code spelunking, appreciate you using this assertion helper!

akshayjshah (approved):
Great catch. Please rename the errors and extend the tests to cover writing request headers, but feel free to land once that's done - no need for more review.

PR #266, opened 2022-06-06
fix minCode constant

pkwarren:
The minCode constant should be set to CodeCanceled, not CodeUnknown. Add
test coverage to ensure that known codes don't unmarshal when using
"code_<num>" encoding.

### DIFF
--- a/code_test.go
+++ b/code_test.go
@@ -15,6 +15,7 @@
 package connect
 
 import (
+	"strconv"
 	"strings"
 	"testing"
 
@@ -38,6 +39,8 @@ func TestCode(t *testing.T) {
 		assertCodeRoundTrips(t, code)
 	}
 	assertCodeRoundTrips(t, Code(999))
+	var canceledInvalid Code
+	assert.NotNil(t, canceledInvalid.UnmarshalText([]byte("code_"+strconv.Itoa(int(CodeCanceled)))))
 }
 
 func assertCodeRoundTrips(tb testing.TB, code Code) {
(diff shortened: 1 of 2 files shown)


code_test.go
```diff
@@ -38,6 +39,8 @@ func TestCode(t *testing.T) {
 		assertCodeRoundTrips(t, code)
 	}
 	assertCodeRoundTrips(t, Code(999))
+	var canceledInvalid Code
+	assert.NotNil(t, canceledInvalid.UnmarshalText([]byte("code_"+strconv.Itoa(int(CodeCanceled)))))
```
pkwarren, line 43: This test failed before updating minCode.

code_test.go
```diff
@@ -47,4 +50,8 @@ func assertCodeRoundTrips(tb testing.TB, code Code) {
 	var decoded Code
 	assert.Nil(tb, decoded.UnmarshalText(encoded))
 	assert.Equal(tb, decoded, code)
+	if code >= minCode && code <= maxCode {
+		var invalid Code
+		assert.NotNil(tb, invalid.UnmarshalText([]byte("code_"+strconv.Itoa(int(code)))))
+	}
```
akshayjshah, line 56: :+1: Can you also leave an explanatory comment here? Something like this:

```
// For the known codes, we only accept the canonical string representation: "canceled", not "code_1".
```

code_test.go
akshayjshah: Gotcha. Definitely see how this test would have kept us from making the mistake we did, but I'm not sure it's worth keeping - it covers the same code as the change to `assertCodeRoundTrips`, and if we make a logical mistake again we'll almost certainly make the same mistake updating this test.

Up to you, I'm just saying that I'd be comfortable without this addition.

akshayjshah (approved):
LGTM, thanks for the fix! Some small suggestions, but you don't need another round of review if you decide to adopt them.

PR #267, opened 2022-06-06
ensure consistent nil response handling

pkwarren:
Update duplex_http_call to handle a nil response in the Read() and
ResponseStatusCode methods for consistency with other methods. Change
the signature of `ResponseStatusCode` to return an error if the returned
response is nil.

### DIFF
--- a/duplex_http_call.go
+++ b/duplex_http_call.go
@@ -17,6 +17,7 @@ package connect
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"net/http"
 	"sync"
@@ -149,6 +150,9 @@ func (d *duplexHTTPCall) Read(data []byte) (int, error) {
 		d.SetError(err)
 		return 0, wrapIfContextError(err)
 	}
+	if d.response == nil {
+		return 0, io.EOF
+	}
 	return d.response.Body.Read(data)
 }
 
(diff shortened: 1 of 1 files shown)


duplex_http_call.go
```diff
@@ -164,9 +168,12 @@ func (d *duplexHTTPCall) CloseRead() error {
 }
 
 // ResponseStatusCode is the response's HTTP status code.
-func (d *duplexHTTPCall) ResponseStatusCode() int {
+func (d *duplexHTTPCall) ResponseStatusCode() (int, error) {
 	<-d.responseReady
-	return d.response.StatusCode
+	if d.response == nil {
+		return 0, fmt.Errorf("nil response from %v", d.request.URL)
```
pkwarren, line 174: This feels better than making any assumptions based on a 0 return code.

duplex_http_call.go
```diff
@@ -149,6 +150,9 @@ func (d *duplexHTTPCall) Read(data []byte) (int, error) {
 		d.SetError(err)
 		return 0, wrapIfContextError(err)
 	}
+	if d.response == nil {
+		return 0, io.EOF
```
pkwarren, line 154: Does this make sense as a valid return from Read for a nil response?

duplex_http_call.go
```diff
@@ -248,7 +255,8 @@ func (d *duplexHTTPCall) makeRequest() {
 	if err := d.validateResponse(response); err != nil {
 		d.SetError(err)
 		return
-	} else if (d.streamType&StreamTypeBidi) == StreamTypeBidi && response.ProtoMajor < 2 {
+	}
+	if (d.streamType&StreamTypeBidi) == StreamTypeBidi && response.ProtoMajor < 2 {
```
pkwarren, line 259: Small style thing here (no else after return). I'm happy to revert though if we're ok with this style.

akshayjshah (approved):

akshayjshah:
Same as usual - lgtm, please address comments but doesn't need additional review.

PR #269, opened 2022-06-06
Use registration order to prioritize client Accept-Encoding

### LINKED ISSUE #259: Random Compression Method Selection
Say you register 4 compression methods on both the handler and the client.

It seems like the server picks the first compression method it recognizes, unless a specific method was used when sending the request.

However, looking at the [client side](https://github.com/bufbuild/connect-go/blob/c35ee255c4f0265262da15a5f431d761a7498b93/compression.go#L160) the order the methods are sent is random since it collects names from a map.

As far as I can tell this makes it impossible to select an order of preference from the client side. Furthermore it appears to me that `gzip` cannot be removed from the pool, only replaced.

akshayjshah:
We're currently randomizing the client's list of accepted encodings when
sending them to the server. This is bad, since the server uses the first
algorithm it recognizes: ideally the client would send them in
preference order.

This PR changes the client to treat the last registered algorithm as the
most preferred. To minimize divergence, we do the same for handlers.

Fixes #259.

### DIFF
--- a/option.go
+++ b/option.go
@@ -29,7 +29,10 @@ type ClientOption interface {
 
 // WithAcceptCompression makes a compression algorithm available to a client.
 // Clients ask servers to compress responses using any of the registered
-// algorithms. It's safe to use this option liberally: servers will ignore any
+// algorithms. The first registered algorithm is treated as the least
+// preferred, and the last registered algorithm is the most preferred.
+//
+// It's safe to use this option liberally: servers will ignore any
 // compression algorithms they don't support. To compress requests, pair this
 // option with WithSendCompression.
 //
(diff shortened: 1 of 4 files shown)


option.go
```diff
@@ -251,18 +254,19 @@ type compressionOption struct {
...
 func (o *compressionOption) applyToHandler(config *handlerConfig) {
-	o.apply(config.CompressionPools)
-}
-
-func (o *compressionOption) apply(m map[string]*compressionPool) {
 	if o.Name == "" || o.CompressionPool == nil {
 		return
 	}
-	m[o.Name] = o.CompressionPool
+	config.CompressionPools[o.Name] = o.CompressionPool
```
bufdev, line 268: Just jumped out at me as a potential NPE, but I do know we always make sure this is initialized right?

bufdev (approved):

option.go
akshayjshah: Yep! Always initialized.

PR #278, opened 2022-06-07
add tests for WithCompressMinBytes

pkwarren:
Add test cases to verify the behavior of the WithCompressMinBytes
handler option.

### DIFF
--- a/connect_ext_test.go
+++ b/connect_ext_test.go
@@ -517,24 +517,26 @@ func TestCompressMinBytes(t *testing.T) {
 	})
 	client := server.Client()
 
-	getPingResponse := func(pingText string) *http.Response {
+	getPingResponse := func(t *testing.T, pingText string) *http.Response {
 		request := &pingv1.PingRequest{Text: pingText}
 		requestBytes, err := proto.Marshal(request)
 		assert.Nil(t, err)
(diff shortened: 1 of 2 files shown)


interceptor_ext_test.go
```diff
@@ -234,7 +234,6 @@ func (h *headerInterceptor) WrapUnary(next connect.UnaryFunc) connect.UnaryFunc
 		h.inspectResponseHeader(req.Spec(), res.Header())
 		return res, nil
 	}
-	return connect.UnaryFunc(call)
```
pkwarren, line 237: There were two places in this test that were unnecessarily wrapped in `connect.UnaryFunc(...)`.

connect_ext_test.go
```diff
@@ -503,6 +504,42 @@ func TestBidiRequiresHTTP2(t *testing.T) {
 	)
 }
 
+func TestCompressMinBytes_Enabled(t *testing.T) {
```
mfridman, line 507: Isn't this testing both compressed and uncompressed, the `_Enabled` suffix seems redundant? There might be a good reason though.

pkwarren: By default CompressMinBytes is set to 0 which effectively disables it. I can remove that though.

mfridman (approved):

mfridman (approved):

PR #280, opened 2022-06-07
add test for custom handler/client compression

pkwarren:
Create a test which adds support for 'deflate' compression using the
stdlib. Expand test coverage in handler_ext_test to add a test for
unsupported content encoding. Consistently call
`d.BlockUntilResponseReady()` in all places in duplex_http_call.

### DIFF
--- a/connect_ext_test.go
+++ b/connect_ext_test.go
@@ -16,6 +16,7 @@ package connect_test
 
 import (
 	"bytes"
+	"compress/flate"
 	"context"
 	"errors"
 	"fmt"
@@ -541,6 +542,38 @@ func TestCompressMinBytes(t *testing.T) {
 	})
 }
 
+func Test_CustomCompression(t *testing.T) {
+	t.Parallel()
+	mux := http.NewServeMux()
+	compressionName := "deflate"
+	decompressor := func() connect.Decompressor {
+		return newDeflateReader(http.NoBody)
+	}
+	compressor := func() connect.Compressor {
+		w, err := flate.NewWriter(&strings.Builder{}, flate.DefaultCompression)
+		if err != nil {
+			t.Fatalf("failed to create flate writer: %v", err)
+		}
+		return w
+	}
+	mux.Handle(pingv1connect.NewPingServiceHandler(
+		pingServer{},
+		connect.WithCompression(compressionName, decompressor, compressor),
+	))
+	server := httptest.NewServer(mux)
+	defer server.Close()
+
+	client := pingv1connect.NewPingServiceClient(server.Client(),
+		server.URL,
+		connect.WithAcceptCompression(compressionName, decompressor, compressor),
+		connect.WithSendCompression(compressionName),
+	)
+	request := &pingv1.PingRequest{Text: "testing 1..2..3.."}
+	response, err := client.Ping(context.Background(), connect.NewRequest(request))
+	assert.Nil(t, err)
+	assert.Equal(t, response.Msg, &pingv1.PingResponse{Text: request.Text})
+}
+
 type failCodec struct{}
 
 func (c failCodec) Name() string {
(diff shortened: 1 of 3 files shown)


connect_ext_test.go
```diff
@@ -541,6 +542,38 @@ func TestCompressMinBytes(t *testing.T) {
...
+	t.Parallel()
+	mux := http.NewServeMux()
+	compressionName := "deflate"
+	decompressor := func() connect.Decompressor {
+		return newDeflateReader(http.NoBody)
+	}
+	compressor := func() connect.Compressor {
+		w, err := flate.NewWriter(&strings.Builder{}, flate.DefaultCompression)
+		if err != nil {
+			t.Fatalf("failed to create flate writer: %v", err)
```
pkwarren, line 555: It might be nice if the interface supported returning an error here, because we could fail back to identity encoding, etc. if the compression has issues. Right now, the only fallback for real use cases would be using `panic`.

connect_ext_test.go
```diff
@@ -714,3 +747,28 @@ func (p pingServer) CumSum(
...
+func (d *deflateReader) Read(p []byte) (int, error) {
+	return d.r.Read(p)
+}
+
+func (d *deflateReader) Close() error {
+	return d.r.Close()
+}
+
+func (d *deflateReader) Reset(reader io.Reader) error {
+	if resetter, ok := d.r.(flate.Resetter); ok {
```
pkwarren, line 768: Unfortunately `flate.NewReader` returns an `io.ReadCloser` instead of a concrete type. This type assertion should always succeed though.

duplex_http_call.go
```diff
@@ -169,7 +169,7 @@ func (d *duplexHTTPCall) CloseRead() error {
 
 // ResponseStatusCode is the response's HTTP status code.
 func (d *duplexHTTPCall) ResponseStatusCode() (int, error) {
-	<-d.responseReady
+	d.BlockUntilResponseReady()
```
pkwarren, line 172: Noticed we were inconsistently using the `BlockUntilResponseReady` vs. directly accessing the channel.

handler_ext_test.go
```diff
@@ -15,21 +15,26 @@
...
+	pingv1 "github.com/bufbuild/connect-go/internal/gen/connect/ping/v1"
 	"github.com/bufbuild/connect-go/internal/gen/connect/ping/v1/pingv1connect"
 )
 
 func TestHandler_ServeHTTP(t *testing.T) {
 	t.Parallel()
 	mux := http.NewServeMux()
 	mux.Handle(pingv1connect.NewPingServiceHandler(
-		pingServer{},
+		successPingServer{},
```
pkwarren, line 35: Changed implementation to return a successful response by default so we can test error codes added elsewhere.

handler_ext_test.go
```diff
@@ -62,4 +67,32 @@ func TestHandler_ServeHTTP(t *testing.T) {
 			"application/proto",
 		}, ", "))
 	})
+
+	t.Run("unsupported_content_encoding", func(t *testing.T) {
+		t.Parallel()
+		req, err := http.NewRequest(http.MethodPost, server.URL+pingProcedure, strings.NewReader("{}"))
+		assert.Nil(t, err)
+		req.Header.Set("Content-Type", "application/json")
+		req.Header.Set("Content-Encoding", "invalid")
```
pkwarren, line 76: Added some test coverage for unsupported `Content-Encoding`.

connect_ext_test.go
doriable: Yeah, since the writer constructor returns an error, it would be nice to handle that. On the other hand, the writer could be constructed outside of the `compressor` but then it would be leaking outside of the `compressor`.

doriable: This is an interesting point, because on one hand, the writer could be constructed outside of the closure, however, it does make sense to encapsulate the writer type inside the closure and not have it leak.

doriable (approved):
This looks good to me :)

handler_ext_test.go
```diff
@@ -62,4 +67,32 @@ func TestHandler_ServeHTTP(t *testing.T) {
...
+		defer resp.Body.Close()
+		type errorMessage struct {
+			Code, Message string
+		}
+		var message errorMessage
+		err = json.NewDecoder(resp.Body).Decode(&message)
+		assert.Nil(t, err)
+		assert.Equal(t, message.Message, `unknown compression "invalid": supported encodings are gzip`)
+		assert.Equal(t, message.Code, connect.CodeUnimplemented.String())
+		assert.Equal(t, resp.StatusCode, http.StatusNotFound)
```
mfridman, line 88: Nit, closer to where response is returned. Reading this top-down I was curious what the status code was and didn't expect it to be at the end.

PR #282, opened 2022-06-07
add tests for client init failures

pkwarren:
Add tests which cover the case where `connect.NewClient` fails during
initialization. In this case, the initial error is saved and any
operations on the client fail fast with the initial error.

### DIFF
--- a/client_ext_test.go
+++ b/client_ext_test.go
@@ -1,0 +1,70 @@
+// Copyright 2021-2022 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


client_ext_test.go
```diff
@@ -0,0 +1,70 @@
...
+		http.DefaultClient,
+		"http://127.0.0.1:8080",
+		// This triggers an error during initialization, so each call will short circuit returning an error.
+		connect.WithSendCompression("invalid"),
+	)
+	validateExpectedError := func(t *testing.T, err error) {
+		t.Helper()
+		assert.NotNil(t, err)
+		var connectErr *connect.Error
+		assert.True(t, errors.As(err, &connectErr))
```
mfridman, line 41: Should this be `require` instead of `assert`, otherwise calling `.Message()` _may_ panic?

pkwarren: This is using an internal `assert` which calls `t.Fatal`: https://github.com/bufbuild/connect-go/blob/38baeb35be9870f922323ee612b8b0c66aa68223/internal/assert/assert.go#L186

doriable (approved):

PR #283, opened 2022-06-07
don't set accept-encoding in responses

pkwarren:
For both the Connect and gRPC protocols, we're setting 'accept-encoding'
or 'grpc-accept-encoding' in the responses (this should only be sent in
requests).


akshayjshah:
gRPC also allows this for responses: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#responses

Connect should, too - I forgot to include this when I documented the protocol. I'll open a small PR to fix the docs.

PR #284, opened 2022-06-08
WIP: require grpc-status trailer, even for successful RPCs

### LINKED ISSUE #279: Client in gRPC mode: Return an error if the response does not contain Trailers.
In the official gRPC-go implementation, if the response does not come with Trailers, the client will return an error with a nil response message.

With a connect-go client in gRPC mode, it will not complain about the absence of Trailers upon a successful unary call. However, it will produce an EOF error if the RPC is finished with an error.

Trailer headers carry gRPC status, error information, and other metadata. Although it may be safe(?) to ignore upon a successful unary call, such a manner does not comply with the gRPC protocol. 

If the client is in gRPC mode and the response does not have Trailers, return an error.

Upon successful unary calls, return the decoded message and an error, and let the user decide whether to exploit the feature.

**Extra context**
I was experimenting with placing a gRPC Cloud Run service behind Firebase Hosting, the Firebase Hosting ingress will trim the HTTP2 Trailer headers in the response.

akshayjshah:
Currently, we're assuming that gRPC responses without a `Grpc-Status` trailer
are successes. This case doesn't come up in practice, but the specification
_requires_ this trailer (the message and base64-encoded status proto are
optional).

We should tighten up and error if we get a response without this trailer. I
think I've got the meat of it working here, but it looks like we're still doing
something wrong in some circumstances with bidi streams (perhaps we're not
_sending_ the trailer from the handler?).

My suggestion is to see if one of you can fix the failing test case, run the
resulting commit through the crosstests, and verify that everything works. If
it's all working, merge it in! If you'd like, I can look the fix over when I
return and we can cut a patch release then.

Fixes #279.

### DIFF
--- a/protocol_grpc.go
+++ b/protocol_grpc.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"math"
 	"net/http"
 	"net/textproto"
@@ -62,7 +63,8 @@ var (
 		{time.Minute, 'M'},
 		{time.Hour, 'H'},
 	}
-	grpcTimeoutUnitLookup = make(map[byte]time.Duration)
+	grpcTimeoutUnitLookup        = make(map[byte]time.Duration)
+	errTrailersWithoutGRPCStatus = errorf(CodeInternal, "gRPC protocol error: no %s trailer", grpcHeaderStatus)
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


pkwarren:
Incorporated this fix (with some tweaks) and the test from #285 into #288.

PR #285, opened 2022-06-08
add test for gRPC call responses without Trailers

lixin9311:
Related issue: https://github.com/bufbuild/connect-go/issues/279

You might need to change the error message or use a predefined message accordingly.

And tests for streaming calls are missing.

```go
assert.True(
	t,
	strings.HasSuffix(connectErr.Message(), "server closed the stream without sending trailers"),
)
```

### DIFF
--- a/connect_ext_test.go
+++ b/connect_ext_test.go
@@ -465,6 +465,64 @@ func TestGRPCMarshalStatusError(t *testing.T) {
 	assertInternalError(t, connect.WithGRPCWeb())
 }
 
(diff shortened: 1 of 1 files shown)


akshayjshah:
Thanks for the PR, @lixin9311! Really appreciate the contribution 😻  Someone from Buf will take a look within the next few days.

pkwarren:
Thanks for the test! I've merged it in to #288 and will commit it as part of that PR.

PR #288, opened 2022-06-09
Require grpc-status trailer with gRPC protocol

### LINKED ISSUE #279: Client in gRPC mode: Return an error if the response does not contain Trailers.

pkwarren:
Currently, we're assuming that gRPC responses without a Grpc-Status trailer
are successes. This case doesn't come up in practice, but the specification
requires this trailer (the message and base64-encoded status proto are
optional).

Fixes #279.

### DIFF
--- a/connect_ext_test.go
+++ b/connect_ext_test.go
@@ -472,6 +472,64 @@ func TestGRPCMarshalStatusError(t *testing.T) {
 	assertInternalError(t, connect.WithGRPCWeb())
 }
 
+func TestGRPCUnaryMissingTrailersError(t *testing.T) {
+	t.Parallel()
+
+	trimTrailers := func(handler http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			r.Header.Del("Te")
+			handler.ServeHTTP(&trimTrailerWriter{w: w}, r)
+		})
+	}
+
+	mux := http.NewServeMux()
+	mux.Handle(pingv1connect.NewPingServiceHandler(
+		pingServer{},
+	))
+	server := httptest.NewUnstartedServer(trimTrailers(mux))
+	server.EnableHTTP2 = true
+	server.StartTLS()
+	defer server.Close()
+
+	assertInternalErrorOnSuccessUnary := func(tb testing.TB, opts ...connect.ClientOption) {
+		tb.Helper()
+		client := pingv1connect.NewPingServiceClient(server.Client(), server.URL, opts...)
+		request := connect.NewRequest(&pingv1.PingRequest{Number: 1, Text: "foobar"})
+		_, err := client.Ping(context.Background(), request)
+		tb.Log(err)
+		assert.NotNil(t, err)
+		var connectErr *connect.Error
+		ok := errors.As(err, &connectErr)
+		assert.True(t, ok)
+		assert.Equal(t, connectErr.Code(), connect.CodeInternal)
+		assert.True(
+			t,
+			strings.HasSuffix(connectErr.Message(), "server closed the stream without sending trailers"),
+		)
+	}
+
+	assertInternalErrorOnFailedUnary := func(tb testing.TB, opts ...connect.ClientOption) {
+		tb.Helper()
+		client := pingv1connect.NewPingServiceClient(server.Client(), server.URL, opts...)
+		request := connect.NewRequest(&pingv1.FailRequest{Code: int32(connect.CodeResourceExhausted)})
+		_, err := client.Fail(context.Background(), request)
+		tb.Log(err)
+		assert.NotNil(t, err)
+		var connectErr *connect.Error
+		ok := errors.As(err, &connectErr)
+		assert.True(t, ok)
+		assert.Equal(t, connectErr.Code(), connect.CodeInternal)
+		assert.True(
+			t,
+			strings.HasSuffix(connectErr.Message(), "server closed the stream without sending trailers"),
+		)
+	}
+
+	// Only applies to gRPC protocol.
+	assertInternalErrorOnFailedUnary(t, connect.WithGRPC())
+	assertInternalErrorOnSuccessUnary(t, connect.WithGRPC())
+}
+
 func TestUnavailableIfHostInvalid(t *testing.T) {
 	t.Parallel()
 	client := pingv1connect.NewPingServiceClient(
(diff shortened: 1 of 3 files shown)


protocol_connect.go
```diff
@@ -354,7 +354,7 @@ func (s *connectClientSender) Send(message any) error {
 	return nil // must be a literal nil: nil *Error is a non-nil error
 }
 
-func (s *connectClientSender) Close(err error) error {
+func (s *connectClientSender) Close(_ error) error {
```
bufdev, line 357: You can just do `(error)`

protocol_grpc.go
```diff
@@ -750,20 +768,34 @@ func grpcHTTPToCode(httpCode int) Code {
 // binary Protobuf format, even if the messages in the request/response stream
 // use a different codec. Consequently, this function needs a Protobuf codec to
 // unmarshal error information in the headers.
-func grpcErrorFromTrailer(bufferPool *bufferPool, protobuf Codec, trailer http.Header) *Error {
+func grpcErrorFromHeaders(bufferPool *bufferPool, protobuf Codec, trailer, header http.Header) *Error {
```
bufdev, line 771: `header` should come before `trailer`

protocol_grpc.go
```diff
@@ -750,20 +768,34 @@ func grpcHTTPToCode(httpCode int) Code {
...
 	}
 
 	code, err := strconv.ParseUint(codeHeader, 10 /* base */, 32 /* bitsize */)
 	if err != nil {
 		return errorf(CodeInternal, "gRPC protocol error: invalid error code %q", codeHeader)
 	}
-	message := grpcPercentDecode(bufferPool, trailer.Get(grpcHeaderMessage))
+	messageHeader := trailer.Get(grpcHeaderMessage)
+	if messageHeader == "" {
+		messageHeader = header.Get(grpcHeaderMessage)
```
bufdev, line 790: This seems to be saying that if you can't get something from the trailer, then get it from the header as a fallback - is this part of the gRPC protocol specification?

pkwarren: Changed the logic to only get values from the trailers or headers (depending on where we find the `grpc-status` header), but not from both. I believe the relevant part of the spec is here: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#responses.

> Response → (Response-Headers *Length-Prefixed-Message Trailers) / _Trailers-Only_

I'll walk through the failing gRPC-Web test again to make sure we're doing things right on our end and also run the crosstests against this branch.

bufdev: `Status must be sent in Trailers even if the status code is OK.`

This would seem to imply that the status can only be sent in trailers, right?

pkwarren: A Trailers-Only response returns HTTP-Status, Content-Type, and Trailers (Status [Status-Message] *Custom-Metadata). For gRPC-Web, we match the behavior in Envoy to skip encoding trailers in the message body and just return them as regular headers: https://github.com/bufbuild/connect-go/blob/871250f0e169d8afe8fe097165a025e1a558e1f8/protocol_grpc.go#L494-L504

I've taken a conservative approach with this PR and only look in both trailers and then headers for the gRPC-Web client. If we see in the future that we need to support a Trailing-Headers response for the gRPC protocol as well, we'd just update https://github.com/bufbuild/connect-go/pull/288 to pass `r.header` instead of `nil`.

akshayjshah: The protocol's logical trailers can be sent as HTTP trailers, or in some cases, as HTTP headers. Confusingly, the latter case is called a "trailers-only" response.

bufdev (approved):
This looks fine from my side, but wait for @akshayjshah's review when he is back

akshayjshah (approved):
~~Definitely want to see tests passing for gRPC-Web clients too, but otherwise lgtm. No need for more review.~~

Thanks for getting this over the line, @pkwarren!

protocol_grpc.go
```diff
@@ -421,7 +428,13 @@ func (r *grpcWebClientReceiver) Receive(message any) error {
 	}
 	// See if the server sent an explicit error in the gRPC-Web trailers.
 	mergeHeaders(r.trailer, r.unmarshaler.WebTrailer())
-	if serverErr := grpcErrorFromTrailer(r.bufferPool, r.protobuf, r.trailer); serverErr != nil {
+	serverErr := grpcErrorFromHeaders(r.bufferPool, r.protobuf, r.header, r.trailer)
```
akshayjshah, line 431: We've already checked `r.header` in `grpcValidateResponse`. Do we need to check it again, or can we pass `nil` here?

akshayjshah (changes requested):
Sorry, I stamped this too early. Let me pull this down and wrap my head around all of this again :/

connect_ext_test.go
```diff
@@ -472,6 +472,82 @@ func TestGRPCMarshalStatusError(t *testing.T) {
...
+	mux := http.NewServeMux()
+	mux.Handle(pingv1connect.NewPingServiceHandler(
+		pingServer{checkMetadata: true},
+	))
+	server := httptest.NewUnstartedServer(trimTrailers(mux))
+	server.EnableHTTP2 = true
+	server.StartTLS()
+	t.Cleanup(func() {
+		server.Close()
+	})
```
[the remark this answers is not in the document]
akshayjshah: Ah, I see. This makes things work with parallelized subtests!

akshayjshah (approved):

akshayjshah:
@pkwarren I pushed a few commits here.

They revert the portion of your change where `grpcErrorFromHeaders` consults both the headers and trailers and automatically chooses the correct bag of metadata. From what I can see, we shouldn't need that - `validateResponse` already handles trailers-only responses and splits the HTTP headers into leading and trailing metadata.

Instead, we can take a brutalist approach and just check the headers in `Receive`. Take a look and let me know if this seems reasonable to you.

pkwarren:
> Instead, we can take a brutalist approach and just check the headers in `Receive`. Take a look and let me know if this seems reasonable to you.

This is much cleaner - thanks for refactoring.

akshayjshah:
Okay, I'm going to squash Buf commits together and rebase this in so we can preserve @lixin9311's commit.

PR #290, opened 2022-06-09
remove unused no-op stream code paths

pkwarren:
Both implementations of protocol (Connect and gRPC) always return a
non-nil sender and receiver. Remove an unused code path and the no-op
stream implementation which is no longer used (and currently untested).


handler.go
```diff
@@ -258,16 +258,6 @@ func (h *Handler) ServeHTTP(responseWriter http.ResponseWriter, request *http.Re
...
-	// If NewStream or SetTimeout errored and the protocol doesn't want the
-	// error sent to the client, sender and/or receiver may be nil. We still
-	// want the error to be seen by interceptors, so we provide no-op Sender
-	// and Receiver implementations.
-	if clientVisibleError != nil && sender == nil {
-		sender = newNopSender(h.spec, responseWriter.Header(), make(http.Header))
-	}
-	if clientVisibleError != nil && receiver == nil {
-		receiver = newNopReceiver(h.spec, request.Header, request.Trailer)
-	}
```
pkwarren, line 270: These were showing up as uncovered by tests - looking at the implementation the sender/receiver are never nil:

* Connect: https://github.com/bufbuild/connect-go/blob/c9b9bcbabc808ef86e5bedf0d080916c3047df8d/protocol_connect.go#L203-L212
* gRPC: https://github.com/bufbuild/connect-go/blob/c9b9bcbabc808ef86e5bedf0d080916c3047df8d/protocol_grpc.go#L173-L180

amckinney:
I'd normally agree with this, but I'm afraid that we'll end up exposing a potential `panic` later on if we ever export the [protocolHandler](https://github.com/bufbuild/connect-go/blob/108eb83516e513d5193cf52cfc50fa87a5999bb7/protocol.go#L72) interface for custom implementations. We likely won't export this in the near term (if ever), but I'd bet that we'd forget to reintroduce this handling again if we needed to.

If we're worried about some of the untested paths, we could implement a lightweight `protocolHandler` instance that returns a `nil` `Sender` and `Receiver` so that the the `nop{Receiver,Sender}` types are used.

What are your thoughts?

akshayjshah:
Haven't looked in detail (too busy with pizza!). 

I think I wanted this so that errors which prevent starting a stream (like failure to negotiate compression) are visible to interceptors. To do that, we need an innermost Sender/Receiver to wrap.

PR #291, opened 2022-06-10
verify error when interceptor returns wrong response

pkwarren:
Interceptors deal in AnyRequest/AnyResponse, so it is possible to return
the wrong type of response without a compile error. Add a test to verify
we correctly handle this error in the client.

### DIFF
--- a/connect_ext_test.go
+++ b/connect_ext_test.go
@@ -606,6 +606,30 @@ func TestInvalidHeaderTimeout(t *testing.T) {
 	})
 }
 
+func TestInterceptorReturnsWrongType(t *testing.T) {
+	t.Parallel()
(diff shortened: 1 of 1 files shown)


connect_ext_test.go
```diff
@@ -606,6 +606,30 @@ func TestInvalidHeaderTimeout(t *testing.T) {
 	})
 }
 
+func TestInterceptorReturnsWrongType(t *testing.T) {
```
pkwarren, line 609: Verifies this section of the client code: https://github.com/bufbuild/connect-go/blob/c9b9bcbabc808ef86e5bedf0d080916c3047df8d/client.go#L103-L106

doriable (approved):

PR #293, opened 2022-06-10
Add tests for nil sender/receiver

pkwarren:
In the HTTP handler, there is code that handles returning nil
sender/receiver and continues to call interceptors. This is currently
unreachable with the exported API - add tests using a custom
protocolHandler implementation.

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -262,8 +262,10 @@ func (h *Handler) ServeHTTP(responseWriter http.ResponseWriter, request *http.Re
 	// error sent to the client, sender and/or receiver may be nil. We still
 	// want the error to be seen by interceptors, so we provide no-op Sender
 	// and Receiver implementations.
+	sendHTTPInternalError := false
 	if clientVisibleError != nil && sender == nil {
 		sender = newNopSender(h.spec, responseWriter.Header(), make(http.Header))
+		sendHTTPInternalError = true
 	}
 	if clientVisibleError != nil && receiver == nil {
 		receiver = newNopReceiver(h.spec, request.Header, request.Trailer)
@@ -274,6 +276,10 @@ func (h *Handler) ServeHTTP(responseWriter http.ResponseWriter, request *http.Re
 		receiver = interceptor.WrapStreamReceiver(ctx, receiver)
 	}
 	h.implementation(ctx, sender, receiver, clientVisibleError)
+	if sendHTTPInternalError {
+		// protocolHandler returned a nil sender, so we can't send a proper HTTP response.
+		responseWriter.WriteHeader(http.StatusInternalServerError)
+	}
 }
 
 type handlerConfig struct {
--- a/handler_test.go
+++ b/handler_test.go
@@ -1,0 +1,62 @@
+package connect
+
+import (
+	"context"
+	"errors"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/bufbuild/connect-go/internal/assert"
+	pingv1 "github.com/bufbuild/connect-go/internal/gen/connect/ping/v1"
+)
+
+var errFailedToCreateStream = errors.New("failed to create stream")
+
+func TestInterceptorsCalledIfSenderReceiverNil(t *testing.T) {
+	t.Parallel()
+	mux := http.NewServeMux()
+	pingPath := "/connect.ping.v1.PingService/Ping"
+	handler := NewUnaryHandler[pingv1.PingRequest, pingv1.PingResponse](
+		pingPath,
+		func(ctx context.Context, request *Request[pingv1.PingRequest]) (*Response[pingv1.PingResponse], error) {
+			t.Error("shouldn't call handler implementation")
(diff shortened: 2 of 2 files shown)


handler.go
```diff
@@ -274,6 +276,10 @@ func (h *Handler) ServeHTTP(responseWriter http.ResponseWriter, request *http.Re
 		receiver = interceptor.WrapStreamReceiver(ctx, receiver)
 	}
 	h.implementation(ctx, sender, receiver, clientVisibleError)
+	if sendHTTPInternalError {
+		// protocolHandler returned a nil sender, so we can't send a proper HTTP response.
+		responseWriter.WriteHeader(http.StatusInternalServerError)
```
pkwarren, line 281: If the sender is nil, that means we wrap a no-op sender (which sends nothing across the wire). In that case, we should at least send back an error from the handler. Without this logic, we'd send back an empty HTTP 200 response.

I'm not sure if we want to encode more details about the `clientVisibleError` in a response body here but this should at least reflect that an error occurred on the client.

handler_test.go
```diff
@@ -0,0 +1,76 @@
...
+	pingv1 "github.com/bufbuild/connect-go/internal/gen/connect/ping/v1"
+)
+
+var errFailedToCreateStream = errors.New("failed to create stream")
+
+func TestInterceptorsCalledIfSenderReceiverNil(t *testing.T) {
+	t.Parallel()
+	mux := http.NewServeMux()
+	pingPath := "/connect.ping.v1.PingService/Ping"
+	handler := NewUnaryHandler[pingv1.PingRequest, pingv1.PingResponse](
```
pkwarren, line 34: Have to use NewUnaryHandler here directly since the test needs to access the unexported `protocolHandlers` field so we can't use the connect generated code (import cycle).

handler_test.go
```diff
@@ -0,0 +1,76 @@
...
+		pingPath,
+		func(ctx context.Context, request *Request[pingv1.PingRequest]) (*Response[pingv1.PingResponse], error) {
+			t.Error("shouldn't call handler implementation")
```
pkwarren, line 37: We shouldn't call the unary method if we get an error creating the stream - from reading the code, the expectation is that we should only call interceptors.

handler_test.go
```diff
@@ -0,0 +1,76 @@
...
+			return nil, NewError(CodeUnimplemented, nil)
+		},
+		WithInterceptors(UnaryInterceptorFunc(func(next UnaryFunc) UnaryFunc {
+			return func(ctx context.Context, request AnyRequest) (AnyResponse, error) {
+				_, err := next(ctx, request)
+				assert.NotNil(t, err)
+				assert.True(t, errors.Is(err, errFailedToCreateStream))
+				return nil, err
```
pkwarren, line 45: Validate that the interceptor is called with the clientVisibleError.

Note that this might limit what an interceptor can really do since changing the response or returning a custom error won't have any impact at the client (if sender is nil). The request will also be an empty request and not very useful if the receiver is nil. This will only be useful for a limited set of interceptors (i.e. logging/tracing).

handler_test.go
```diff
@@ -0,0 +1,76 @@
...
+			}
+		})),
+	)
+	// Override gRPC handler with custom one
+	handler.protocolHandlers = append([]protocolHandler{&nilProtocolHandler{}}, handler.protocolHandlers...)
```
pkwarren, line 50: For now, have to manually access unexported field to trigger the code path in the handler when the sender/receiver are nil.

pkwarren (commented):
This is an alternative to https://github.com/bufbuild/connect-go/pull/290 and should verify the code path if we ever introduce a new protocol handler in the future that returns a nil sender/receiver or open the API to add new protocol handlers.

handler.go
amckinney: It looks like the `clientVisibleError` could be a typed `*connect.Error`. If so, we should map its error code to the associated HTTP code rather than always using `http.StatusInternalError`. For example, I noticed that `clientVisibleError` is sometimes set to to the `timeoutErr` (which will often be initialized with `CodeInvalidArgument`).

However, the `clientVisibleError` is a detail of the `protocolHandler` that was used to serve the request (e.g. gRPC, Connect, etc). Fortunately, the gRPC and Connect [error codes](https://connect.build/docs/protocol/#error-codes) match, so we _should_ be able to use the same code transformation regardless of the `protocolHandler` that was used. I'd need to take a closer look at this though to refresh.

```go
responseWriter.WriteHeader(connectCodeToHTTP(CodeOf(clientVisibleError)))
```

However, this might be another place for concern if we ever expose another `protocolHandler` that doesn't have the same `connect.Code` -> HTTP Code transformation. We'll need to verify this behavior before we move forward here - I have a feeling we might not want to use the `responseWriter.WriteHeader` call in this context.

akshayjshah:
Mea culpa - I've steered you wrong here. :sob: #290 is definitely the better approach.

PR #298, opened 2022-06-17
Always use unix-style paths for CodeGeneratorResponses

bufdev:
Fixes https://github.com/bufbuild/connect-go/issues/297.

`protoc` and `buf` both expect `CodeGeneratorResponses` to have unix-style paths https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/compiler/plugin.proto#L124=


akshayjshah (approved):
Looks good! Didn't know about that part of the API contract.

Issue #295, opened 2022-06-13
RFC: Client side load balancer


mattrobenolt:
Within gRPC, one big feature is service discovery via xDS and other client side load balancing features. This is actually one of the biggest selling points imo for running gRPC internally for communication between services without needing to introduce another external proxy such as Envoy to facilitate this.

Now, I understand with connect-go, we are free to use a gRPC client and maintain these features, but would be nice to see a more official connect-go implementation.

Granted, one nice part about connect-go is the fact that you can just use a normal http.Client or specifically, an `HTTPClient` interface which the default `http.Client` implements.

What I'd like to propose is some general purpose, likely an `http.RoundTripper` that implements some basic features of the gRPC load balancer. I believe the gRPC implementation is too... much, but having basic connection pooling and round robin behavior per-IP address opens up the door to basic service discovery and getting the max value out of it.

For example, between internal services, let's say in Kubernetes, it's common to reference an A record that represents the service itself. This A record contains multiple IP addresses. A naive `http.Client` opens up one persistent connection to the hostname, and then directs all requests to the same IP address internally. Granted, the client will open up concurrent connections and potentially bind connections to another host, but there's no predictability in this behavior.

Having a RoundTripper that was intelligent enough to do a DNS lookup, then round robin, or other load balancing algorithms across the IPs would be super nice.

I understand this isn't necessarily in the scope of `connect-go` since this is sorta being outsourced to the larger Go community by adopting `net/http` as the interface, but I think having some official RoundTripper provided by `connect-go` would be super useful.

I'm going to be working on a POC for this, unrelated to `connect-go` with not necessarily the intention for it to be upstreamed, but if we can come to agreements on if this is something connect-go should provide (even in a separate repo), I'd be happy to help contribute there.

In general, I personally want a RoundTripper for generic HTTP, which would happen to be also ideal within the context of `connect-go`. Right now, this is the barrier of entry to adopting `connect-go` to replace our gRPC clients. We've already replaced our gRPC servers. :)

I'm happy to provide more insight or suggestions as needed.

akshayjshah:
🎉 We intended to do exactly this, right down to focusing on the http.RoundTripper interface! We haven't started work yet, so the timing is perfect - PlanetScale's requirements can help inform our design.

I think it probably makes sense to keep all this in a separate repository - it's useful for any HTTP client, and I bet we'll end up with some dependencies on connect-grpchealth-go. 

Some questions:

[...]

PR #300, opened 2022-06-17
feat: Add NewHTTPClient

mattrobenolt:
The HTTPClient interface is effectively the same thing as an
http.RoundTripper, but instead, I understand the desire for HTTPClient's
interface to work with a *http.Client.

It's also rather intuitive to directly use an http.RoundTripper which is
just the inner implementation of an *http.Client.

This makes it slightly easier to use different transports or different
implementations by just swapping out a RoundTripper without needing to
wrap in an *http.Client yourself.

The stdlib *http.Client comes with extra baggage that might not be
desirable for use within RPC, such as following redirects and management
for cookies.

### DIFF
--- a/connect.go
+++ b/connect.go
@@ -218,6 +218,18 @@ type HTTPClient interface {
 	Do(*http.Request) (*http.Response, error)
 }
 
+type httpClient struct{ http.RoundTripper }
+
+func (c httpClient) Do(req *http.Request) (*http.Response, error) {
+	return c.RoundTrip(req)
+}
+
+// NewHTTPClient wraps an http.RoundTripper to make it satisfy the HTTPClient
+// interface.
+func NewHTTPClient(rt http.RoundTripper) HTTPClient {
+	return httpClient{rt}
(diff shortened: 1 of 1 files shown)


akshayjshah:
Agreed that `http.Client` comes with a little more than we'd like: following redirects _by default_ sure seems bad for RPC. I'm less bothered by cookie support, since the zero value of `Client` disables it. I do think I'd miss the whole-transaction `Timeout` offered by `http.Client`.

Less concretely, we've avoided adding code like this to this repo because it muddies the "it's just net/http" message to less-painstaking readers.

What do you think of putting this into the same package of `net/http` extensions we discussed in #295?

mattrobenolt:
Yeah, that's entirely fine. This is obviously something trivial to do on your own, and I agree it'd make sense in some future `httputil` package or whatever we come up with.

akshayjshah:
OK, great - I'll close this and plan to revisit this in our soon-to-be-written HTTP package.

PR #303, opened 2022-06-20
add CI github actions build for Windows

### LINKED ISSUE #299: Add Windows to GitHub Actions
See #297. This should be relatively straightforward, we can do it similarly to how we did it in bufbuild/buf.

pkwarren:
Add a new github actions build to run Windows build and test.

Fixes #299.


connect_ext_test.go
```diff
@@ -93,8 +93,9 @@ func TestServer(t *testing.T) {
 			assert.Equal(t, connect.CodeOf(err), connect.CodeInvalidArgument)
 		})
 		t.Run("ping_timout", func(t *testing.T) {
-			ctx, cancel := context.WithTimeout(context.Background(), time.Nanosecond)
+			ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond)
```
pkwarren, line 96: Tests were failing on Windows - changed to use a longer timeout and explicitly wait prior to making the call.

.github/workflows/windows.yaml
```diff
@@ -0,0 +1,56 @@
...
+        uses: actions/cache@v3
+        with:
+          path: ${{ steps.go-cache-paths.outputs.go-mod }}
+          key: ${{ runner.os }}-connect-ci-go-mod-${{ hashFiles('**/go.sum') }}
+      - name: Test
+        shell: bash
+        if: success()
+        run: |
+          go build ./...
+          go test -vet=off -race -cover ./...
```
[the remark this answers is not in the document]
akshayjshah: Makes sense to me! I'd maybe drop `-cover`, since we're not writing any platform-specific code.

.github/workflows/windows.yaml
```diff
@@ -0,0 +1,54 @@
...
+permissions: read-all
+jobs:
+  ci:
+    runs-on: windows-2019
+    strategy:
+      matrix:
+        go-version: [1.18.x]
+        include:
+          - go: 1.18.x
+            latest: true
```
akshayjshah, line 22: Since we're not running lint, we don't need this `include` block - I don't think we're referencing `latest` anywhere.

.github/workflows/windows.yaml
```diff
@@ -0,0 +1,54 @@
...
+    branches: [main]
+  schedule:
+    - cron: '15 22 * * *'
+  workflow_dispatch: {} # support manual runs
+# Prevent writing to the repository using the CI token.
+# Ref: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions
+permissions: read-all
+jobs:
+  ci:
+    runs-on: windows-2019
```
akshayjshah, line 16: Can we use `windows-latest` instead? Outside the CLI and core, I'd like to minimize the number of places we need to bump version numbers. (Plus, I don't think any of us are tracking Windows server releases carefully.)

akshayjshah (approved):
Some small suggestions and edits, but please land once they're addressed! No need for more review.

PR #305, opened 2022-06-20
attempt to fix flaky timeout tests

pkwarren:
Instead of using time.Sleep with small increments, instead create a
context using `WithDeadline(...)` with an expired time.


akshayjshah (commented):
I think that this changes the code we're exercising pretty dramatically - I'd expect the client to return an error before even hitting the network now.

The tests are pretty fast, and we don't do this in many places. I'd consider adding something like https://github.com/uber/tchannel-go/blob/186ff8fb310040caaa0b0092baaf6048ddd04d78/testutils/timeout.go#L49 to our tests and setting the scale factor to 10 (or 100) in CI.

akshayjshah (approved):

PR #306, opened 2022-06-22
fix race condition caught by unit tests

pkwarren:
Update to cache the underlying error, not the connect.Error (as this is
mutated with updates to meta).


akshayjshah (approved):
Wow, I did _not_ expect this. 🤦  Thanks!

bufdev (approved):

PR #308, opened 2022-06-22
Dedupe compression algorithm names

mattrobenolt:
After #269, it was possible to generate an Accept-Encoding header with
duplicate values.

This is purely aesthetic, but an intentional side effect. This happens
if you're trying to override a previously defined one.

When the values were purely in map, they naturally deduped because they
keys overwrote previous values.

This adds tests to cover this compressionPool for the strict ordering of
names, as well as removing duplicate values.

I added a very light generic orderedSet[T comparable] to this, since I
think there are other places this can be leveraged. It's just a nice
tool to have. I can remove this if you'd prefer.

### DIFF
--- a/compression_test.go
+++ b/compression_test.go
@@ -1,0 +1,77 @@
+// Copyright 2021-2022 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package connect
+
+import (
+	"io"
+	"testing"
+
+	"github.com/bufbuild/connect-go/internal/assert"
+)
+
+func TestReadOnlyCompressionPools(t *testing.T) {
+	t.Parallel()
+	names := []string{"foo", "foo", "bar", "baz"}
+	compressionPools := map[string]*compressionPool{}
+	compressionNames := []string{}
+	for _, name := range names {
+		name := name
+		compressionPools[name] = newCompressionPool(
+			func() Decompressor { return noopDecompressor{name} },
+			func() Compressor { return noopCompressor{name} },
+		)
+		compressionNames = append(compressionNames, name)
+	}
+	pools := newReadOnlyCompressionPools(
+		compressionPools,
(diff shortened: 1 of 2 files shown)


compression_test.go
```diff
@@ -0,0 +1,87 @@
...
+package connect
+
+import (
+	"io"
+	"testing"
+
+	"github.com/bufbuild/connect-go/internal/assert"
+)
+
+type noopCompressor struct{ name string }
```
bufdev, line 24: internal style guide: public types first, then private - move below tests

compression_test.go
```diff
@@ -0,0 +1,87 @@
...
+	return len(p), nil
+}
+func (noopDecompressor) Close() error          { return nil }
+func (noopDecompressor) Reset(io.Reader) error { return nil }
+
+func TestReadOnlyCompressionPools(t *testing.T) {
+	t.Parallel()
+
+	names := []string{"foo", "foo", "bar", "baz"}
+
```
bufdev, line 44: nit: we're less heavy on newlines :-)

sorted_set.go
```diff
@@ -0,0 +1,34 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package connect
+
+type sortedSet[T comparable] struct {
```
bufdev, line 17: I would just move this to `compression.go`, at the bottom of the file

bufdev, line 17: I don't think this is sorted, this is more `orderedSet`

mattrobenolt: Happy to do this, but extracted it just in case it was thought to be used elsewhere. I originally had it at the bottom.

akshayjshah:
Thanks, @mattrobenolt! I'll take a detailed look later tonight.

akshayjshah (approved):
Really appreciate you flagging this and PRing a fix, @mattrobenolt! I'm going to push a commit making the changes I suggested here and then merge it all in - no need to put you through another round of edits.

mattrobenolt:
Thanks for the carry. :) I will learn.

compression_test.go
```diff
@@ -0,0 +1,56 @@
...
+package connect
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/bufbuild/connect-go/internal/assert"
+	"google.golang.org/protobuf/types/known/emptypb"
```
mattrobenolt, line 24: Unrelated: TIL the emptypb, this is useful with something else I was testing.

akshayjshah: It's the weirdest (and therefore my favorite) of the well-known types :)

I end up using a lot in cases like this.

PR #315, opened 2022-07-07
enable ci/windows testing on go 1.19rc1

pkwarren:


akshayjshah (approved):

akshayjshah:
We shouldn't be using 1.19 gofmt! We should only be checking formatting and lint on 1.18.x.

akshayjshah (approved):
LGTM

PR #311, opened 2022-07-01
add support for ReadMaxBytes option

pkwarren:
Update connect-go to allow configuring a ReadMaxBytes option. This is
useful to avoid excessive memory/cpu usage when dealing with very large
messages.

PR #318, opened 2022-07-14
fix issues on 32-bit platforms

pkwarren:
Update comparison to work properly on 32-bit architectures.

Addresses an issue introduced in #311.

### DIFF
--- a/compression.go
+++ b/compression.go
@@ -81,7 +81,7 @@ func (c *compressionPool) Decompress(dst *bytes.Buffer, src *bytes.Buffer, readM
 		return errorf(CodeInvalidArgument, "get decompressor: %w", err)
 	}
 	reader := io.Reader(decompressor)
-	if readMaxBytes > 0 && readMaxBytes < math.MaxInt64 {
+	if readMaxBytes > 0 && int64(readMaxBytes) < math.MaxInt64 {
 		reader = io.LimitReader(decompressor, int64(readMaxBytes)+1)
 	}
 	bytesRead, err := dst.ReadFrom(reader)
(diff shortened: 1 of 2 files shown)


doriable (approved):
🙏🏼

akshayjshah (approved):

compression.go
```diff
@@ -81,7 +81,7 @@ func (c *compressionPool) Decompress(dst *bytes.Buffer, src *bytes.Buffer, readM
 		return errorf(CodeInvalidArgument, "get decompressor: %w", err)
 	}
 	reader := io.Reader(decompressor)
-	if readMaxBytes > 0 && readMaxBytes < math.MaxInt64 {
+	if readMaxBytes > 0 && int64(readMaxBytes) < math.MaxInt64 {
```
tomchwojkofrank, line 84: For 32 bit the clause after the `&&` will always be true.
Why not `readMaxBytes < math.MaxInt`?

pkwarren: The main reason this exists is just not to overflow when adding one below. Also https://pkg.go.dev/io#LimitReader and https://pkg.go.dev/io#ReaderFrom both deal with `int64` so easier to use that consistently.

tomchwojkofrank: But if `readMaxBytes` is `math.MaxInt32`, and you add anything to it, it will overflow?
If everything else is `int64` then the most correct thing to do is to use `int64` consistently everywhere.
If that isn't possible, then the bounds check should be checking against the variable's int size.

tomchwojkofrank: I'd either change the type of `readMaxBytes` to be `int64`, or create a local `int64` set to `int64(readMaxBytes)` and use that for the remainder of the function. Reduces the risk of missing inserting a cast somewhere, or having to constantly reason about the type and decide whether to cast.

pkwarren: Good call out - opened https://github.com/bufbuild/connect-go/pull/319 to improve.

PR #319, opened 2022-07-14
improve decompress signature

pkwarren:
Remove duplicate casts in compression pool's Decompress function by
ensuring the maxReadBytes parameter is an int64. This requires fewer
casts in the implementation to avoid potential issues.

### DIFF
--- a/compression.go
+++ b/compression.go
@@ -75,21 +75,21 @@ func newCompressionPool(
 	}
 }
 
-func (c *compressionPool) Decompress(dst *bytes.Buffer, src *bytes.Buffer, readMaxBytes int) *Error {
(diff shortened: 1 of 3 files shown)


compression.go
```diff
@@ -75,21 +75,21 @@ func newCompressionPool(
 	}
 }
 
-func (c *compressionPool) Decompress(dst *bytes.Buffer, src *bytes.Buffer, readMaxBytes int) *Error {
+func (c *compressionPool) Decompress(dst *bytes.Buffer, src *bytes.Buffer, readMaxBytes int64) *Error {
```
pkwarren, line 78: compressionPool isn't an exported type so we can change this to avoid casting in the logic below.

tomchwojkofrank (approved):
Thanks, looks good!

PR #321, opened 2022-07-15
Map HTTP/2 RST_STREAM codes back to RPC codes

akshayjshah:
HTTP/2 includes its own set of error codes, which are sent in RST_STREAM
frames. When servers send one of these codes to clients, we should map
the HTTP/2 code back to one of our status codes.

### DIFF
--- a/error.go
+++ b/error.go
@@ -283,7 +283,7 @@ func wrapIfRSTError(err error) error {
 	}
 	msg = strings.TrimSuffix(msg, fromPeerSuffix)
 	i := strings.LastIndex(msg, ";")
-	if i < 0 || i == len(msg) {
+	if i < 0 || i >= len(msg)-1 {
 		return err
 	}
 	msg = msg[i+1:]
--- a/error.go
+++ b/error.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"net/url"
 	"strings"
 
 	"google.golang.org/protobuf/proto"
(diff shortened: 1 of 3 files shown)


akshayjshah:
The string munging to extract HTTP/2 error codes is quite unpleasant; filed https://github.com/golang/go/issues/53896 to see if the Go team has any interest in making this a bit cleaner.

error.go
```diff
@@ -247,3 +248,59 @@ func wrapIfLikelyWithGRPCNotUsedError(err error) error {
...
+	msg := err.Error()
+	if !strings.HasPrefix(msg, streamErrPrefix) {
+		return err
+	}
+	if !strings.HasSuffix(msg, fromPeerSuffix) {
+		return err
+	}
+	msg = strings.TrimSuffix(msg, fromPeerSuffix)
+	i := strings.LastIndex(msg, ";")
+	if i < 0 || i == len(msg) {
```
pkwarren, line 286:
```suggestion
	if i < 0 {
```

I don't think LastIndex will return `len(msg)` - I think it will be in range 0-len(msg)-1.

pkwarren (approved):

error.go
akshayjshah: Off by one error on my part, thank you for catching this! I meant to check that `i >= len(msg)-1`, since we're slicing with `[i+1]` below.

pkwarren (approved):

Shudogg (commented):
Msg

Issue #328, opened 2022-07-20
Allow UnmarshalNew() in the ErrorDetail interface


ctsquires:
**Is your feature request related to a problem? Please describe.**

Quality of life improvement for parsing `ErrorDetail` as a client.

Currently parsing the slice of ErrorDetails into many possible underlying types feels a bit repetitive and we can't use a standard switch case on the type to figure out which error detail was returned. It would be nice if the ErrorDetail interface exposed the `UnmarshalNew()` functionality of the underlying `anypb.Any` type to allow using a traditional switch case.

This functionality is referenced in the anypb documentation https://pkg.go.dev/google.golang.org/protobuf/types/known/anypb#hdr-Unmarshaling_an_Any. 

Currently I believe clients need to iterate through every type and details, checking if the detail can successfully be unmarshalled to that type and then actioning if it successfully unmarshals.

eg. We currently have a few types that can be returned, and we construct an internal error model to hold these types if we receive them from a server, with the current setup we need to do something like below (we could iterate through the details and check each type, but both end up with similar amounts of code):

```go

func extractErrorDetails(connectErr *connect.Error) (MyErrorType, error) {
	var errorOpts []myErrorOption
	retryInfo := tryExtractRetryInfo(connectErr)
	if retryInfo != nil {
		errorOpts = append(errorOpts, retryInfoOption(retryInfo))
	}
	helpInfo := tryExtractHelpInfo(connectErr)
	if helpInfo != nil {
		errorOpts = append(errorOpts, helpInfoOption(retryInfo))
	}
	return newMyErrorType(errorOpts...), nil
}

func tryExtractRetryInfo(connectErr *connect.Error) *pb.RetryInfo {
	var retryInfo errdetails.RetryInfo
	for _, any := range connectErr.Details() {
		if err := any.UnmarshalTo(&retryInfo); err == nil {
			return &retryInfo
		}
	}
}

func tryExtractHelpInfo(connectErr *connect.Error) (*pb.HelpInfo, ok) {
	var helpInfo errdetails.HelpInfo
	for _, any := range connectErr.Details() {
		if err := any.UnmarshalTo(&helpInfo); err == nil {
			return &helpInfo
		}
	}
}
```

**Describe the solution you'd like**

It would be nice if the `ErrorDetail` interface exposed `UnmarshalNew()` 

```go
type ErrorDetail interface {
	proto.Message

	MessageName() protoreflect.FullName
	UnmarshalTo(proto.Message) error
[...]

PR #329, opened 2022-07-25
Send binary data with error details

akshayjshah:
Currently, the connect protocol serializes error details as JSON. This
requires having protobuf descriptors for all types used as details,
which makes it difficult for proxies to translate to and from the gRPC
protocols: it's hard to know which types are used as details because
they're not listed in the schema, and we don't want to be rebuilding
proxies every time a downstream service changes its schema.

This PR switches to an updated version of the connect protocol. Error
details are still serialized with human-readable JSON where possible,
but we include the binary protobuf data as a base64-encoded string.
This makes it easy to convert to and from `proto.Any` without
descriptors.

As a side effect, this converts `ErrorDetail` into a concrete type and
reduces our public API's ties to the `proto.Any` interface. It also
addresses #328.


error.go
```diff
@@ -23,26 +23,47 @@ import (
...
+	if err != nil {
+		return nil, err
+	}
+	return &ErrorDetail{pb: pb}, nil
+}
+
+// Type is the fully-qualified name of the Detail's Protobuf message (for
+// example, acme.foo.v1.FooDetail).
+func (d *ErrorDetail) Type() string {
+	return strings.TrimPrefix(d.pb.TypeUrl, defaultAnyResolverPrefix)
```
bufdev, line 59: Can you add a comment as to why this is trimmed? I can't exactly remember and might be helpful for people in the future

bufdev (approved):
Looks generally good to me

error.go
akshayjshah: Sure, can do.

doriable (approved):

protocol_connect.go
```diff
@@ -803,48 +794,89 @@ func (u *connectUnaryUnmarshaler) UnmarshalFunc(message any, unmarshal func([]by
...
+		// of which need JSON escaping.
+		out = []byte(`{"@type": "`)
+		out = append(out, strings.TrimPrefix(defaultAnyResolverPrefix, d.pb.TypeUrl)...)
+		out = append(out, '"')
 	}
-	return (&protoJSONCodec{}).Marshal(wire)
+	// Since we know that the "@value" key and base64-encoded bytes don't need
+	// escaping, it's simplest to work with the output as raw bytes.
+	out = bytes.TrimRight(out, "}")
+	out = append(out, `, "@value": "`...)
```
pkwarren, line 817: Are there any edge cases here where someone could break things - i.e. setting `json_name` to `@value`? It might be safest to marshal the JSON representation to another key?

pkwarren (approved):

protocol_connect.go
akshayjshah: I don't think it's possible to generate conflicts here; `@` isn't allowed in protobuf field names, and none of the implementations I'm aware of allow customizing this.

akshayjshah: Update, I'm uninformed: the `json_name` option allows all sorts of nonsense.

akshayjshah: Changed to a nested object :/

pkwarren (approved):

protocol_connect.go
timostamm: _All_ sorts of nonsense indeed.

```proto
    string a = 3 [json_name = "sameJsonName"];
    string b = 4 [json_name = "sameJsonName"];
    string c = 5;
    string d = 6 [json_name = "c"];
    string e = 7 [json_name = ""];
    string f = 8 [json_name = "@type"];
```

I really liked `@value` and don't think it's a real problem, considering the example above compiles. If `@value` is the source of truth, and the other properties are informal, it would be fine to override any user-defined `@value` with our property.

That said, the nested object does make it clear that the encoded form is the source of truth, and that `debug` is really only there to have a human-readable form on the wire.

timostamm (approved):

PR #337, opened 2022-08-02
add HTTP error helper

### LINKED ISSUE #331: Provide function to write RPC errors from HTTP middleware
Even though Connect handlers work with `net/http` middleware, it's often necessary to re-implement existing HTTP middleware as Connect interceptors so that clients get nicely-annotated errors. This is a little silly.

It'd be nice for Connect to include a function that HTTP middleware can use to encode Connect errors. `func WriteError(http.ResponseWriter, *http.Request, error)` might work.

The status quo is okay, but forces users to jump through more hoops that we'd like.

rhbuf:
The mapping from connect error codes to HTTP status codes is currently private and not accessible to users who are implementing connect in their existing HTTP servers. This helper has been added to allow those users to easily update existing HTTP middlewares to handle connect errors.

fixes #331


http_util.go
```diff
@@ -0,0 +1,36 @@
...
+package connect
+
+import (
+	"encoding/json"
+	"net/http"
+)
+
+// WriteError should be called from HTTP middlewares to handle connect errors. It
+// writes a status code encodes a connect error as JSON,  and writes to the response.
+func WriteError(responseWriter http.ResponseWriter, err error) error {
```
rhbuf, line 24: @akshayjshah originally had a request in here, but I didn't see any reason to include it. Potentially it was there to close the request body, and when I lifted this code from https://github.com/bufbuild/connect-go/blob/main/protocol_connect.go#L553 I did see that function closing the request. But when looking into https://pkg.go.dev/net/http#Request.Close I believe that `WriteError` doesn't need to close the request if the server will do so anyway.

http_util_test.go
```diff
@@ -0,0 +1,74 @@
...
+		if readErr != nil {
+			t.Fatalf(readErr.Error())
+		}
+		assert.Equal(t, respErr.Code, CodeUnknown)
+		assert.Equal(t, respErr.Message, err.Error())
+	})
+}
+
+func readConnectErr(resp *http.Response) (*connectWireError, error) {
+	data, err := io.ReadAll(resp.Body)
```
pkwarren, line 65: It would be good to add `t.Helper()` here.

http_util.go
```diff
@@ -0,0 +1,36 @@
...
+
+package connect
+
+import (
+	"encoding/json"
+	"net/http"
+)
+
+// WriteError should be called from HTTP middlewares to handle connect errors. It
+// writes a status code encodes a connect error as JSON,  and writes to the response.
```
pkwarren, line 23:
```suggestion
// writes a status code, encodes a connect error as JSON, and writes to the response.
```

http_util.go
```diff
@@ -0,0 +1,36 @@
...
+func WriteError(responseWriter http.ResponseWriter, err error) error {
+	responseWriter.Header().Set(headerContentType, connectUnaryContentTypeJSON)
+	responseWriter.WriteHeader(connectCodeToHTTP(CodeOf(err)))
+	data, marshalErr := json.Marshal(newConnectWireError(err))
```
pkwarren, line 27: nit: Instead of writing to an intermediate []byte, can you use `json.NewEncoder(w).Encode(...)`?

pkwarren (commented):
Looks good - just a couple minor things.

akshayjshah:
Awesome, I'm glad you've got some bandwidth to take this on @rhbuf! This is a great start.

Unfortunately, this function only returns valid responses for unary requests using the Connect protocol. If we're going to ship this, we shouldn't make users figure out the protocol in use, the compression scheme, etc - I think we should respond appropriately to unary Connect, streaming Connect, gRPC, and gRPC-Web requests, so that RPC clients get full-fidelity error codes, messages, and details.

rhbuf:
> Awesome, I'm glad you've got some bandwidth to take this on @rhbuf! This is a great start.
> 
> Unfortunately, this function only returns valid responses for unary requests using the Connect protocol. If we're going to ship this, we shouldn't make users figure out the protocol in use, the compression scheme, etc - I think we should respond appropriately to unary Connect, streaming Connect, gRPC, and gRPC-Web requests, so that RPC clients get full-fidelity error codes, messages, and details. Doing that requires access to the request headers.

Roger that, I will come back in a week or so and flesh this out. I've switched directions in my current work and no longer needs this HTTP helper function, but I want to finish what I've started when I have some time.

akshayjshah:
> I've switched directions in my current work and no longer needs this HTTP helper function, but I want to finish what I've started when I have some time.

Don't sweat it! This isn't a "you touch it, you buy it" system :) I'll try to take over and finish up; if I don't get to it and you're free, feel free to pick it up again.

akshayjshah:
@rhbuf and @pkwarren, I took a shot at finishing this up - let me know what you think!

pkwarren (approved):

mattrobenolt:
🙏 thanks all, this is what I've been waiting for as well.

PR #341, opened 2022-08-05
add SendMaxBytes option for clients and handlers

### LINKED ISSUE #330: SendMaxBytes config option
In https://github.com/bufbuild/connect-go/pull/311, a ReadMaxBytes option was added, but to go along with that, when you control both the client and server, an equivalent SendMaxBytes (mostly relevant for the client) to pair with ReadMaxBytes would be nice and prevent sending anything over the wire.

In our case, we configure ReadMaxBytes on the server, and would like to pair this also on our client so if a user attempts to send a message that is too large (and would ultimately be rejected by the server anyways), we can bail early on the client before doing any extra work.

This coincides with the gRPC `MaxCallSendMsgSize` option.

pkwarren:
In addition to ReadMaxBytes, add a SendMaxBytes option which will
configure the maximum payload sent by clients or handlers.

Fixes #330.


mattrobenolt:
❤️ thanks all!

akshayjshah (approved):
This is great - very neatly done, with excellent test coverage. Thank you!

Address nits and questions as you see fit, no need for further review before merging.

PR #347, opened 2022-08-16
Update dependencies (incl. golangci-lint)

akshayjshah:
- Lint on Go 1.19
- Use new Godoc links
- Update deps, including golangci-lint


cmd/protoc-gen-connect-go/main.go
```diff
@@ -19,23 +19,23 @@
...
 //
-//   version: v1
-//   plugins:
-//     name: go
-//     out: gen
-//     name: connect-go
-//     out: gen
+//	version: v1
+//	plugins:
+//	  name: go
```
pkwarren, line 28: Shouldn't there be a '-' before each name to separate plugins?

pkwarren (approved):

cmd/protoc-gen-connect-go/main.go
akshayjshah: 🤦🏽‍♂️ Yep!

PR #350, opened 2022-08-17
In iterators, allocate new msg for each Receive

### LINKED ISSUE #345: Don't reuse messages between Receive() calls
Disclosure: I'm not sure how well this pattern is reflected across the board for all streaming APIs, but I'm assuming it behaves similarly.

When using a service such as:

```proto
service Foo {
  rpc Bar(BarRequest) returns (stream BarResponse){}
}
```

Within the client streaming decoder, the same object is received _into_ for each iteration.

Specifically in a loop such as:

```go
for stream.Receive() {
  msg := stream.Msg()
}
```

`stream.Msg()` is the exact same object in memory. In most cases that use `proto.Unmarshal(...)`, it seems to be safe, though I'm not 100% sure.

If you happen to use a different codec, it's possible that the codec might not expect this to be an already potentially populated object.

While I don't think this is directly an issue with connect-go, since it appears the built in codec works fine `proto.Unmarshal(...)` seems to handle this case sufficiently, it's not a bug perse.

For contrast though, this behavior is distinctly different in grpc-go, so if there's code that's intended to interop, it might catch folks offguard (like it did for us).

In grpc-go, within `stream.Recv()`, it distinctly allocates a new struct for each call.

What I'd propose here though for connect-go is changing the behavior of `stream.Receive()` to allocate a new struct each time to avoid any issues. It's a slight regression in performance due to an extra allocation, but safely I think is preferable for the out of the box behavior.

Similarly though, I'd propose introducing a `ReceiveMsg(m any)` API that took a pointer, so I, as the caller, can control this behavior. Personally, in my case, I'd allocate one message first, then on each iteration, call `m.Reset()` or something similar to control this myself.

I think splitting `stream.Receive()` from a `stream.ReceiveMsg(m any)` makes sense from a UX POV too. I'd expect since the implementation details of `Receive()` are hidden, I'd expect more that this is 100% safe and I shouldn't be concerned with what it's doing. Whereas when `ReceiveMsg(m any)` takes the struct to inflate, I'd expect to be in control and make sure the struct is safe for re-use on each cycle.

This explicit split too allows a bit more control of any memory management if I were to want to leverage a memory pool (`sync.Pool`) or something similar for heavy workloads. Neither API currently allows that, whereas with grpc-go, we could.

I put together a concrete example to show off the differences: https://github.com/mattrobenolt/connect-go-example

The output of running this is as follows:

```bash
$ go run main.go
connect: Got greeting 0xc0001a4400: []string{"greeting 1"}
connect: Got greeting 0xc0001a4400: []string{"greeting 1", "greeting 2"}
grpc-go: Got greeting 0xc000042e40: []string{"greeting 1"}
grpc-go: Got greeting 0xc000352540: []string{"greeting 2"}
```

In grpc-go, you can see that it allocates a new struct within it's generated code: https://github.com/mattrobenolt/connect-go-example/blob/main/gen/greet/v1/greet_grpc.pb.go#L60-L67

I didn't test, but I assume within: https://github.com/bufbuild/connect-go/blob/main/client_stream.go#L91-L97

but I'm not actually sure how to mirror this behavior with generics. Since we don't know the concrete type of `Res`, we can't `new(Res)` or assign it to nil, etc. So it might require a slightly more complex fix, unfortunately. Maybe `s.msg = *(new(Res))` or similar?

Update:

Patching `s.msg = *(new(Res))` seems to be the closest immediate fix? It ends up, obviously reusing the same pointer, since it's the same chunk of memory, but it is filled with a new empty struct overwriting it. This might be the result you're going for.

My change:

```diff
diff --git a/client_stream.go b/client_stream.go
index 49cbea1..c0a9c3e 100644
--- a/client_stream.go
+++ b/client_stream.go
@@ -92,6 +92,7 @@ func (s *ServerStreamForClient[Res]) Receive() bool {
 	if s.constructErr != nil || s.receiveErr != nil {
 		return false
 	}
+	s.msg = *(new(Res))
 	s.receiveErr = s.conn.Receive(&s.msg)
 	return s.receiveErr == nil
 }
```

And my new test case output:
```
$ go run main.go
connect: Got greeting 0xc0001a4470: []string{"greeting 1"}
connect: Got greeting 0xc0001a4470: []string{"greeting 2"}
grpc-go: Got greeting 0xc0002d2300: []string{"greeting 1"}
grpc-go: Got greeting 0xc0003b22c0: []string{"greeting 2"}
```

So as we can see, it's the same memory address, but doesn't exhibit the same issue since the new object allocated is zero'd out.

Update 2:

ok, tbh this seems _better_, but I can't fully grasp the consequences of this, but it makes more logical sense to me.

```go
diff --git a/client_stream.go b/client_stream.go
index 49cbea1..89ec09f 100644
--- a/client_stream.go
+++ b/client_stream.go
@@ -76,7 +76,7 @@ func (c *ClientStreamForClient[Req, Res]) CloseAndReceive() (*Response[Res], err
 // exported constructor function.
 type ServerStreamForClient[Res any] struct {
 	conn StreamingClientConn
-	msg  Res
+	msg  *Res
 	// Error from client construction. If non-nil, return for all calls.
 	constructErr error
 	// Error from conn.Receive().
@@ -92,7 +92,8 @@ func (s *ServerStreamForClient[Res]) Receive() bool {
 	if s.constructErr != nil || s.receiveErr != nil {
 		return false
 	}
-	s.receiveErr = s.conn.Receive(&s.msg)
+	s.msg = new(Res)
+	s.receiveErr = s.conn.Receive(s.msg)
 	return s.receiveErr == nil
 }
 
@@ -100,7 +101,7 @@ func (s *ServerStreamForClient[Res]) Receive() bool {
 // returned message points to data that will be overwritten by the next call to
 // Receive.
 func (s *ServerStreamForClient[Res]) Msg() *Res {
-	return &s.msg
+	return s.msg
 }
 
 // Err returns the first non-EOF error that was encountered by Receive.
```

And now we, more as I'd expect, get a new memory address for each `Receive() + Msg()` combo.

akshayjshah:
Currently, iterator-style streams re-use the same memory for each
received message. This is a satisfying little efficiency, but it's very
likely to cause confusion among users - they'll retain a reference to a
message, call `Receive()` again, and wonder why the retained reference
has changed out from under them. We should anticipate this confusion and
allocate for each `Receive`. As a nice side effect, this makes
`vtprotobuf` Codecs behave properly by default - no need to explicitly
reset messages.

This PR also adds a `Conn()` method to each stream that exposes the
underlying connection. This lets users who want a different high-level
stream API write their own wrapper types: perhaps they'd like to avoid
allocations more aggressively, or they want every stream constructor to
also require an auth token.

Fixes #345.

### DIFF
--- a/client_stream.go
+++ b/client_stream.go
@@ -70,13 +70,19 @@ func (c *ClientStreamForClient[Req, Res]) CloseAndReceive() (*Response[Res], err
 	return response, c.conn.CloseResponse()
 }
 
+// Conn exposes the underlying StreamingClientConn. This may be useful if
+// you'd prefer to wrap the connection in a different high-level API.
+func (c *ClientStreamForClient[Req, Res]) Conn() (StreamingClientConn, error) {
+	return c.conn, c.err
+}
+
 // ServerStreamForClient is the client's view of a server streaming RPC.
 //
 // It's returned from [Client].CallServerStream, but doesn't currently have an
 // exported constructor function.
 type ServerStreamForClient[Res any] struct {
 	conn StreamingClientConn
-	msg  Res
+	msg  *Res
 	// Error from client construction. If non-nil, return for all calls.
 	constructErr error
 	// Error from conn.Receive().
@@ -92,15 +98,17 @@ func (s *ServerStreamForClient[Res]) Receive() bool {
 	if s.constructErr != nil || s.receiveErr != nil {
 		return false
 	}
-	s.receiveErr = s.conn.Receive(&s.msg)
+	s.msg = new(Res)
+	s.receiveErr = s.conn.Receive(s.msg)
 	return s.receiveErr == nil
 }
 
-// Msg returns the most recent message unmarshaled by a call to Receive. The
-// returned message points to data that will be overwritten by the next call to
-// Receive.
+// Msg returns the most recent message unmarshaled by a call to Receive.
 func (s *ServerStreamForClient[Res]) Msg() *Res {
-	return &s.msg
+	if s.msg == nil {
+		s.msg = new(Res)
+	}
+	return s.msg
 }
 
 // Err returns the first non-EOF error that was encountered by Receive.
@@ -138,6 +146,12 @@ func (s *ServerStreamForClient[Res]) Close() error {
 		return s.constructErr
 	}
 	return s.conn.CloseResponse()
+}
+
+// Conn exposes the underlying StreamingClientConn. This may be useful if
+// you'd prefer to wrap the connection in a different high-level API.
+func (s *ServerStreamForClient[Res]) Conn() (StreamingClientConn, error) {
+	return s.conn, s.constructErr
 }
 
 // BidiStreamForClient is the client's view of a bidirectional streaming RPC.
@@ -217,4 +231,10 @@ func (b *BidiStreamForClient[Req, Res]) ResponseTrailer() http.Header {
 		return http.Header{}
 	}
 	return b.conn.ResponseTrailer()
+}
+
+// Conn exposes the underlying StreamingClientConn. This may be useful if
+// you'd prefer to wrap the connection in a different high-level API.
+func (b *BidiStreamForClient[Req, Res]) Conn() (StreamingClientConn, error) {
+	return b.conn, b.err
 }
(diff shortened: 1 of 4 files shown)


pkwarren (approved):

mattrobenolt:
This is great, thank you. ❤️

PR #352, opened 2022-08-22
add a test for concurrent bidi streams

pkwarren:
Create a test which spins up 100 goroutines each calling a bidi call
against the same server (with random values) to verify concurrent
streams.

### DIFF
--- a/connect_ext_test.go
+++ b/connect_ext_test.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"math/rand"
 	"net/http"
 	"net/http/httptest"
 	"strings"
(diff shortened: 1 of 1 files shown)


connect_ext_test.go
```diff
@@ -387,6 +388,44 @@ func TestServer(t *testing.T) {
...
+			var total int64
+			sum := client.CumSum(context.Background())
+			startWg.Wait()
+			for i := 0; i < 100; i++ {
+				num := rand.Int63n(1000)
+				total += num
+				assert.Nil(t, sum.Send(&pingv1.CumSumRequest{Number: num}))
+				resp, err := sum.Receive()
+				assert.Nil(t, err)
+				assert.Equal(t, total, resp.Sum)
```
akshayjshah, line 419: I'm fairly sure that `assert` uses methods on `testing.T` that aren't safe to call concurrently. You may want to double-check me on that, though :)

If I'm right about that, I think we could get away with a `chan error` and some non-blocking writes in the client goroutines. After all the client goroutines shut down, we can try reading from the channel and fail the test if we find an error. (I don't think we want to serialize the test by introducing locks.)

connect_ext_test.go
```diff
@@ -387,6 +388,44 @@ func TestServer(t *testing.T) {
...
+	t.Parallel()
+	mux := http.NewServeMux()
+	mux.Handle(pingv1connect.NewPingServiceHandler(pingServer{}))
+	server := httptest.NewUnstartedServer(mux)
+	server.EnableHTTP2 = true
+	server.StartTLS()
+	t.Cleanup(server.Close)
+	var done sync.WaitGroup
+	var startWg sync.WaitGroup
+	startWg.Add(1)
```
akshayjshah, line 404: Huh, I'm in the habit of using a plain `chan struct{}` instead of start wait group. I like this more!

What do you think of making this a one-liner? `var start, done sync.WaitGroup`

.golangci.yml
```diff
@@ -83,3 +83,6 @@ issues:
     - linters:
         - revive
       text: "receiver-naming:.*for invalid-type$"
+    - linters: [gosec]
+      path: connect_ext_test.go
+      text: "G404: Use of weak random number generator.*"
```
akshayjshah, line 88: Can we do this with a `//nolint: gosec` comment instead? It's less annoying to keep a comment on the same line as the rand usage.

connect_ext_test.go
pkwarren: Replaced calls to internal `assert.*` with `t.Error*` calls which appear safe to call from multiple goroutines.

akshayjshah (approved):

akshayjshah:
Good to go, up to you whether you'd like to change the lint ignore :)

PR #355, opened 2022-09-01
Add support for http.MaxBytesHandler

akshayjshah:
The `ReadMaxBytes` option is nice, but we should also return
properly-coded errors when `net/http.MaxBytesHandler` middleware is in
play. This has no API surface area, but it's a nice improvement.

### DIFF
--- a/envelope.go
+++ b/envelope.go
@@ -18,7 +18,6 @@ import (
 	"bytes"
 	"encoding/binary"
 	"errors"
-	"fmt"
 	"io"
 )
 
@@ -202,7 +201,7 @@ func (r *envelopeReader) Read(env *envelope) *Error {
 		if connectErr, ok := asError(err); ok {
 			return connectErr
 		}
-		if maxBytesErr := asMaxBytesError("read 5 byte message prefix", err); maxBytesErr != nil {
+		if maxBytesErr := asMaxBytesError(err, "read 5 byte message prefix"); maxBytesErr != nil {
 			// We're reading from an http.MaxBytesHandler, and we've exceeded the read limit.
 			return maxBytesErr
 		}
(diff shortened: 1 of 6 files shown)


akshayjshah:
@robbertvanginkel I think this came up early in our discussion of the separation of concerns between `net/http` middleware and Connect interceptors.

envelope.go
```diff
@@ -227,6 +232,11 @@ func (r *envelopeReader) Read(env *envelope) *Error {
 		for remaining > 0 {
 			bytesRead, err := io.CopyN(env.Data, r.reader, remaining)
 			if err != nil && !errors.Is(err, io.EOF) {
+				situation := fmt.Sprintf("read %d byte message", size)
```
twilly, line 235: nit: you're possibly throwing out the construction of `situation`

akshayjshah: Meaning that we may be allocating a string but not using it?

maxbytes_go118.go
```diff
@@ -0,0 +1,28 @@
...
+//go:build !go1.19
+
+package connect
+
+import "strings"
+
+func asMaxBytesError(situation string, err error) *Error {
+	const expect = "http: request body too large"
+	text := err.Error()
+	if !(text == expect) && !strings.HasSuffix(text, ": "+expect) {
```
pkwarren, line 24:
```suggestion
	if text != expect && !strings.HasSuffix(text, ": "+expect) {
```

maxbytes_go118.go
```diff
@@ -0,0 +1,28 @@
...
+
+//go:build !go1.19
+
+package connect
+
+import "strings"
+
+func asMaxBytesError(situation string, err error) *Error {
+	const expect = "http: request body too large"
+	text := err.Error()
```
pkwarren, line 23: Any benefit to using errors.Is here instead or will we never expect it to be wrapped?

envelope.go
twilly: Yep! It's probably not important given this error path should be an uncommon event.

akshayjshah: Gotcha, good point - easy enough to fix!

maxbytes_go118.go
akshayjshah: This is embarrassingly obvious, I regret making you suggest this 😊

maxbytes_go118.go
akshayjshah: We can't use `errors.Is` on Go 1.18 - the `http.MaxByteError` type was introduced in Go 1.19.

pkwarren (approved):

PR #358, opened 2022-09-08
Add support for parsing charset set in Content-Type header

buildbreaker:
In some cases, libraries add `charset` as part of the `Content-Type` header (https://www.rfc-editor.org/rfc/rfc9110.html#media.type)

The change here is to avoid failing on requests which have the entry for `Content-Type: application/json; charset=utf-8` since we're currently doing an exact string look up

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -24,8 +24,7 @@ import (
 const (
 	codecNameProto           = "proto"
 	codecNameJSON            = "json"
-	codecCharsetUTF8         = "charset=utf-8"
-	codecNameJSONCharsetUTF8 = codecNameJSON + "; " + codecCharsetUTF8
+	codecNameJSONCharsetUTF8 = codecNameJSON + "; charset=utf-8"
 )
 
 // Codec marshals structs (typically generated from a schema) to and from bytes.
@@ -72,11 +71,13 @@ func (c *protoBinaryCodec) Unmarshal(data []byte, message any) error {
 	return proto.Unmarshal(data, protoMessage)
 }
 
-type protoJSONCodec struct{}
+type protoJSONCodec struct {
+	name string
+}
 
 var _ Codec = (*protoJSONCodec)(nil)
 
-func (c *protoJSONCodec) Name() string { return codecNameJSON }
+func (c *protoJSONCodec) Name() string { return c.name }
 
 func (c *protoJSONCodec) Marshal(message any) ([]byte, error) {
 	protoMessage, ok := message.(proto.Message)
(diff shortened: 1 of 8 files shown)


buildbreaker:
I'm going to need one of your help on this one 😂 

I'm not confident that this is the right place we'd want the tests to be but I'm also getting an interesting invalid memory dereference :(

handler.go
```diff
@@ -172,7 +173,12 @@ func (h *Handler) ServeHTTP(responseWriter http.ResponseWriter, request *http.Re
 	}
 
 	// Find our implementation of the RPC protocol in use.
-	contentType := request.Header.Get("Content-Type")
+	encodedContentType := request.Header.Get("Content-Type")
+	contentTypeList := strings.Split(encodedContentType, ";")
+	var contentType = encodedContentType
+	if len(contentTypeList) > 1 {
+		contentType = strings.TrimSpace(contentTypeList[0])
```
pkwarren, line 180: This might be a good place to use https://pkg.go.dev/mime#ParseMediaType instead.

protocol.go
```diff
@@ -85,7 +85,7 @@ type protocolHandler interface {
 	SetTimeout(*http.Request) (context.Context, context.CancelFunc, error)
 
 	// NewConn constructs a HandlerConn for the message exchange.
-	NewConn(http.ResponseWriter, *http.Request) (handlerConnCloser, bool)
+	NewConn(http.ResponseWriter, *http.Request, string) (handlerConnCloser, bool)
```
pkwarren, line 88: Updated this internal interface to pass the parsed media type (without params) down to the protocol handlers. This way we can re-use the parsed data, strip out parameters in the response content type, and properly look up the codec. Without this we'd have to repeat the parsing in the protocol implementations and they don't provide a handy way to return an error.

akshayjshah:
I agree with @timostamm - this is too lenient. We should immediately reject requests with Content-Types like `application/json; content-type=shift-jis` or `application/proto; endianness=middle`.

Can we instead use a combination of `mime.ParseMediaType` and `mime.FormatMediaType` to normalize the Content-Type in the request path? Then we can change the default configuration to register two JSON codecs: one for `application/json` and one for `application/json; charset=utf8`.

buildbreaker:
@akshayjshah Would this mean that users who want to support `application/json; charset=utf-8` that they would need to register an additional handler? The client/handler params only permit for registering a single codec

protocol.go
akshayjshah: Rather than introducing a new parameter here, why not overwrite the request's content-type header with the canonicalized value?

error_writer.go
```diff
@@ -94,7 +98,10 @@ func (w *ErrorWriter) IsSupported(request *http.Request) bool {
 //
 // Write does not read or close the request body.
 func (w *ErrorWriter) Write(response http.ResponseWriter, request *http.Request, err error) error {
-	ctype := request.Header.Get(headerContentType)
+	ctype, _, parseErr := mime.ParseMediaType(request.Header.Get(headerContentType))
+	if parseErr != nil {
+		return parseErr
+	}
```
akshayjshah, line 104: Two things:

1. Can we extract this repeated block into a utility function?
2. We should also be considering charset here - we shouldn't send UTF-8 if the client has requested `application/json; charset=shift-jis`.

codec.go
```diff
@@ -94,6 +96,28 @@ func (c *protoJSONCodec) Unmarshal(binary []byte, message any) error {
...
+
+func (c *protoJSONUTF8Codec) Name() string { return codecNameJSONCharsetUTF8 }
+
+func (c *protoJSONUTF8Codec) Marshal(message any) ([]byte, error) {
+	return c.jsonCodec.Marshal(message)
+}
+
+func (c *protoJSONUTF8Codec) Unmarshal(binary []byte, message any) error {
+	return c.jsonCodec.Unmarshal(binary, message)
+}
```
akshayjshah, line 119: Rather than introducing a new codec, can we have the existing codec keep the name as a field?

handler_ext_test.go
```diff
@@ -74,15 +74,34 @@ func TestHandler_ServeHTTP(t *testing.T) {
...
+			server.URL+pingProcedure,
+			strings.NewReader("{}"),
+		)
+		assert.Nil(t, err)
+		req.Header.Set("Content-Type", "application/json; charset=utf-8")
+		resp, err := client.Do(req)
+		assert.Nil(t, err)
+		defer resp.Body.Close()
+		assert.Equal(t, resp.StatusCode, http.StatusOK)
+	})
```
akshayjshah, line 103: Nice, appreciate the additional testing 😍

akshayjshah (approved):
I'm pushing a commit with a few small cleanups (noted in comments here). Otherwise, this looks much better! Thanks for bearing with comments from me & Timo.

(@pkwarren, especially appreciate you jumping in to help unblock Alan right away!)

PR #364, opened 2022-09-22
Expose peer information to handlers and clients

akshayjshah:
On requests and streams, expose the peer's address. For clients, the address is the host or host:port, parsed from the URL. For servers, the address is an IP:port pair, provided by `net/http` as `Request.RemoteAddr`.

In https://github.com/bufbuild/connect-go/issues/357, a handful of users asked for this information for logging, per-IP rate limiting, and a variety of other server-side concerns. It's also necessary for OpenTelemetry interceptors (https://github.com/bufbuild/connect-go/issues/344).

Fixes https://github.com/bufbuild/connect-go/issues/357.

### DIFF
--- a/connect.go
+++ b/connect.go
@@ -28,6 +28,7 @@ import (
 	"errors"
 	"io"
 	"net/http"
+	"net/url"
 )
 
 // Version is the semantic version of the connect module.
@@ -68,6 +69,7 @@ const (
 // StreamingHandlerConn implementations do not need to be safe for concurrent use.
 type StreamingHandlerConn interface {
 	Spec() Spec
+	Peer() Peer
 
 	Receive(any) error
 	RequestHeader() http.Header
@@ -97,8 +99,9 @@ type StreamingHandlerConn interface {
 // implementations must support limited concurrent use. See the comments on
 // each group of methods for details.
 type StreamingClientConn interface {
-	// Spec must be safe to call concurrently with all other methods.
+	// Spec and Peer must be safe to call concurrently with all other methods.
 	Spec() Spec
+	Peer() Peer
 
 	// Send, RequestHeader, and CloseRequest may race with each other, but must
 	// be safe to call concurrently with all other methods.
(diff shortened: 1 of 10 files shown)


akshayjshah:
@ElliotMJackson and @joshcarp, added you as reviewers in case you're interested or want to use this as a chance to start digging into `connect-go`. No pressure if you're still getting ramped up on other stuff.

protocol_connect.go
```diff
@@ -217,6 +220,10 @@ type connectClient struct {
 	protocolClientParams
 }
 
+func (c *connectClient) Peer() Peer {
+	return newPeerFromURL(c.URL)
```
pkwarren, line 224: This appears to be the URL passed to the client constructor. Do you think people would want to have this be set to the URL which was ultimately accessed (in case of redirects)?

connect.go
```diff
@@ -243,6 +253,21 @@ type Spec struct {
 	IsClient   bool   // otherwise we're in a handler
 }
 
+// Peer describes the other party to an RPC. When accessed client-side, Addr
+// contains the host or host:port from the server's URL. When accessed
+// server-side, Addr contains the client's address in IP:port format.
+type Peer struct {
+	Addr string
```
pkwarren, line 260: gRPC exposes this as `net.Addr` (which includes a Network string: https://pkg.go.dev/google.golang.org/grpc/peer#Peer). Do you think we won't need to provide that additional data?

protocol_connect.go
akshayjshah: Possibly? I'm not sure how to accomplish that, though, since we need this information _before_ we've attempted to make a request.

connect.go
akshayjshah: Possibly? We can always expose a `Network` attribute down the road too. I'm not sure how to reliably detect non-TCP transports through the net/http abstractions - for example, unix domain sockets are sort of the wild west, and I have no idea what HTTP/3 will end up looking like.

pkwarren (approved):

PR #374, opened 2022-10-13
Example of a possible fix for indicating server errors

### LINKED ISSUE #222: Clients can't differentiate server-sent and client-synthesized errors
Talked to @robbertvanginkel about this for a while today: in short, the problem is that clients can't tell whether an `*Error` was sent by the server, or whether it was synthesized within the client code (e.g., to wrap an underlying networking error). `grpc-go` has the same problem, and [documents the resulting complexity thoroughly](https://github.com/grpc/grpc/blob/master/doc/statuscodes.md).

I don't think we can differentiate based on status code, because many codes are semantically appropriate for both clients and servers.

We _could_ do something like this:

```go
type synthesizedError struct {
  err error
}

func (e *synthesizedError) Unwrap() error { return e.err }
func (e *synthesizedError) Error() string { return e.err.Error() }

func IsFromServer(err error) bool {
  if se := new(synthesizedError); errors.As(err, &se) {
    return false
  }
  return true
}
``` 

and then be careful to always wrap client-created errors in `synthesizedError`. Introducing this change would be backward-compatible, so let's defer deciding until we've released Connect and have some more user experience reports.

_Edit, months later_: it's probably simpler and much less error-prone to do the opposite of what I suggested above, and specially flag errors that are actually read off the wire.

joshcarp:
Add `connect.IsWireErr`, so clients can distinguish synthesized from server-sent errors.

Fixes #222.


error.go
```diff
@@ -100,17 +100,32 @@ func (d *ErrorDetail) Value() (proto.Message, error) {
...
+	meta        http.Header
+	isServerErr bool
 }
 
 // NewError annotates any Go error with a status code.
 func NewError(c Code, underlying error) *Error {
 	return &Error{code: c, err: underlying}
 }
 
+// IsServerErr returns true if the error was produced by the server.
```
akshayjshah, line 115: Let's rethink this name a little; `IsServerErr` suggests to me that we're categorizing errors into 4xx and 5xx. Perhaps `IsFromWire`?

Also, can we give some examples in the GoDoc? I think the idea is approachable, but a little subtle.

protocol_grpc.go
```diff
@@ -361,6 +361,7 @@ func (cc *grpcClientConn) Receive(msg any) error {
 		// means that we're _not_ getting a message. For users to realize that
 		// the stream has ended, Receive must return an error.
 		serverErr.meta = cc.responseHeader.Clone()
+		serverErr.isServerErr = true
```
akshayjshah, line 364: This misses the code path in line 347. Rather than setting `isServerErr` up at this level, we should do it in `grpcErrorFromTrailer` - we're much closer to the wire there.

protocol_connect.go
```diff
@@ -478,6 +479,7 @@ func (cc *connectStreamingClientConn) Receive(msg any) error {
 		// For users to realize that the stream has ended, Receive must return an
 		// error.
 		serverErr.meta = cc.responseHeader.Clone()
+		serverErr.isServerErr = true
```
akshayjshah, line 482: I think we can avoid both this change and the one above by handling this closer to the wire - let's add this code to `connectWireError.asError` instead.

akshayjshah (changes requested):
The overall approach looks good! Before we merge these changes in, let's get the code that sets `isServerErr` closer to the HTTP layer.

error.go
```diff
@@ -100,17 +100,32 @@ func (d *ErrorDetail) Value() (proto.Message, error) {
...
+	if !errors.As(err, &se) {
+		return false
+	}
+	return se.isServerErr
+}
+
+// IsServerErr returns true if the error was produced by the server.
+func (e *Error) IsServerErr() bool {
+	return e.isServerErr
+}
```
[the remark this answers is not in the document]
joshcarp: In that case the method isn't really needed because it's the same as inspecting the field

error.go
joshcarp: Although talking API design we probably want it to be `IsWireErr` because `connect.IsFromWire` doesn't give enough context

error.go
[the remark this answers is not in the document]
joshcarp: Removed IsServerErr, can be reintroduced if we ever need it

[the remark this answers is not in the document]
akshayjshah: Yep, sounds right to me!

akshayjshah (approved):

PR #394, opened 2022-11-15
Add RPC protocol to Peer

akshayjshah:
Currently, observability interceptors must parse the `Content-Type`
header to determine which RPC protocol is in use. This isn't awful, but
it's so easy for `connect-go` to expose directly that we might as well.

I chose to use our own strings rather than repurposing OpenTelemetry's
semantic conventions here. The OTel strings are less nice in this
context, and the packages are all unstable and subject to change.
@joshcarp, LMK if this seems like a bad tradeoff to you.

### DIFF
--- a/client_ext_test.go
+++ b/client_ext_test.go
@@ -97,8 +97,8 @@ func TestClientPeer(t *testing.T) {
 			_, closeErr := clientStream.CloseAndReceive()
 			assert.Nil(t, closeErr)
 		})
-		assert.NotNil(t, clientStream.Peer().Addr)
-		assert.NotNil(t, clientStream.Peer().Protocol)
+		assert.NotZero(t, clientStream.Peer().Addr)
+		assert.NotZero(t, clientStream.Peer().Protocol)
 		err = clientStream.Send(&pingv1.SumRequest{})
 		assert.Nil(t, err)
 		// server streaming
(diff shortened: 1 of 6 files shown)


client_ext_test.go
```diff
@@ -98,6 +98,7 @@ func TestClientPeer(t *testing.T) {
 			assert.Nil(t, closeErr)
 		})
 		assert.NotNil(t, clientStream.Peer().Addr)
+		assert.NotNil(t, clientStream.Peer().Protocol)
```
joshcarp, line 101:
```suggestion
		assert.NotEmpty(t, clientStream.Peer().Addr)
		assert.NotEmpty(t, clientStream.Peer().Protocol)
```
Addr + Protocol are strings so `assert.NotEmpty` is never going to fail

client_ext_test.go
```diff
@@ -113,6 +114,7 @@ func TestClientPeer(t *testing.T) {
 			assert.Nil(t, bidiStream.CloseResponse())
 		})
 		assert.NotNil(t, bidiStream.Peer().Addr)
+		assert.NotNil(t, bidiStream.Peer().Protocol)
```
joshcarp, line 117:
```suggestion
		assert.NotEmpty(t, bidiStream.Peer().Addr)
		assert.NotEmpty(t, bidiStream.Peer().Protocol)
```

joshcarp (approved):
I like it a lot

mattrobenolt:
So I actually already derived this on my own, so this is a nice addition I can use. Would it make sense to move it to it's own type/enum?

Something like:

```go
type Protocol string

const (
    ProtocolConnect = Protocol("connect")
...
)
```

or whatever? Might make it easier to use for folks trying to consume this (like me).

akshayjshah:
> Would it make sense to move it to it's own type/enum?

I'd rather not make a separate type, but I'm happy to do some enum-named constants!

From this codebase's perspective, I don't think named types are worth the API surface, since they don't stop us from mistakenly passing around string literals. I can see the utility of having them in code consuming these strings, but I'll leave that up to the consumer.

mattrobenolt:
I think a named constant would at least be useful. :)

PR #395, opened 2022-11-16
Bump test coverage to 90%

joshcarp:
Adds misc tests, almost to 90%

### DIFF
--- a/client_ext_test.go
+++ b/client_ext_test.go
@@ -146,6 +146,7 @@ func (a *assertPeerInterceptor) WrapStreamingClient(next connect.StreamingClient
 	return func(ctx context.Context, spec connect.Spec) connect.StreamingClientConn {
 		conn := next(ctx, spec)
 		assert.NotZero(a.tb, conn.Peer().Addr)
+		assert.NotEqual(a.tb, conn.Spec(), connect.Spec{})
 		return conn
 	}
 }
@@ -153,6 +154,7 @@ func (a *assertPeerInterceptor) WrapStreamingClient(next connect.StreamingClient
 func (a *assertPeerInterceptor) WrapStreamingHandler(next connect.StreamingHandlerFunc) connect.StreamingHandlerFunc {
 	return func(ctx context.Context, conn connect.StreamingHandlerConn) error {
 		assert.NotZero(a.tb, conn.Peer().Addr)
+		assert.NotEqual(a.tb, conn.Spec(), connect.Spec{})
 		return next(ctx, conn)
 	}
 }
--- a/client_ext_test.go
+++ b/client_ext_test.go
@@ -146,7 +146,7 @@ func (a *assertPeerInterceptor) WrapStreamingClient(next connect.StreamingClient
 	return func(ctx context.Context, spec connect.Spec) connect.StreamingClientConn {
 		conn := next(ctx, spec)
 		assert.NotZero(a.tb, conn.Peer().Addr)
-		assert.NotEqual(a.tb, conn.Spec(), connect.Spec{})
+		assert.NotZero(a.tb, conn.Spec())
 		return conn
 	}
 }
@@ -154,7 +154,7 @@ func (a *assertPeerInterceptor) WrapStreamingClient(next connect.StreamingClient
 func (a *assertPeerInterceptor) WrapStreamingHandler(next connect.StreamingHandlerFunc) connect.StreamingHandlerFunc {
 	return func(ctx context.Context, conn connect.StreamingHandlerConn) error {
 		assert.NotZero(a.tb, conn.Peer().Addr)
-		assert.NotEqual(a.tb, conn.Spec(), connect.Spec{})
+		assert.NotZero(a.tb, conn.Spec())
 		return next(ctx, conn)
 	}
 }
--- a/compression_test.go
+++ b/compression_test.go
@@ -16,6 +16,8 @@ package connect
 
 import (
 	"context"
+	"errors"
+	"io"
 	"net/http"
 	"net/http/httptest"
 	"testing"
@@ -53,4 +55,47 @@ func TestAcceptEncodingOrdering(t *testing.T) {
 	)
 	_, _ = client.CallUnary(context.Background(), NewRequest(&emptypb.Empty{}))
 	assert.True(t, called)
+}
+
+func TestFailCompression(t *testing.T) {
+	t.Parallel()
+	mux := http.NewServeMux()
+	server := httptest.NewServer(mux)
+	t.Cleanup(server.Close)
+	pingclient := NewClient[emptypb.Empty, emptypb.Empty](
+		server.Client(),
+		server.URL,
+		withFailCompression(),
+		WithSendCompression("error"),
+		WithCompressMinBytes(1),
+	)
+	t.Run("request_uncompressed", func(t *testing.T) {
+		t.Parallel()
+		_, err := pingclient.CallUnary(context.Background(), NewRequest(&emptypb.Empty{}))
+		assert.NotEqual(t, err, errors.New(""))
+	})
+}
+
+type failDecompressor struct {
+	Decompressor
+}
+
+type failCompressor struct{}
+
+func (failCompressor) Write([]byte) (int, error) {
+	return 0, errors.New("failCompressor")
+}
+
+func (failCompressor) Close() error {
+	return errors.New("failCompressor")
+}
+
+func (failCompressor) Reset(io.Writer) {}
+
+func withFailCompression() ClientOption {
+	return WithAcceptCompression(
+		"error",
+		func() Decompressor { return &failDecompressor{} },
+		func() Compressor { return &failCompressor{} },
+	)
 }
--- a/compression_test.go
+++ b/compression_test.go
@@ -57,25 +57,6 @@ func TestAcceptEncodingOrdering(t *testing.T) {
 	assert.True(t, called)
 }
 
-func TestFailCompression(t *testing.T) {
-	t.Parallel()
-	mux := http.NewServeMux()
-	server := httptest.NewServer(mux)
-	t.Cleanup(server.Close)
-	pingclient := NewClient[emptypb.Empty, emptypb.Empty](
-		server.Client(),
-		server.URL,
-		withFailCompression(),
-		WithSendCompression("error"),
-		WithCompressMinBytes(1),
-	)
-	t.Run("request_uncompressed", func(t *testing.T) {
-		t.Parallel()
-		_, err := pingclient.CallUnary(context.Background(), NewRequest(&emptypb.Empty{}))
-		assert.NotEqual(t, err, errors.New(""))
-	})
-}
-
 type failDecompressor struct {
 	Decompressor
 }
@@ -92,10 +73,12 @@ func (failCompressor) Close() error {
 
 func (failCompressor) Reset(io.Writer) {}
 
-func withFailCompression() ClientOption {
-	return WithAcceptCompression(
-		"error",
-		func() Decompressor { return &failDecompressor{} },
-		func() Compressor { return &failCompressor{} },
-	)
+func WithErrorCompression() Option {
+	return &compressionOption{
+		Name: "error",
+		CompressionPool: newCompressionPool(
+			func() Decompressor { return &failDecompressor{} },
+			func() Compressor { return &failCompressor{} },
+		),
+	}
 }
--- a/compression_test.go
+++ b/compression_test.go
@@ -16,8 +16,6 @@ package connect
 
 import (
 	"context"
-	"errors"
-	"io"
 	"net/http"
 	"net/http/httptest"
 	"testing"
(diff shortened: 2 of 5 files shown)


compression_test.go
```diff
@@ -54,3 +56,46 @@ func TestAcceptEncodingOrdering(t *testing.T) {
...
+		server.Client(),
+		server.URL,
+		withFailCompression(),
+		WithSendCompression("error"),
+		WithCompressMinBytes(1),
+	)
+	t.Run("request_uncompressed", func(t *testing.T) {
+		t.Parallel()
+		_, err := pingclient.CallUnary(context.Background(), NewRequest(&emptypb.Empty{}))
+		assert.NotEqual(t, err, errors.New(""))
```
pkwarren, line 75: Can we add a better assertion here (or just use `assert.Error`)?

connect_ext_test.go
```diff
@@ -543,6 +543,51 @@ func TestTimeoutParsing(t *testing.T) {
...
+	t.Parallel()
+	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {})
+	server := httptest.NewServer(handler)
+	defer server.Close()
+	client := pingv1connect.NewPingServiceClient(
+		server.Client(),
+		server.URL,
+	)
+	wantErr := error(connect.NewError(connect.CodeCanceled, errors.New("fail")))
+	ctx := errorContext{Context: context.Background(), Error: wantErr}
```
pkwarren, line 574: Instead of a new type, do you just want to test calling the method with a previously canceled context (the more likely issue)?

connect_ext_test.go
```diff
@@ -543,6 +543,51 @@ func TestTimeoutParsing(t *testing.T) {
...
+	stream := client.CumSum(ctx)
+	err := stream.Send(nil)
+	var connectErr *connect.Error
+	assert.NotNil(t, err)
+	assert.True(t, errors.As(err, &connectErr))
+	assert.Equal(t, connectErr.Code(), connect.CodeCanceled)
+
+	wantErr = errors.New("fail")
+	ctx = errorContext{Context: context.Background(), Error: wantErr}
```
pkwarren, line 583: This is where we'd probably want to test deadline exceeded (in addition to canceled). The context type is an interface but it is documented to either return Canceled or DeadlineExceeded: https://pkg.go.dev/context#Context

connect_ext_test.go
```diff
@@ -699,6 +744,42 @@ func TestBidiRequiresHTTP2(t *testing.T) {
...
+		)
+		return pingclient, &contentEncoding
+	}
+	t.Run("request_uncompressed", func(t *testing.T) {
+		t.Parallel()
+		pingclient, contentEncoding := serve()
+		_, _ = pingclient.Ping(context.Background(), connect.NewRequest(&pingv1.PingRequest{
+			Text: "ping",
+		}))
+		assert.Equal(t, *contentEncoding, "")
```
pkwarren, line 771: nit: `assert.Empty(...)`

connect_ext_test.go
```diff
@@ -699,6 +744,42 @@ func TestBidiRequiresHTTP2(t *testing.T) {
 	)
 }
 
+func TestCompressMinBytesClient(t *testing.T) {
+	t.Parallel()
+	serve := func() (pingv1connect.PingServiceClient, *string) {
+		mux := http.NewServeMux()
+		var contentEncoding string
+		go mux.Handle("/", http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
+			contentEncoding = request.Header.Get("Content-Encoding")
```
pkwarren, line 753: nit: This feels a bit racy to use a `*string` here and rely on the handler being called (without using sync.WaitGroup or other things to ensure things have completed).

Once option might to be to return a function which takes a PingRequest (or a string) and returns a PingResponse and content encoding. Then you can ensure the concurrency is sane and potentially also minimize duplication in the tests below.

compression_test.go
```diff
@@ -54,3 +56,29 @@ func TestAcceptEncodingOrdering(t *testing.T) {
...
+	return 0, errors.New("failCompressor")
+}
+
+func (failCompressor) Close() error {
+	return errors.New("failCompressor")
+}
+
+func (failCompressor) Reset(io.Writer) {}
+
+func WithErrorCompression() Option {
```
joshcarp, line 76: This is exported to be tested in `connect_ext_test.go`.

connect_ext_test.go
joshcarp: Yes that would work. 
Technically I was also testing returning a `connect.Erorr` but that would never happen irl so I've removed that

client_ext_test.go
```diff
@@ -146,13 +146,15 @@ func (a *assertPeerInterceptor) WrapStreamingClient(next connect.StreamingClient
 	return func(ctx context.Context, spec connect.Spec) connect.StreamingClientConn {
 		conn := next(ctx, spec)
 		assert.NotZero(a.tb, conn.Peer().Addr)
+		assert.NotEqual(a.tb, conn.Spec(), connect.Spec{})
```
akshayjshah, line 149:
```suggestion
		assert.NotZero(a.tb, conn.Spec())
```

client_ext_test.go
```diff
@@ -146,13 +146,15 @@ func (a *assertPeerInterceptor) WrapStreamingClient(next connect.StreamingClient
...
 		return conn
 	}
 }
 
 func (a *assertPeerInterceptor) WrapStreamingHandler(next connect.StreamingHandlerFunc) connect.StreamingHandlerFunc {
 	return func(ctx context.Context, conn connect.StreamingHandlerConn) error {
 		assert.NotZero(a.tb, conn.Peer().Addr)
+		assert.NotEqual(a.tb, conn.Spec(), connect.Spec{})
```
akshayjshah, line 157:
```suggestion
		assert.NotZero(a.tb, conn.Spec())
```

compression_test.go
akshayjshah: I think we should be able to put this in `connect_ext_test.go` and unexport it:

```
func withErrorCompression() connect.Option {
  return connect.WithCompression(
    "error",
    func() connect.Decompressor { return &failDecompressor{} },
    func() connect.Compressor { return &failCompressor{} },
  )
}
```

connect_ext_test.go
```diff
@@ -543,6 +543,43 @@ func TestTimeoutParsing(t *testing.T) {
...
+		server.URL,
+	)
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+	stream := client.CumSum(ctx)
+	err := stream.Send(nil)
+	var connectErr *connect.Error
+	assert.NotNil(t, err)
+	assert.True(t, errors.As(err, &connectErr))
+	assert.Equal(t, connectErr.Code(), connect.CodeCanceled)
```
akshayjshah, line 580: Can we add an `assert.False(t, connect.IsWireError(err))` too? The client should be short-circuiting and returning an error before even writing anything to the network.

connect_ext_test.go
```diff
@@ -699,6 +736,43 @@ func TestBidiRequiresHTTP2(t *testing.T) {
...
+		server := httptest.NewServer(mux)
+		t.Cleanup(server.Close)
+		pingclient := pingv1connect.NewPingServiceClient(
+			http.DefaultClient, server.URL,
+			connect.WithSendGzip(),
+			connect.WithCompressMinBytes(8),
+		)
+		_, err := pingclient.Ping(context.Background(), connect.NewRequest(request))
+		return contentEncoding, err
+	}
```
akshayjshah, line 757: Love that we're testing this! Even with the updates, though, the approach seems a little convoluted. I think we can simplify a bit by having the helper function take the expected encoding as an argument and make the assertions:

```go
func assertContentType = func(tb testing.TB, text, expect string) {
  tb.Helper()
  mux := http.NewServeMux()
  mux.Handle("/", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
    assert.Equal(tb, request.Header.Get("Content-Encoding"), expect)
  })
  server := httptest.NewServer(mux)
  tb.Cleanup(server.Close)
  _, err := pingv1connect.NewPingServiceClient(
    server.Client,
    server.URL,
    connect.WithSendGzip(),
    connect.WithCompressMinBytes(8),
  ).Ping(context.Background(), connect.NewRequest(&pingv1.PingRequest{Text: text})
  assert.Nil(tb, err)
}
t.Run("request_uncompressed", func(t *testing.T) {
    assertContentType(t, "ping", "")
})
t.Run("request_compressed", func(t *testing.T) {
    assertContentType(t, "pingping", "gzip")
})
```

connect_ext_test.go
```diff
@@ -1393,6 +1467,305 @@ func TestBidiStreamServerSendsFirstMessage(t *testing.T) {
...
+			},
+		})
+		t.Cleanup(server.Close)
+		stream := client.CumSum(context.Background())
+		assert.Nil(t, stream.Send(nil))
+		_, err := stream.Receive()
+		assert.NotNil(t, err)
+		assert.Equal(t, connect.CodeOf(err), connect.CodeInternal)
+		assert.Nil(t, stream.CloseRequest())
+		t.Parallel()
```
akshayjshah, line 1498: Here and in the subsequent subtests, I think the `t.Parallel()` should probably be at the top.

connect_ext_test.go
```diff
@@ -1393,6 +1467,305 @@ func TestBidiStreamServerSendsFirstMessage(t *testing.T) {
...
 
+func TestStreamForServer(t *testing.T) {
+	t.Parallel()
+	newPingServer := func(pingServer pingv1connect.PingServiceHandler) (pingv1connect.PingServiceClient, *httptest.Server) {
+		mux := http.NewServeMux()
+		mux.Handle(pingv1connect.NewPingServiceHandler(pingServer))
+		server := httptest.NewUnstartedServer(mux)
+		server.EnableHTTP2 = true
+		server.StartTLS()
+		t.Cleanup(server.Close)
```
akshayjshah, line 1478: The subtests all call `t.Cleanup(server.Close)` with their sub-t's, so we don't need this here.

connect_ext_test.go
```diff
@@ -1393,6 +1467,305 @@ func TestBidiStreamServerSendsFirstMessage(t *testing.T) {
...
+		res, err := stream.CloseAndReceive()
+		assert.NotNil(t, err)
+		assert.Equal(t, connect.CodeOf(err), connect.CodeUnknown)
+		assert.Nil(t, res)
+	})
+}
+
+func TestConnectHTTPErrorCodes(t *testing.T) {
+	t.Parallel()
+	checkHTTPStatus := func(t *testing.T, connectErr connect.Code, wantHttpStatus int) {
```
akshayjshah, line 1643:
```suggestion
	checkHTTPStatus := func(t *testing.T, connectCode connect.Code, wantHttpStatus int) {
```

connect_ext_test.go
```diff
@@ -1393,6 +1467,305 @@ func TestBidiStreamServerSendsFirstMessage(t *testing.T) {
...
+		assert.Nil(t, err)
+		req.Header.Set("Content-Type", "application/json")
+		resp, err := server.Client().Do(req)
+		assert.Nil(t, err)
+		assert.Equal(t, wantHttpStatus, resp.StatusCode)
+		defer resp.Body.Close()
+		connectClient := pingv1connect.NewPingServiceClient(server.Client(), server.URL)
+		connectResp, err := connectClient.Ping(context.Background(), connect.NewRequest(&pingv1.PingRequest{}))
+		assert.NotNil(t, err)
+		assert.Nil(t, connectResp)
```
akshayjshah, line 1669: Do these last four lines test anything that's not covered elsewhere?

interceptor_ext_test.go
```diff
@@ -139,6 +139,33 @@ func TestOnionOrderingEndToEnd(t *testing.T) {
...
+	assert.Nil(t, sumStream.Send(&pingv1.SumRequest{Number: 1}))
+	resp, err := sumStream.CloseAndReceive()
+	assert.Nil(t, err)
+	assert.NotNil(t, resp)
+	countUpStream, err := connectClient.CountUp(context.Background(), connect.NewRequest(&pingv1.CountUpRequest{}))
+	assert.Nil(t, err)
+	for countUpStream.Receive() {
+		assert.NotNil(t, countUpStream.Msg())
+	}
+	defer countUpStream.Close()
```
akshayjshah, line 166: We're already at the end of the function, so the defer isn't doing anything. Should this just be `assert.Nil(countUpStream.Close()`?

akshayjshah (approved):
Awesome :tada: Thank you for taking this on, @joshcarp!

Once you make the suggested changes and tests/lint/etc are passing, go ahead and merge this.

connect_ext_test.go
joshcarp: copypasta issue 🙃

connect_ext_test.go
joshcarp: Yes actually; the client panics when the server sends a zero connect error code which has an open issue at the moment: https://github.com/bufbuild/connect-go/issues/396

You can't get the http status code from a connect unary client (that I know of) so doing both was the only way I came up with

connect_ext_test.go
joshcarp: Yeah this makes a lot more sense than what's there

PR #398, opened 2022-11-18
Fix panic on zero send from server

joshcarp:
- Maps a zero code of `connectWireError` to a CodeUnknown in `connectWireError) asError() *Error `
This could have been done in the unmarshaling to `connectWireError` but I chose to leave that to be exactly what was read off the wire. 
- Adds nil guard in `validateResponse` for asError
- Adds tests for error mapping

Fixes: https://github.com/bufbuild/connect-go/issues/396

### DIFF
--- a/client_ext_test.go
+++ b/client_ext_test.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"testing"
 
 	"github.com/bufbuild/connect-go"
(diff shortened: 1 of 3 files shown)


protocol_connect.go
```diff
@@ -919,9 +922,12 @@ func newConnectWireError(err error) *connectWireError {
 }
 
 func (e *connectWireError) asError() *Error {
-	if e == nil || e.Code == 0 {
+	if e == nil {
 		return nil
 	}
+	if e.Code == 0 {
```
pkwarren, line 928: Does this need to map other invalid codes to unknown besides zero?

client_ext_test.go
```diff
@@ -133,6 +134,110 @@ func TestClientPeer(t *testing.T) {
...
+			http.MethodPost,
+			server.URL+"/"+pingv1connect.PingServiceName+"/Ping",
+			strings.NewReader("{}"),
+		)
+		assert.Nil(t, err)
+		req.Header.Set("Content-Type", "application/json")
+		resp, err := server.Client().Do(req)
+		assert.Nil(t, err)
+		assert.Equal(t, wantHttpStatus, resp.StatusCode)
+		defer resp.Body.Close()
```
akshayjshah, line 161: nit: move this up a line, so we clean up resources even if the status code assertion fails.

protocol_connect.go
akshayjshah: Yes. We should map anything outside the known codes to `CodeUnknown`.

akshayjshah (approved):
Let's be more aggressive about mapping out-of-range codes to `CodeUnknown`. Otherwise, LGTM - thanks for the fix Josh!

PR #406, opened 2022-11-28
Ensure streaming ResponseWriters implement Flush

### LINKED ISSUE #393: Feature: Return error if ResponseWriter does not implement Flush() for streaming RPCs.
If you wrap your `http.ResponseWriter` in middleware before handing to connect go, you may wind up not exposing the `http.Flusher` interface, leading to calls to `Send` not actually flushing the `http.ResponseWriter` buffer and thus the connection appearing to "hang".

This problem is tricky to detect and debug because it often works correctly; it only fails in some cases when using streaming RPCs. In an effort to make it harder to footgun oneself, I request that connect-go detect this situation and return an error from a streaming `Send` call.

Instead of returning an error, one could panic: this would make it much more noticeable. However, it is unclear if this is worse or better than unexpectedly hanging in production. I think returning an error that can be handled is at least obviously better. One could also make the case for some kind of warning message, but I'm not sure if there's a good way to do that (and perhaps, it's just not loud enough.)

**Additional context**
Apparently, the buf cli ran into this issue too, when migrating to Connect. It's a common footgun in Go. Maybe Go itself could be improved? I left some thoughts on golang/go#5465.

akshayjshah:
The `http.Flusher` interface isn't part of the `http.ResponseWriter`
interface, but it's implemented by the HTTP/1 and HTTP/2 writers
provided by the standard library. It's also critical to Connect:
streaming handlers must flush messages to ensure that the client
receives them. This PR adds a check for unflushable response writers;
when the response writer for a streaming procedure can't be flushed, we
now return an explicit error instead of hanging.

Fixes #393.

### DIFF
--- a/connect_ext_test.go
+++ b/connect_ext_test.go
@@ -2124,6 +2124,9 @@ func (l *trimTrailerWriter) Flush() {
 }
 
 func (l *trimTrailerWriter) removeTrailers() {
+	for _, v := range l.w.Header().Values("Trailer") {
+		l.w.Header().Del(v)
+	}
 	l.w.Header().Del("Trailer")
 	for k := range l.w.Header() {
 		if strings.HasPrefix(k, http.TrailerPrefix) {
--- a/connect_ext_test.go
+++ b/connect_ext_test.go
@@ -1774,6 +1774,61 @@ func TestFailCompression(t *testing.T) {
 	assert.Equal(t, connect.CodeOf(err), connect.CodeInternal)
 }
 
+func TestUnflushableResponseWriter(t *testing.T) {
+	assertIsFlusherErr := func(t *testing.T, err error) {
+		t.Helper()
+		assert.NotNil(t, err)
+		assert.Equal(t, connect.CodeOf(err), connect.CodeInternal, assert.Sprintf("got %v", err))
+		assert.True(
+			t,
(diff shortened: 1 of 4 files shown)


jchadwick-buf (approved):
LGTM! Thanks for the quick follow-up.

joshcarp:
I don't quite understand why this needs to be checked within the handler itself; wouldn't it be safer if it failed at creation of the `connect.NewServerStreamHandler` or `connect.NewBidiStreamHandler` and fail at startup instead?

joshcarp:
whoops, apologies

joshcarp (approved):

akshayjshah:
I'm not sure why the build is failing on Windows :/ I won't be able to dig into it until late tonight (at the earliest) - if anyone else has time and is interested, please take a look! (Just ping this thread so nobody else picks it up at the same time.)

pkwarren:
> I'm not sure why the build is failing on Windows :/ I won't be able to dig into it until late tonight (at the earliest) - if anyone else has time and is interested, please take a look! (Just ping this thread so nobody else picks it up at the same time.)

@akshayjshah - Pushed a fix which should resolve the test failure. The `testTrailerWriter` wasn't deleting keys of trailers set with the `Trailer` header.

PR #407, opened 2022-11-28
Format imports with GCI

akshayjshah:
`golangci-lint` includes `gci` by default, so we can disable `goimports`. The auto-fix function still works, so `make lintfix` is enough to reformat any files with incorrectly grouped imports.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -36,6 +36,7 @@ linters:
     - funlen            # rely on code review to limit function length
     - gocognit          # dubious "cognitive overhead" quantification
     - gofumpt           # prefer standard gofmt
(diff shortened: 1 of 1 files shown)


joshcarp (approved):

pkwarren (approved):

bufdev:
As discussed in core PR, enforcement without automation of import sorting via make generate or typical IDE setups means more time for our devs on a consistent basis, with little to no benefit to us over removing this item from the style guide

akshayjshah:
I hear the points you're making on the core PR. I think you may be misunderstanding the proposed change (both there and here), the contents of this diff, and its intrusiveness on engineers' work.

PR #409, opened 2022-12-01
add API to make it easier to work with ErrorDetail if you have/want an Any

jhump:
If your code already has an instance of `Any`, it's a pain (and not even guaranteed to be possible at runtime!) to unmarshal the message therein only to re-wrap it in order to put it in an `ErrorDetail`.

Similarly, if you can't unmarshal the contained type, it may be useful to access the `Any` envelope.

This adds two new APIs for just this.

### DIFF
--- a/error.go
+++ b/error.go
@@ -44,8 +44,9 @@ type ErrorDetail struct {
 	wireJSON string // preserve human-readable JSON
 }
 
-// NewErrorDetail constructs a new error detail. This returns an error
-// if msg cannot be marshalled into an Any message.
+// NewErrorDetail constructs a new error detail. If msg is an *[anypb.Any] then
+// it is used as is. Otherwise, it is first marshaleled into an *[anypb.Any]
+// value. This returns an error if msg cannot be marshalled.
 func NewErrorDetail(msg proto.Message) (*ErrorDetail, error) {
 	// If it's already an Any, don't wrap it inside another.
 	if pb, ok := msg.(*anypb.Any); ok {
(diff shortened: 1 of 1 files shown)


jhump:
@akshayjshah, what do you think?

I actually encountered this when I was trying to formulate adapters from gRPC generated interfaces to Connect. In those cases, I needed to convert a gRPC `*status.Error` into a `*connect.Error`, and vice versa. Without methods like this, there's no easier way to do that.

Also, this came up in public Slack: https://bufbuild.slack.com/archives/CRZ680FUH/p1669918357272869

error.go
```diff
@@ -44,8 +44,13 @@ type ErrorDetail struct {
 	wireJSON string // preserve human-readable JSON
 }
 
-// NewErrorDetail constructs a new error detail.
+// NewErrorDetail constructs a new error detail. This returns an error
+// if msg cannot be marshalled into an Any message.
```
akshayjshah, line 48: A surprising number of people aren't familiar with the WKTs. Can we turn `Any` into a newfangled GoDoc link to the docs for `*anypb.Any`? I think `[google.golang.org/protobuf/types/known/anypb.Any]` works.

Also, can we document that this doesn't re-wrap if you pass an Any?

akshayjshah (approved):
LGTM with the smaller surface area! Some docs suggestions, but no need for more review.

akshayjshah:
Happy to cut a release as soon as this is merged, too!

error.go
jhump: Done. I've used simple type links, so I changed `Any` to `[anypb.Any]`.

An alternative would be to write it as `[Any]` and then add a footnote:
```
[Any]: https://pkg.go.dev/google.golang.org/protobuf/types/known/anypb#Any
```
Would you prefer that alternative? ([reference](https://tip.golang.org/doc/comment#links))

PR #410, opened 2022-12-02
Send gRPC error metadata only as HTTP trailers

akshayjshah:
In #406, I was too cavalier about `net/http.ResponseWriter`'s API to
send HTTP trailers. Because we're not calling `WriteHeader` before
between predeclaring our trailers and actually setting the trailer
values, we end up sending the same data as both HTTP headers and
trailers.

### DIFF
--- a/connect_ext_test.go
+++ b/connect_ext_test.go
@@ -19,6 +19,7 @@ import (
 	"compress/flate"
 	"compress/gzip"
 	"context"
+	"encoding/binary"
 	"errors"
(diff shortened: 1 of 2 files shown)


pkwarren (approved):

rhbuf (approved):
I found [an example](https://pkg.go.dev/net/http#example-ResponseWriter-Trailers) where they illustrate this behaviour.

lgtm after ci fixes

PR #418, opened 2022-12-08
Import base go file even if there are no basetypes required 

joshcarp:
If a service solely requires types from other packages the result is that the connect services don't have access to the filedescriptorsets. This causes a lot of issues such as server reflection failing. 

This PR blank imports the base go file that should always exist

For the proto file:

```proto
// service/eliza.proto
syntax = "proto3";
package service;
import "models/eliza.proto";

service ElizaService {
  rpc Say(ext.SayRequest) returns (ext.SayResponse) {}
}

// models/eliza.proto
syntax = "proto3";
package ext;

message SayRequest {
  string sentence = 1;
}
message SayResponse {
  string sentence = 1;
}
```

The following diff occurs in the generated `eliza.connect.go` code
```diff
package serviceconnect

import (
	context "context"
	errors "errors"
	models "github.com/bufbuild/connect-demo/internal/gen/buf/connect/demo/eliza/v1/models"
+	_ "github.com/bufbuild/connect-demo/internal/gen/buf/connect/demo/eliza/v1/service"
	connect_go "github.com/bufbuild/connect-go"
	http "net/http"
	strings "strings"
)

```

There is no change if the base types are already in the `service/eliza.proto` (the way we recommend using proto files)

Closes TCN-786


pkwarren:
Change looks good to me - should we regenerate the code in connect-go with this plugin?

Not sure if it is worth it, but we might consider creating a small example for this under internal/proto and a test which verifies the global registry contains the service and all messages after importing the generated connect code.

akshayjshah (commented):
LGTM - nice fix, @joshcarp! As @pkwarren suggested, please do add a test case to prevent us from breaking this again in the future.

pkwarren (approved):
nice work!

Issue #380, opened 2022-10-21
document in the FAQ how client and server might access authenticated identity of the remote


jhump:
The `Peer` object only provides an address. For clients, it just echos back the host that was used in the HTTP request, without returning anything about the remote server's identity. It could at least return the resolved IP, to provide more information when multiple A records are available.

But it also provides no way to get the _authenticated_ identity, when using TLS. This is easily available in the response object returned from the `http.Client` or the request object provided to the `http.Handler`.

The [gRPC version of this type](https://pkg.go.dev/google.golang.org/grpc/peer#Peer) has a generic `AuthInfo` field with an interface type, and users can then try to type assert to a specific implementation. The idea is that different authn mechanisms might be used to authenticate parties (like JWTs or other custom auth cookies for client authn), so the representation needs to be flexible enough so an interceptor could provide the identity (instead of hard-coding the representation used in mutually-authenticated TLS). Speaking of which, there is not a way for an authenticating interceptor to override the peer since there is no exported setter (or constructor which allows setting it).

jhump:
I take it back about an interceptor not being able to set the Peer. The interceptor uses interfaces to model requests, responses, and streams, so it could provide its own implementation that overrides the `Peer()` method.

akshayjshah:
On the client side, users can already use `net/http/httptrace`. On the server side, users can employ one of the _many_ authn/authz packages that wrap `http.Handler` - and if they like, they can attach the authenticated identity to the context.

If we're going to add more API surface to Connect, we should provide something notably better than these two options. What could we do?

On the client side, we have access to a `tls.ConnState` on the HTTP response. We could also use `httptrace` to collect some info automatically. We could use one or both of those mechanisms to include some additional information in `connect.AnyResponse`. But since `connect.AnyResponse` is shared between the client and server, that makes the server more awkward - we'd end up with an API that just returns nil there. In general, I haven't seen much demand for this sort of thing on Slack or in other Github issues.

I do see more demand for server-side authentication. That said, I'm not sure what we get from a gRPC-style API. The gRPC `authinfo.AuthInfo` interface doesn't provide any useful capabilities, and the suggestion that implementations embed `authinfo.CommonAuthInfo` only provides the odd `authinfo.Privacy` enum. In short, this seems like a bunch of types that accomplish little beyond `context.Value`. Without a more substantive `AuthInfo` interface, I'm not sure how we can do any better.

[...]

PR #422, opened 2022-12-19
Expose *http.Request to server Peer

ImSingee:
In some cases, we need access to the underlying *http.Request in the handler.

For example, we need access to the underlying request.TLS to identify peer identity.

### DIFF
--- a/client_ext_test.go
+++ b/client_ext_test.go
@@ -116,7 +116,7 @@ func TestClientPeer(t *testing.T) {
 		})
 		assert.NotZero(t, bidiStream.Peer().Addr)
 		assert.NotZero(t, bidiStream.Peer().Protocol)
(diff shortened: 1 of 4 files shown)


akshayjshah (changes requested):
Hey @ImSingee - thanks for opening your first PR here!

I don't think we want to expose the full HTTP request - it's much more than #380 requires. It's also error-prone, because consuming data from the body is very likely to break further stream decoding.

I'd prefer to hold off on merging anything like this until the discussion on #380 comes to a firmer conclusion. At most, we'd expose the TLS connection state on `Peer`.

akshayjshah:
@ImSingee Appreciate the contribution, but I'm going to close this issue until we decide on a path forward in the linked issue.

PR #423, opened 2022-12-19
Don't clobber custom User-Agents

akshayjshah:
If the caller sets a User-Agent manually, we shouldn't clobber it. This
should unblock some internal use cases where people want to use
User-Agents to identify internal services, environments, etc.


akshayjshah:
cc @mattrobenolt and @mdlayher

mdlayher (approved):
LGTM

pkwarren (approved):

joshcarp (approved):

PR #435, opened 2023-01-05
Allow handlers to implement a simpler interface

akshayjshah:
This PR resurrects a simpler version of [a feature we had earlier in `connect-go` development](https://github.com/bufbuild/connect-go/blob/fa11ba1c45213d5f716eec92af4a019b38f766af/internal/gen/proto/go-rerpc/rerpc/ping/v1test/ping_rerpc.pb.go#L357): it lets handlers that don't need access to headers or trailers implement a simpler interface (without the generic `connect.Request` or `connect.Response` structs). For unary methods, the simpler interface matches `grpc-go`'s generated code.

This came up in https://github.com/bufbuild/connect-go/discussions/421. The argument isn't primarily ergonomics or ease of migration, but that the business logic in a [hexagonal architecture](https://8thlight.com/insights/a-color-coded-guide-to-ports-and-adapters)-style application shouldn't be visibly coupled to `connect-go`.

For now, I've just implemented a proof of concept for handlers. If we actually move forward with this, we'd want something similar for clients. Looking at this code, I still feel that this doesn't belong in `protoc-gen-connect-go` - I'd rather have a completely separate plugin that works with these stripped-down interfaces. We'd probaby start with something community-supported instead of putting another binary into this repo.

@bufdev and @joshcarp, what do you think?

### DIFF
--- a/cmd/protoc-gen-connect-go/main.go
+++ b/cmd/protoc-gen-connect-go/main.go
@@ -162,6 +162,7 @@ func generateService(g *protogen.GeneratedFile, service *protogen.Service) {
 	generateServerInterface(g, service, names)
 	generateServerConstructor(g, service, names)
 	generateUnimplementedServerImplementation(g, service, names)
+	generateSimpleServerConstructor(g, service, names)
 }
 
 func generateClientInterface(g *protogen.GeneratedFile, service *protogen.Service, names names) {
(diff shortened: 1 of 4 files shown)


joshcarp:
Overall I really like this solution; One thing I found even about gRPC interface implementations with testing is that a lot of the time you'd need to still need a lot of setup of the context otherwise nil pointer errors arise everywhere if the implementation tried to access metadata/ctx vars injected by interceptors etc. This makes it nice and explicit that you can't access any of that, so if these interfaces are implemented you'll be able to test it without any other ctx injection/setup.

On the other hand having more than 1 way to do something isn't great, but imo the tradeoff between more code generated vs the added benefit of easy testing and optimizing for the (possibly?) more common case is worth it.

thesiti92:
this pr i think would allow putting grpc-gateway in front of a connect-go server and make calls directly to the server handler functions instead of having to make local requests. We need to use grpc-gateway to have more configurable request urls, file downloads, etc...

PR #436, opened 2023-01-05
Make top-level Handler example runnable

### LINKED ISSUE #433: pkg.go.dev example doesn't compile
**Describe the bug**

The example doesn't work here:
https://pkg.go.dev/github.com/bufbuild/connect-go#example-package-Handler

Fails with the error
```
Output:

go: downloading github.com/bufbuild/connect-go v1.4.1
go: downloading google.golang.org/protobuf v1.28.1
package play.ground
	prog.go:8:2: use of internal package github.com/bufbuild/connect-go/internal/gen/connect/ping/v1 not allowed
package play.ground
	prog.go:9:2: use of internal package github.com/bufbuild/connect-go/internal/gen/connect/ping/v1/pingv1connect not allowed
```
The solution here would be either
- Move protos from internal to something that isn't restrictive like `private`
- Create a BSR module and remove internal/gen completely
- Create another example in https://github.com/bufbuild/connect-go/blob/main/handler_example_test.go so that it is not rendered as a runnable example

From the docs [here](https://go.dev/blog/examples)

> To achieve this we can use a “whole file example.” A whole file example is a file that ends in _test.go and contains exactly one example function, no test or benchmark functions, and at least one other package-level declaration. When displaying such examples godoc will show the entire file.

akshayjshah:
The code in `handler_example_test.go` uses internal packages, so it
doesn't run on `pkg.go.dev`. (When pkgsite converts the example to a
`main` function, it's no longer in the `connect-go` module.)

This PR fixes the problem by switching from a whole-file example to a
one-function example. **On balance, I think this is worse** - it hides the
service _implementation_, which is more important than making the
example runnable.

@joshcarp, what do you think?

Fixes #433.


joshcarp:
Not in love with this. I think an important part of showing a service example is showing the look of the interface that you're going to need to implement (the actual running of the server is less important imo)

akshayjshah:
Agree, will close :)

PR #437, opened 2023-01-13
Add NewWireError

### LINKED ISSUE #420: expose an API for constructing a wire error
When using `connect.NewError`, it is not possible create an error such that `err.IsWireError()` will return true.

This capability is particularly useful for dealing with bidi streaming APIs that represent a conversation, where each response message correlates with a request message. It is common for these kinds of APIs to support partial success by having a response message indicate a failure just for that one request message. Often such APIs use `google.rpc.Status` in the response message to indicate a possible failure. Sometimes they [do not](https://github.com/grpc/grpc/blob/v1.49.2/src/proto/grpc/reflection/v1alpha/reflection.proto#L132-L136).

For this sort of use, it is commonplace for the calling Go code to translate that status into an error. Since the error was received over the wire during an RPC call, it should technically be a wire error, so that any other handling/propagation logic higher up the stack can react to it correctly.

This could be addressed as easily as introducing a new `NewWireError` function with the same signature and semantics as `NewError` except that the returned error will return true when `err.IsWireError()` is called.

akshayjshah:
Currently, only the Connect runtime can create errors that return true
when tested by `IsWireError`. However, streaming RPCs commonly indicate
partial failure by including an error code and message in their response
message. To propagate these partial failures properly, clients need a
way to construct wire errors. This commit adds `NewWireError` to satisfy
this use case.

Fixes #420.

### DIFF
--- a/error.go
+++ b/error.go
@@ -118,6 +118,25 @@ func NewError(c Code, underlying error) *Error {
 	return &Error{code: c, err: underlying}
 }
 
+// NewWireError is similar to [NewError], but the resulting *Error returns true
+// when tested with [IsWireError].
+//
+// This is useful for clients trying to propagate partial failures from
+// streaming RPCs. Often, these RPCs include error information in their
+// response messages (for example, [gRPC server reflection] and
+// OpenTelemtetry's [OTLP]). Clients propagating these errors up the stack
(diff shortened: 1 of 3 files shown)


jhump (approved):
LGTM. Thanks!

PR #440, opened 2023-01-25
Match Content-Type charset case-insensitively

ichizero:
The connect handler returns http.StatusUnsupportedMediaType when a request has a `Content-Type: application/json; charset=UTF-8` header. However, according to [RFC 9110 Section 8.3.2]( https://httpwg.org/specs/rfc9110.html#rfc.section.8.3.2), the charset parameter value should be treated as case-insensitive.

In this PR, I have modified the charset parameter value for user requests and acceptable content types of handlers to all be handled in lowercase.


akshayjshah (approved):
This is great - thank you, @ichizero!

PR #444, opened 2023-01-26
Minimize allocations parsing Content-Type

mattrobenolt:
For context, this function is run on every request within ServeHTTP, and the mime parsing and now lowercasing, is a marginal, but non trivial amount of memory allocations.

The expensive bit is hitting the mime.FormatMediaType path, when we already have a canonical form.

This removes the most expensive part over arguable the most common cases where there are no additional parameters on the Content-Type.

```
$ go test -bench '^BenchmarkCanonicalizeContentType$' -run '^$' .
goos: darwin
goarch: arm64
pkg: github.com/bufbuild/connect-go
BenchmarkCanonicalizeContentType/simple-10         	92344741	        12.85 ns/op	       0 B/op	       0 allocs/op
BenchmarkCanonicalizeContentType/with_charset-10   	 1744219	       693.8 ns/op	     424 B/op	       6 allocs/op
BenchmarkCanonicalizeContentType/with_other_param-10         	 1969113	       614.4 ns/op	     424 B/op	       6 allocs/op
PASS
ok  	github.com/bufbuild/connect-go	5.800s
```

### DIFF
--- a/protocol_test.go
+++ b/protocol_test.go
@@ -27,6 +27,7 @@ func TestCanonicalizeContentType(t *testing.T) {
 		arg  string
 		want string
 	}{
+		{name: "uppercase should be normalized", arg: "APPLICATION/json", want: "application/json"},
 		{name: "charset param should be treated as lowercase", arg: "application/json; charset=UTF-8", want: "application/json; charset=utf-8"},
 		{name: "non charset param should not be changed", arg: "multipart/form-data; boundary=fooBar", want: "multipart/form-data; boundary=fooBar"},
 		{name: "no parameters should be normalized", arg: "APPLICATION/json;  ", want: "application/json"},
--- a/protocol_test.go
+++ b/protocol_test.go
@@ -29,6 +29,7 @@ func TestCanonicalizeContentType(t *testing.T) {
 	}{
 		{name: "charset param should be treated as lowercase", arg: "application/json; charset=UTF-8", want: "application/json; charset=utf-8"},
 		{name: "non charset param should not be changed", arg: "multipart/form-data; boundary=fooBar", want: "multipart/form-data; boundary=fooBar"},
+		{name: "no parameters should not have base normalized", arg: "APPLICATION/json;  ", want: "application/json"},
 	}
 	for _, tt := range tests {
 		tt := tt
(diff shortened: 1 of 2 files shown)


mattrobenolt:
As a follow up, maybe a micro-optimization, but it might be worth having a switch statement to check for the known types to avoid even going through the `mime.ParseMediaType` path.

I only think this is a worthwhile optimization to make because for a connect client and a connect server, which in our case is all the time, we know the types it sent within a set of known types. So it seems pretty valid IMO to have a fast path for known client <> server hot path.

mattrobenolt:
Gross, but kinda what I came up with:

```diff
diff --git a/protocol.go b/protocol.go
index 3fd4369..2ba638f 100644
--- a/protocol.go
+++ b/protocol.go
@@ -316,6 +316,10 @@ func flushResponseWriter(w http.ResponseWriter) {
 }
 
 func canonicalizeContentType(ct string) string {
+	switch ct {
+	case "application/connect+proto", "application/connect+json", "application/json", "application/proto", "application/grpc", "application/grpc-web", "application/grpc+json", "application/grpc+proto", "application/grpc-web+json", "application/json; charset=utf-8", "application/json; charset=UTF-8":
+		return ct
+	}
 	base, params, err := mime.ParseMediaType(ct)
 	if err != nil {
 		return ct
```

I'm not sure if there's a simpler way to build this list, but this makes this function effectively free when paired with a connect-go client.

In go1.19+, I believe this switch gets optimized to a jump table as well.

protocol_test.go
```diff
@@ -29,6 +29,7 @@ func TestCanonicalizeContentType(t *testing.T) {
 	}{
 		{name: "charset param should be treated as lowercase", arg: "application/json; charset=UTF-8", want: "application/json; charset=utf-8"},
 		{name: "non charset param should not be changed", arg: "multipart/form-data; boundary=fooBar", want: "multipart/form-data; boundary=fooBar"},
+		{name: "no parameters should not have base normalized", arg: "APPLICATION/json;  ", want: "application/json"},
```
akshayjshah, line 32: Either the name or the assertion seems wrong to me. Should the description be "no parameters _should_ have base normalized"?

mattrobenolt: wups, lemme fix

akshayjshah:
@mattrobenolt Spiritually, I'm on board with being more even aggressive here - as long as we're in here optimizing, let's create a fast path that avoids all use of `mime`. The hard-coded switch you're suggesting isn't _totally_ out of the question, but I'd like to avoid baking so much knowledge of the protocol(s) and encodings into this function. I think we can do nearly as well if we leverage a little knowledge of the mime-type grammar.

Let me push a commit here and see what you think.

mattrobenolt:
I like this. Pretty solid middleground.

akshayjshah:
@mattrobenolt LMK how this looks to you - with the commits I've pushed, the fast path is down to zero allocs and ~10ns/op. I think this is an okay middle ground - it gets a little into the weeds of what a valid content-type is, but it doesn't directly couple the `canonicalizeContentType` function to gRPC or Connect protocol details.

akshayjshah:
😆 OK you're faster than me

akshayjshah (approved):

mattrobenolt:
@akshayjshah I updated the benchmark in the description with the new hot path. Not quite down to 2ns, but I'll take 12ns. :)

PR #445, opened 2023-01-26
Optimize header access

mattrobenolt:
There are a lot of attempts to directly set header values already, this unifies all Get/Set/Add/Del accesses to assume a canonical form already.

In my personal testings, this actually stands out in our hot paths in flame graphs doing header Get/Sets. Within this codebase, we already have constants in place already in the correct formats.

I've dug so deeply into this in the past that I created https://github.com/mattrobenolt/go-httpheaders which specifically highlights benchmarks of direct access vs using the Get/Set helpers.

Since these are called multiple times in very hot paths, it's a worthwhile optimization since we can trust the inputs.

### DIFF
--- a/protocol.go
+++ b/protocol.go
@@ -37,6 +37,7 @@ const (
 const (
 	headerContentType = "Content-Type"
 	headerUserAgent   = "User-Agent"
+	headerTrailer     = "Trailer"
 
 	discardLimit = 1024 * 1024 * 4 // 4MiB
 )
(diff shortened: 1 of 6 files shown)


protocol.go
```diff
@@ -37,6 +37,7 @@ const (
 const (
 	headerContentType = "Content-Type"
 	headerUserAgent   = "User-Agent"
+	headerTrailer     = "Trailer"
```
mattrobenolt, line 40: Figured since this was referenced multiple times, might as well elevate it to a constant.

akshayjshah (approved):
This is great - thank you! I pushed a commit that changes the "XXX:" comments to match Buf's style a bit better, but the remainder of the change is excellent.

mattrobenolt:
Perfect, those comments were more for yall while reviewing anyways, since I wasn't entirely sure if my assumption was true or not.

PR #446, opened 2023-01-26
Minimize allocations writing User-Agent header

mattrobenolt:
The fmt.Sprintf stood out again in my profiling as a relatively hot path, and through the lifetime of an application, this value will not change. So we can generate this header value at init() time and avoid the extra memory allocation as well on each request.

### DIFF
--- a/protocol_connect.go
+++ b/protocol_connect.go
@@ -52,7 +52,7 @@ const (
 // defaultConnectUserAgent returns a User-Agent string similar to those used in gRPC.
 //
 //nolint:gochecknoglobals
-var defaultConnectUserAgent = []string{fmt.Sprintf("connect-go/%s (%s)", Version, runtime.Version())}
+var defaultConnectUserAgent = fmt.Sprintf("connect-go/%s (%s)", Version, runtime.Version())
 
 type protocolConnect struct{}
 
@@ -248,7 +248,7 @@ func (c *connectClient) WriteRequestHeader(streamType StreamType, header http.He
 	// We know these header keys are in canonical form, so we can bypass all the
 	// checks in Header.Set.
 	if header.Get(headerUserAgent) == "" {
-		header[headerUserAgent] = defaultConnectUserAgent
+		header[headerUserAgent] = []string{defaultConnectUserAgent}
 	}
 	header[connectHeaderProtocolVersion] = []string{connectProtocolVersion}
 	header[headerContentType] = []string{
--- a/protocol_connect.go
+++ b/protocol_connect.go
@@ -49,6 +49,11 @@ const (
 	connectStreamingContentTypePrefix = "application/connect+"
 )
 
+// defaultConnectUserAgent returns a User-Agent string similar to those used in gRPC.
+//
+//nolint:gochecknoglobals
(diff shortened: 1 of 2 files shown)


mattrobenolt:
I'm not sure if you'd want to exclude these linter failures since this pattern is already used elsewhere, and doing this any other way would also require a global variable. (with a sync.Once)

protocol_connect.go
```diff
@@ -49,6 +49,11 @@ const (
 	connectStreamingContentTypePrefix = "application/connect+"
 )
 
+// defaultConnectUserAgent returns a User-Agent string similar to those used in gRPC.
+//
+//nolint:gochecknoglobals
+var defaultConnectUserAgent = []string{fmt.Sprintf("grpc-go-connect/%s (%s)", Version, runtime.Version())}
```
pkwarren, line 55:
```suggestion
var defaultConnectUserAgent = []string{fmt.Sprintf("connect-go/%s (%s)", Version, runtime.Version())}
```

mattrobenolt: I think you beat me by a second. :)

protocol_connect.go
```diff
@@ -49,6 +49,11 @@ const (
 	connectStreamingContentTypePrefix = "application/connect+"
 )
 
+// defaultConnectUserAgent returns a User-Agent string similar to those used in gRPC.
+//
+//nolint:gochecknoglobals
+var defaultConnectUserAgent = []string{fmt.Sprintf("connect-go/%s (%s)", Version, runtime.Version())}
```
pkwarren, line 55: The only potential negative I could think about this is someone could potentially modify this slice in an interceptor (and impact all future calls). Very unlikely but that might be a reason to make this a const string and just alloc the slice each time.

mattrobenolt: It can't be a const string since we need to build it at runtime. But yeah, someone would have to awkwardly do like, `Header["User-Agent"][0] = "..."` and it'd hijack it. Any other mutation such as append wouldn't since it'd create a new slice reference.

I'm fine with keeping it as `[]string{defaultConnectUserAgent}` and force a new allocation since the expensive part was the string formatting.

bufdev: I much prefer leaving nothing to chance with globals, so making a new string slice each time seems sensible to me...the global string slice feels like a future security issue somehow

bufdev:
Also just a comment, we/I hate force pushes on PRs because it overwrites review context...I only mention it as I assume you're going to be a long-time contributor and figured worth getting on same page :-)

mattrobenolt:
> Also just a comment, we/I hate force pushes on PRs because it overwrites review context

👍 

Would you typically then squash merge? As long as it's squashed, that's fine with me.

pkwarren:
> Would you typically then squash merge? As long as it's squashed, that's fine with me.

Yep - that's the only option on most of our repos.

pkwarren (approved):
LGTM. Will wait for @akshayjshah/@bufdev to merge.

akshayjshah (approved):
LGTM.

I'd prefer to stop here - we're eating a slice allocation every time we write this header, but I don't want a global `[]string` here. (I say this because long ago I tried to optimize away all the slice allocations, but the safety tradeoffs weren't worth it for a Buf-supported library.)

PR #447, opened 2023-01-27
Optimize Peer lookups

mattrobenolt:
I don't know if there was an intention behind lazily initializing the Peer struct on-demand within the Peer() method, but from what I can tell, Peer() ends up being called 100% of the time in every path I've seen, and within NewClient, it's ultimately called twice.

Within NewClient, we also parse the params.URL then multiple times through url.ParseRequestURI then again through url.Parse, which is rather expensive and requires at least 1 heap allocation for the url.URL struct which we discard.

This behavior changes so the URL itself is parsed 1 time, then pass the url.URL struct to both functions that need it. As well as memoize the actual Peer on our concrete Client struct.

There is a slight change in behavior, but from my understanding, this seems safe:

validateRequestURL runs through url.ParseRequestURI, and newPeerFromURL was passed through url.Parse instead. From my understanding and my testing, url.ParseRequestURI is more strict than url.Parse, but since validateRequestURL was always called first, I don't particularly see how a less strict URL could have gotten through to the subsequent call to create the Peer.

### DIFF
--- a/connect.go
+++ b/connect.go
@@ -291,12 +291,11 @@ type Peer struct {
 	Protocol string
 }
 
-func newPeerFromURL(urlString, protocol string) Peer {
-	peer := Peer{Protocol: protocol}
-	if u, err := url.Parse(urlString); err == nil {
-		peer.Addr = u.Host
+func newPeerFromURL(url *url.URL, protocol string) Peer {
+	return Peer{
+		Addr:     url.Host,
+		Protocol: protocol,
 	}
-	return peer
 }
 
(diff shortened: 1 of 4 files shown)


akshayjshah:
Makes sense to me. I think I had some vague thought that `Peer` might contain some mutable data (a map or slice or something), so we'd want to make a fresh one on each call to `Peer()`. Clearly, that didn't end up happening, so this is a very nice improvement.

I pushed a commit that inlines the grpc-specific NewPeer function, but that's just a taste thing - code's great. :heart:

akshayjshah (approved):

PR #454, opened 2023-02-06
Update to Go 1.20

bufdev:
`make` was not working with 1.20, this fixes it and upgrades CI to use 1.19 and 1.20. Note that `go.mod` is not changed from 1.18 as we do not use any 1.19-specific features yet, and want to remain as compatible as possible. There was one new issue with `golangci-lint` that required JSON tags, so I just fixed that.


Makefile
```diff
@@ -54,19 +54,10 @@ lintfix: $(BIN)/golangci-lint $(BIN)/buf ## Automatically fix some lint errors
...
-	@# the modified, cached, and new (--others) files, and a second for the
-	@# deleted files.
-	comm -23 \
-		<(git ls-files --cached --modified --others --no-empty-directory --exclude-standard | sort -u | grep -v $(LICENSE_IGNORE) ) \
-		<(git ls-files --deleted | sort -u) | \
-		xargs $(BIN)/license-header \
-			--license-type apache \
-			--copyright-holder "Buf Technologies, Inc." \
-			--year-range "$(COPYRIGHT_YEARS)"
+	$(BIN)/license-header \
```
bufdev, line 57: This also updates to the new `license-header` flow`

Makefile
```diff
@@ -84,16 +75,16 @@ $(BIN)/protoc-gen-connect-go:
...
 $(BIN)/license-header: Makefile
 	@mkdir -p $(@D)
 	GOBIN=$(abspath $(@D)) $(GO) install \
-		  github.com/bufbuild/buf/private/pkg/licenseheader/cmd/license-header@v1.9.0
+		  github.com/bufbuild/buf/private/pkg/licenseheader/cmd/license-header@90fa81df0e9ef86ed505956be1ab1e8ccd49aa52
 
 $(BIN)/golangci-lint: Makefile
 	@mkdir -p $(@D)
-	GOBIN=$(abspath $(@D)) $(GO) install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.50.0
+	GOBIN=$(abspath $(@D)) $(GO) install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.51.0
```
bufdev, line 87: This is required to get Golang 1.20 to work

handler_ext_test.go
```diff
@@ -135,7 +135,8 @@ func TestHandler_ServeHTTP(t *testing.T) {
 		assert.Equal(t, resp.StatusCode, http.StatusNotFound)
 
 		type errorMessage struct {
-			Code, Message string
+			Code    string `json:"code,omitempty"`
```
bufdev, line 138: This failed on `golangci-lint` v1.51.0 otherwise

akshayjshah (approved):

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/bufbuild/connect-go
 
-go 1.18
+go 1.19
```
nicksnyder, line 3: @akshayjshah the PR description says ` Note that go.mod is not changed from 1.18 as we do not use any 1.19-specific features yet, and want to remain as compatible as possible.` but it looks like it got updated?

This inconvenienced a customer: https://bufbuild.slack.com/archives/C03H66632C9/p1676658888902709

bufdev: Whoops, this likely should not have made it through, that looks to be on me

akshayjshah: I didn't update the PR description, so that's my fault. This changeset drops a set of Go 1.18-specific patches and the accompanying build tag complexity.

The compatibility guarantee that this library (and all our other Go libraries) offer is the same as the Go compiler itself - current version + one older one. The customer was using a compiler that's >1y old and is no longer being patched or supported by the Go team.

Do we want to support compilers that the Go team doesn't?

nicksnyder: Unless there was more complexity removed in subsequent PRs, the complexity in this PR diff doesn't seem like that big of a deal to support (at least for a bit longer). Was it getting in our way?

If we are using Go support as a baseline, I would think we would want to support at least N+1 (e.g. last 3 version of Go) to give customers a buffer (so they can upgrade Go and then upgrade us) since we are downstream of Go.

bufdev: (supporting go 1.2 is a joke, but I know he only updates when he has to)

akshayjshah: Go isn't Java or Swift: as of the last available data, only 3% of Go users take >1y to update to new releases. >50% update within 2 months. (Source is the [2021 Go developer survey](https://go.dev/blog/survey2020-results#TOC_7). Answers to that question were consistent enough year to year that they've stopped reporting results.) The customer in question was updating to Go 1.19 the very same day they ran into this problem.

Making compiler support a nuanced, ad-hoc decision made in each project has notable downsides: whether a `connect-go` service builds with a given compiler version becomes an emergent property of the particular libraries in question, rather than a clear policy.

I think it's easy to make a small change that avoids the particular issue we ran into here without preventing us from doing routine housekeeping: drop support for old compilers ~1 month after they're end-of-lifed, rather than ~1w as we did here.

jhump: > I know @jhump maintains support going back to like go 1.2 in his libraries, as a good counterexample.

FWIW, I only provided that kind of support because it was easy to do so (not many new language features that were compelling to adopt). At first, I did it intentionally because FullStory (once upon a time) had code running on the Go runtime for Google AppEngine (which at the time was very out of date from mainstream Go releases).

Even after our migration off of AppEngine, I basically let my protoreflect repo continue to support Go versions for as long as the official protobuf and grpc repos did (so I only removed old versions when those other packages stopped supporting them).

@bufdev, you'll be glad to know that is no longer the case. As of v1.15.0, jhump/protoreflect now requires Go 1.18 or higher 😛

mattrobenolt: 👋 as a marginal outsider and heaver user, I fully support using latest and latest-1 versions.

I think, if anything, connect-go is a bit more bleeding edge software and under active development. As a consumer of this library, I find it quite a bit more essential to stay up to date, and we're not as likely to be dealing with legacy software that's been shipped in maintenance mode for years.

Maybe in the future there comes a time when there's demand for some more backwards compatibility, but for the time being, I'd vote to stay as latest as we can to adopt newest features with the least amount of backwards compatibility shims for older versions.

We (PlanetScale) will trail behind adopting the latest new major versions of go until the `x.x.1` releases usually to ensure some extra level of stability, but other than that, we try to stay on latest as much as possible.

PR #455, opened 2023-02-06
Use lowercase for all header keys

bufdev:
See https://github.com/bufbuild/connect-go/issues/453.

This goes ahead and uses lowercase for all of our header keys, see the comment in `header.go` on this PR.

Note that this breaks a bunch of our tests, which is worrisome in itself - in theory, this should cause no issues with the tests, so I'm not sure if this is because of the tests, or because of some deeper concerns, but someone with more intimate familiarity with the codebase may know off the top of their heads.

Another option would be to special-case when `protocolGRPC.web == true`, and feed this the whole way down, but if the position we have is that we're case-insensitive on headers, then all lowercase should not matter, and we can more easily do it without special-casing. However, an argument for special-casing can be made: (1) if we do all lowercase and then clients come to rely on this, that's bad, and we can say that we only do this for grpc-web for legacy reasons (2) this may break existing poorly-behaving callers, but per the gRPC and Connect specs, it should not so I'm not sure if this is our responsibility.

### DIFF
--- a/header.go
+++ b/header.go
@@ -17,6 +17,7 @@ package connect
 import (
 	"encoding/base64"
 	"net/http"
+	"strings"
 )
 
 // EncodeBinaryHeader base64-encodes the data. It always emits unpadded values.
@@ -51,6 +52,18 @@ func mergeHeaders(into, from http.Header) {
 	}
 }
 
+// Despite the IETF spec for HTTP saying that header keys should be case-insensitive,
+// key gRPC-Web clients require headers to be lowercase. We think the spec should be
+// fixed, however we want to remain compatible with all gRPC-Web clients. As such,
+// we universally lowercase our header keys via strings.ToLower. Since we bypass
+// the Get and Set methods on http.Header, we have to do lowercase for all of our
+// canonical operations below.
+//
+// Note that strings.ToLower stops early if the key is already lowercase, and should
+// not allocate any additional memory.
+//
+// See https://github.com/bufbuild/connect-go/issues/453 for more details.
+
 // getCanonicalHeader is a shortcut for Header.Get() which
 // bypasses the CanonicalMIMEHeaderKey operation when we
 // know the key is already in canonical form.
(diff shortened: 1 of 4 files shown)


mattrobenolt:
Won't this introduce a decent performance hit for everyone else? From my understanding, the canonical header form is for internals to net/{http,http2}, and is going to break APIs.

For example, all keys within an `http.Header` are in the expected canonical form, so any caller doing `header.Get(...)` pulls the key out of the map correctly.

I'm fairly confident this is going to break consumers of anyone trying to use normal net/http stuff, and will be a performance hit otherwise, since there will be conversions internally to the actual canonical form.

I understand the intent of the change, but I think this needs to happen when writing the protocol, which is not what happens here.

mattrobenolt:
Yeah, so in `x/net/http2` for example, when encoding into HPACK headers get lowercased: https://cs.opensource.google/go/x/net/+/refs/tags/v0.5.0:http2/write.go;l=331-370

mattrobenolt:
So it seems in `net/http`, it does seem to keep the headers in the same form they are in the `http.Header` map and doesn't coerce them when writing to the wire.

My concerns are strictly around the APIs for interacting with `http.Header` which will break as a result of this for users.

So for example: https://go.dev/play/p/ogCmJu8ciyD

This is wrong behavior to anyone outside of connect-go and anyone that might be writing a middleware or interceptor.

timostamm:
Agree with @mattrobenolt here. I'm fairly sure this is the only place we need to convert keys to lowercase ourselves: https://github.com/bufbuild/connect-go/blob/main/protocol_grpc.go#L582-L592

gRPC-web trailers are marshaled into the response body, so they side-step the regular normalization.

bufdev:
I think I kicked up a lot of dust here heh :-) Full disclosure, just put this up to get the conversation going - which it now has, and it appears as though this isn't the right solution.

PR #459, opened 2023-02-09
More explicit error on empty JSON bodies

### LINKED ISSUE #452: Unexpected errors on empty HTTP request bodies
When doing a HTTP request without a body I end up with this response: 

```json
{"code":"invalid_argument","message":"unmarshal into *emptypb.Empty: proto: syntax error (line 1:1): unexpected token "}   
```

This is fixed by adding an empty body `{}` to my HTTP request, so it's not a major issue. However, I'd expect empty bodies to result in the empty value for my requests. Is that a strange expectation?

Basically, I want this to happen on the server side before my request handler is invoked, in Go-pseudo:
```go
if r.ContentLength == 0 {
  r.Body = bytes.NewBuffer("{}")
}

// rest of the Connect magic

### PR DESCRIPTION
**Pull request by akshayjshah**
Return a more understandable error when the JSON codec attempts to
unmarshal an empty body.

Fixes #452.

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -15,6 +15,7 @@
 package connect
 
 import (
+	"errors"
 	"fmt"
 
 	"google.golang.org/protobuf/encoding/protojson"
@@ -92,6 +93,9 @@ func (c *protoJSONCodec) Unmarshal(binary []byte, message any) error {
 	protoMessage, ok := message.(proto.Message)
 	if !ok {
 		return errNotProto(message)
+	}
+	if len(binary) == 0 {
+		return errors.New("empty string is not a valid JSON object")
 	}
 	var options protojson.UnmarshalOptions
 	return options.Unmarshal(binary, protoMessage)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: codec.go [hunk 1]
```diff
@@ -93,6 +94,9 @@ func (c *protoJSONCodec) Unmarshal(binary []byte, message any) error {
 	if !ok {
 		return errNotProto(message)
 	}
+	if len(binary) == 0 {
+		return errors.New("empty string is not a valid JSON object")
```
mattrobenolt, line 98: Can we elevate this up to a package level `var ErrEmptyBody` or something to make it easier for a user to check like, `err == connect.ErrEmptyBody`

joshcarp (approved):

codec.go
joshcarp: I think this error is still mostly for manual inspection; 
it's useful in testing and manual verification if something goes wrong but I don't think clients should ever be sending invalid json and then building error handling around this error because they should check that empty strings aren't sent in the first place

akshayjshah: Tend to agree with Josh; I'd like to see a little more evidence that people need programmatic comparison to this error - e.g., they're writing interceptors that handle this case and retry the inner function with a zero-value request message.

Public API is forever, so I'd rather be conservative in adding it.

PR #460, opened 2023-02-14
Add stable codec interface and implementations

jchadwick-buf:
Unfortunately, protobuf offers no story whatsoever for canonicalization of protobuf or protojson output. It seems the best we can get is to just make it deterministic within each implementation.

Question: I did not add tests for this particular code because codecs are currently not tested on their own. Should we start having unit tests for the codecs now?

Related issues:
* https://github.com/golang/protobuf/issues/1121
* https://github.com/golang/protobuf/issues/1373

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -53,6 +53,8 @@ type Codec interface {
 
 // stableCodec is an extension to Codec for serializing with stable output.
 type stableCodec interface {
+	Codec
+
 	// MarshalStable marshals the given message with stable field ordering.
 	//
 	// MarshalStable should return the same output for a given input. Although
--- a/codec.go
+++ b/codec.go
@@ -15,6 +15,8 @@
 package connect
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
 
 	"google.golang.org/protobuf/encoding/protojson"
@@ -49,6 +51,30 @@ type Codec interface {
 	Unmarshal([]byte, any) error
 }
 
+// StableCodec is an extension to Codec for serializing with stable output.
+type StableCodec interface {
+	// MarshalStable marshals the given message with stable field ordering.
+	//
+	// MarshalStable should return the same output for a given input. Although
+	// it is not guaranteed to be canonicalized, the marshalling routine for
+	// MarshalStable will opt for the most normalized output available for a
+	// given serialization.
+	//
+	// For practical reasons, it is possible for MarshalStable to return two
+	// different results for two inputs considered to be "equal" in their own
+	// domain, and it may change in the future with codec updates, but for
+	// any given concrete value and any given version, it should return the
+	// same output.
+	MarshalStable(any) ([]byte, error)
+
+	// IsBinary returns true if the marshalled data is binary for this codec.
+	//
+	// If this function returns false, the data returned from Marshal and
+	// MarshalStable are considered valid text and may be used in contexts
+	// where text is expected.
+	IsBinary() bool
+}
+
 type protoBinaryCodec struct{}
 
 var _ Codec = (*protoBinaryCodec)(nil)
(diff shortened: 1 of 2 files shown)


codec.go
```diff
@@ -49,6 +51,30 @@ type Codec interface {
 	Unmarshal([]byte, any) error
 }
 
+// StableCodec is an extension to Codec for serializing with stable output.
+type StableCodec interface {
```
bufdev, line 55: Is adding a new interface here the right call? This feels like we're polluting the namespace a bit.

codec.go
```diff
@@ -49,6 +51,30 @@ type Codec interface {
...
+	// any given concrete value and any given version, it should return the
+	// same output.
+	MarshalStable(any) ([]byte, error)
+
+	// IsBinary returns true if the marshalled data is binary for this codec.
+	//
+	// If this function returns false, the data returned from Marshal and
+	// MarshalStable are considered valid text and may be used in contexts
+	// where text is expected.
+	IsBinary() bool
```
bufdev, line 75: Codecs shouldn't need to expose anything about their format.

codec.go
```diff
@@ -49,6 +51,30 @@ type Codec interface {
...
+	// it is not guaranteed to be canonicalized, the marshalling routine for
+	// MarshalStable will opt for the most normalized output available for a
+	// given serialization.
+	//
+	// For practical reasons, it is possible for MarshalStable to return two
+	// different results for two inputs considered to be "equal" in their own
+	// domain, and it may change in the future with codec updates, but for
+	// any given concrete value and any given version, it should return the
+	// same output.
+	MarshalStable(any) ([]byte, error)
```
bufdev, line 68: Where is this called?

jchadwick-buf: Not yet. Note that this PR merges into a feature branch, not `main`, with the intent of being able to have code reviewed before incrementally before it's ready. I would like to add tests for this, but I suspect the choice to not test the `codec.go` code directly (thusfar) was deliberate.

codec.go
jchadwick-buf: Not sure. Me and Akshay discussed potential approaches. "Interface smuggling" seemed to be a sensible way to add a new optional extension like this.

codec.go
akshayjshah: Agreed with John here. To make Envoy and other middleboxes work, the runtime needs to know whether the encoded data is text or binary.

codec.go
akshayjshah: As a starting position, I'd love to keep this interface unexported and document the optional methods on the exported `Codec` interface. Can we see how that works out, and whether it ends up feeling too implicit?

In practice, I expect that _very_ few users will end up both customizing codecs and caring about GETs.

akshayjshah (approved):
I think unit tests for the codecs are appropriate now. I leaned toward end-to-end tests during connect's development b/c they were more robust to refactoring and gave me a better feel for the user experience.

Can you add a few tests? Otherwise this looks great.

codec.go
bufdev: Can we unexport it?

jchadwick-buf: Done.

jchadwick-buf:
Added tests and unexported the interface. The tests are a bit anemic at this point in time, if anyone has any ideas for how they could be expanded with more edge cases I'd like to. I'm working on stuff a bit out of order at the moment.

codec_test.go
```diff
@@ -0,0 +1,62 @@
...
+			0b00010_010, 4, 0x74, 0x65, 0x78, 0x74,
+		},
+	)
+	assertStableMarshalEquals(
+		t,
+		&protoJSONCodec{},
+		&pingv1.PingRequest{Text: "text", Number: 1},
+		[]byte(`{"number":"1","text":"text"}`),
+	)
+}
```
akshayjshah, line 45: A few suggestions to improve test coverage:

* I always like `testing/quick` to do these sorts of round-trip testing. There are a handful of examples in this repository.
* Rather than directly asserting the wire format, can we create a protobuf message with a `map[string]string` and assert that repeatedly marshaling it always produces the same output?

akshayjshah (approved):
LGTM! Think the testing could be improved a bit, but as long as those changes are confined to `codec_test.go` this can land without further review.

codec_test.go
jchadwick-buf: It was a little tricky but I think I have made tests that work how you wanted.

jchadwick-buf:
Changes were confined to `codec_test.go` other than adding the `Codec` interface into `stableCodec`. Since `stableCodec` is not exported and only for our usage I think that should be OK. Going to merge into the `get-support` branch as per Akshay's direction.

PR #467, opened 2023-02-18
Parse the url only once per Client

mattrobenolt:
This stood out on flame graphs due ultimately to the http.NewRequestFromContext call, which ends up calling url.Parse and this happens on every single RPC call.

I did an optimization in #447 which similarly, and this is a natural extension of this.

We can just parse the url string once during NewClient, validate at that point, then tag along the parsed *url.URL everywhere else and never use it as a string again. This value never mutates through the lifetime of a Client and isn't publicly available on any structs.

Before:
image

After:
image

### DIFF
--- a/client.go
+++ b/client.go
@@ -17,8 +17,11 @@ package connect
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"net/http"
+	"net/url"
+	"strings"
 )
 
 // Client is a reusable, concurrency-safe client for a single procedure.
@@ -55,7 +58,7 @@ func NewClient[Req, Res any](httpClient HTTPClient, url string, options ...Clien
 			Protobuf:         config.protobuf(),
 			CompressMinBytes: config.CompressMinBytes,
 			HTTPClient:       httpClient,
-			URL:              url,
+			URL:              config.URL,
 			BufferPool:       config.BufferPool,
 			ReadMaxBytes:     config.ReadMaxBytes,
 			SendMaxBytes:     config.SendMaxBytes,
@@ -171,6 +174,7 @@ func (c *Client[Req, Res]) newConn(ctx context.Context, streamType StreamType) S
 }
 
 type clientConfig struct {
+	URL                    *url.URL
 	Protocol               protocol
 	Procedure              string
 	CompressMinBytes       int
@@ -184,9 +188,14 @@ type clientConfig struct {
 	SendMaxBytes           int
 }
 
-func newClientConfig(url string, options []ClientOption) (*clientConfig, *Error) {
-	protoPath := extractProtoPath(url)
+func newClientConfig(rawURL string, options []ClientOption) (*clientConfig, *Error) {
+	url, err := parseRequestURL(rawURL)
+	if err != nil {
+		return nil, err
+	}
+	protoPath := extractProtoPath(url.Path)
 	config := clientConfig{
+		URL:              url,
 		Protocol:         &protocolConnect{},
 		Procedure:        protoPath,
 		CompressionPools: make(map[string]*compressionPool),
@@ -228,4 +237,20 @@ func (c *clientConfig) newSpec(t StreamType) Spec {
 		Procedure:  c.Procedure,
 		IsClient:   true,
 	}
+}
+
+func parseRequestURL(rawURL string) (*url.URL, *Error) {
+	url, err := url.ParseRequestURI(rawURL)
+	if err == nil {
+		return url, nil
+	}
+	if !strings.Contains(rawURL, "://") {
+		// URL doesn't have a scheme, so the user is likely accustomed to
+		// grpc-go's APIs.
+		err = fmt.Errorf(
+			"URL %q missing scheme: use http:// or https:// (unlike grpc-go)",
+			rawURL,
+		)
+	}
+	return nil, NewError(CodeUnavailable, err)
 }
--- a/duplex_http_call.go
+++ b/duplex_http_call.go
@@ -62,6 +62,13 @@ func newDuplexHTTPCall(
 	// to mutate the req.URL, we don't feel the effects of it.
 	url = cloneURL(url)
 	pipeReader, pipeWriter := io.Pipe()
+
+	// This is mirroring what http.NewRequestContext did, but
+	// using an already parsed url.URL object, rather than a string
+	// and parsing it again. This is a bit funny with HTTP/1.1
+	// explicitly, but this is logic copied over from
+	// NewRequestContext and doesn't effect the actual version
+	// being transmitted.
 	request := (&http.Request{
 		Method:     http.MethodPost,
 		URL:        url,
--- a/duplex_http_call.go
+++ b/duplex_http_call.go
@@ -284,15 +284,15 @@ func (d *duplexHTTPCall) getError() error {
 }
 
 // See: https://cs.opensource.google/go/go/+/refs/tags/go1.20.1:src/net/http/clone.go;l=22-33
-func cloneURL(u *url.URL) *url.URL {
-	if u == nil {
+func cloneURL(oldURL *url.URL) *url.URL {
+	if oldURL == nil {
 		return nil
 	}
-	u2 := new(url.URL)
-	*u2 = *u
-	if u.User != nil {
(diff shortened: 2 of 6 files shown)


client.go
```diff
@@ -229,3 +238,19 @@ func (c *clientConfig) newSpec(t StreamType) Spec {
 		IsClient:   true,
 	}
 }
+
+func parseRequestURL(rawURL string) (*url.URL, *Error) {
```
mattrobenolt, line 242: I moved this out of protocol.go and into client.go since this isn't protocol specific and was used exactly the same for each protocol. So I decided to change this simple to a `parseRequestURL` rather than `validateRequestURL` which seemed a bit more fitting for how it is used now and accomplishes the same goal.

duplex_http_call.go
```diff
@@ -52,19 +53,22 @@ type duplexHTTPCall struct {
...
-	)
-	request.Header = header
-	client := &duplexHTTPCall{
+	request := (&http.Request{
+		Method:     http.MethodPost,
+		URL:        url,
+		Header:     header,
+		Proto:      "HTTP/1.1",
+		ProtoMajor: 1,
+		ProtoMinor: 1,
```
mattrobenolt, line 67: This is mirroring what's expected in stdlib. I hate it, but it is what it is. Happy to comment this up, but we can really skip any of the helpers and conveniences that `NewRequestWithContext` provides since our inputs are well known.

duplex_http_call.go
```diff
@@ -73,15 +77,6 @@ func newDuplexHTTPCall(
 		request:           request,
 		responseReady:     make(chan struct{}),
 	}
-	if err != nil {
```
mattrobenolt, line 76: This entire path isn't needed anymore since constructing the Request can't fail now.

protobuf_util.go
```diff
@@ -22,8 +22,8 @@ import (
 // corresponding to the Protobuf package, service, and method. It always starts
 // with a slash. Within connect, we use this as (1) Spec.Procedure and (2) the
 // path when mounting handlers on muxes.
-func extractProtoPath(url string) string {
-	segments := strings.Split(url, "/")
+func extractProtoPath(path string) string {
```
mattrobenolt, line 25: This was mildly interesting in that now passing explicitly the url.Path here resolves a minor bug in this logic.

Before, a full url was passed into here such as `http://localhost/foo`, which ends up being parsed down to `/localhost/foo` rather than just simply `/foo`.

I doubt in practice this was an issue, but this felt more correct.

protocol_connect.go
```diff
@@ -74,13 +74,9 @@ func (*protocolConnect) NewHandler(params *protocolHandlerParams) protocolHandle
 
 // NewClient implements protocol, so it must return an interface.
 func (*protocolConnect) NewClient(params *protocolClientParams) (protocolClient, error) {
-	url, err := validateRequestURL(params.URL)
```
mattrobenolt, line 77: We do this validate/parse step now when creating the actual Client before we even get here, so there's no reason to do any validation per-protocol.

client.go
```diff
@@ -171,6 +174,7 @@ func (c *Client[Req, Res]) newConn(ctx context.Context, streamType StreamType) S
 }
 
 type clientConfig struct {
+	URL                    *url.URL
```
mattrobenolt, line 177: I couldn't tell anywhere clientConfig or any way this got exposed even though the fields are suggested to be public, so this felt safe to do. It seems this is only used internally, and most of the other fields on this struct get reused to be passed along to the `protocolClientParams` too.

duplex_http_call.go
```diff
@@ -52,19 +53,26 @@ type duplexHTTPCall struct {
...
 	httpClient HTTPClient,
-	url string,
+	url *url.URL,
 	spec Spec,
 	header http.Header,
 ) *duplexHTTPCall {
+	// ensure we make a copy of the url before we pass along to the
+	// Request. This ensures if a transport out of our control wants
+	// to mutate the req.URL, we don't feel the effects of it.
+	url = cloneURL(url)
```
mattrobenolt, line 63: This is erroring on the side of caution since we can't really control what an http.RoundTripper may do, and I'm not sure if it's expected that a RoundTripper is not allowed to mutate this, or is expected to make it's own copy if it were?

Doing this at least mirrors the safety we had when we passed along a string and it was parsed every time.

duplex_http_call.go
```diff
@@ -276,3 +275,17 @@ func (d *duplexHTTPCall) getError() error {
 	defer d.errMu.Unlock()
 	return d.err
 }
+
+// See: https://cs.opensource.google/go/go/+/refs/tags/go1.20.1:src/net/http/clone.go;l=22-33
+func cloneURL(u *url.URL) *url.URL {
```
mattrobenolt, line 280: This was trivial enough to grab from stdlib and it's not exposed publicly. Was actually surprised there was no `(*url.URL).Clone()` or similar.

mattrobenolt:
Just to follow up on this, we've been running this patch in production now and the improvement is measurable in that the url.Parse time is entirely gone.

This isn't significant on it's own, but since this is per-RPC, it's definitely a relatively expensive operation when the URL doesn't change. It also skips the method validation `http.validateMethod` since we're just using the constants and know it's valid.

Creating the `io.Pipe()` now what dominates here, and that's pretty essential.

akshayjshah:
> Creating the io.Pipe() now what dominates here, and that's pretty essential.

Weeeeeelllllll.....it's not really. IIRC most of your RPCs are unary, so we don't actually need the pipe or the extra goroutine - we've got all the data in memory and can easily pass it as the request body. We only need all the code in `duplex_http_call.go` for _streaming_ requests; unary and server streaming RPCs could use a different, faster implementation of `StreamingClientConn`. We had code written for this early in Connect's life, and IIRC it was ~30% faster.

I ended up deciding to drop the alternate code path and exercise the request-streaming code on every RPC because it's by far the most complex code in `connect-go`, and I wanted to flush out any bugs quickly. I think we're past that point now, so we can at least consider a PR to speed up unary and server streaming calls.

mattrobenolt:
Oh, if we can drop this and the goroutine, that'd be a decently large win. I guess it makes sense this is only for a bidirectional stream. I'd suspect it can be dropped also for a server steam only with a unary request? I'd be down to look into this unless you have a good grasp or what would be involved.

Regardless, deduping the url.Parse is still good.

akshayjshah:
> Oh, if we can drop this and the goroutine, that'd be a decently large win. I guess it makes sense this is only for a bidirectional stream. I'd suspect it can be dropped also for a server steam only with a unary request? I'd be down to look into this unless you have a good grasp or what would be involved.

Yep, it should also work for server streaming. IIRC you'd need a different implementation of `StreamingClientConn` that errors when `Send` is called more than once; other than that it shouldn't be too too bad. We'd love it if you took a stab at a PR!

Agreed that this PR is good either way :)

duplex_http_call.go
akshayjshah: RoundTripper implementations aren't supposed to mutate the URL:

> // RoundTrip should not modify the request, except for
	// consuming and closing the Request's Body. RoundTrip may
	// read fields of the request in a separate goroutine. Callers
	// should not mutate or reuse the request until the Response's
	// Body has been closed.

In practice, I don't think many people read the docs very carefully so this extra safety is a good idea.

duplex_http_call.go
akshayjshah: Makes sense. This is fine with me, but I would like a few comments and a link the the implementation of `NewRequestWithContext`. (The hard-coded HTTP/1.1 constants are offputting to read!)

akshayjshah (approved):
Definitely much nicer! Appreciate the optimization, especially since it also simplifies some irritating error handling.

Please do fix the lint errors about variable name length - our house style frowns upon 1-2 letter variable names unless their scope is _very_ small. Once the lint errors are fixed and the `NewRequestWithContext` comments are added in, this is good to merge.

mattrobenolt:
@akshayjshah pushed both changes.

In my defense, I copy/pasted `cloneURL` from stdlib. I'm sure you preferred that over me doing a `//go:linkname`. :)

akshayjshah (approved):

PR #473, opened 2023-03-13
Connect GET support (client)

jchadwick-buf:


client_ext_test.go
```diff
@@ -123,6 +123,10 @@ func TestClientPeer(t *testing.T) {
 		t.Parallel()
 		run(t)
 	})
+	t.Run("connect+get", func(t *testing.T) {
+		t.Parallel()
+		run(t, connect.WithGetURLMaxBytes(4096))
+	})
```
akshayjshah, line 129: 😻 This is _exactly_ the kind of testing diff I was hoping for - nice and clean!

option.go
```diff
@@ -240,6 +230,26 @@ func WithSendMaxBytes(max int) Option {
 	return &sendMaxBytesOption{Max: max}
 }
 
+// WithIdempotency declares the idempotency of the procedure. This can determine
+// whether a procedure call can safely be retried, and may affect which request
+// modalities are allowed for a given procedure call.
+//
+// In most cases, you should not need to manually set this. It is normally set
+// by the code generator for your schema.
```
akshayjshah, line 238: I'd include a code sample showing how to set idempotency level for a protobuf RPC. Connect's technically encoding-agnostic, but virtually everyone uses protobuf.

option.go
```diff
@@ -240,6 +230,26 @@ func WithSendMaxBytes(max int) Option {
...
+func WithIdempotency(idempotencyLevel IdempotencyLevel) Option {
+	return &idempotencyOption{idempotencyLevel: idempotencyLevel}
+}
+
+// WithGetURLMaxBytes allows using HTTP Get requests for requests whose URLs
+// would fall under the provided maximum number of bytes. Please note that this
+// feature has some performance impact for requests made to RPCs which are
+// applicable, as a message may need to be marshalled twice.
+//
+// If this value is set to zero, no HTTP Get requests will be allowed.
```
akshayjshah, line 248: Should clarify that this only applies to idempotent requests with no side effects. It'd also be good if we clarified _why_ this is important (middle boxes limit URL size) and what a reasonable value might be (what are the default limits for NGINX, Apache, Envoy, etc?).

Maybe `WithCacheableRequests(maxURLSize int)` or `WithHTTPGet(maxURLSize int)` would be a better name?

protocol_connect.go
```diff
@@ -820,6 +830,45 @@ func (u *connectStreamingUnmarshaler) EndStreamError() *Error {
...
+		query.Set(connectUnaryMessageQueryParameter, string(data))
+	}
+	url.RawQuery = query.Encode()
+	if len(url.String()) > m.getURLMaxBytes {
+		return false
+	}
+	m.duplexCall.SetMethod(http.MethodGet)
+	*m.duplexCall.URL() = url
+	return true
+}
```
akshayjshah, line 870: Few thoughts here:

* We're missing support for the `cmp` parameter, which indicates the compression scheme in use. (I think we may have missed that on the server side, too.)
* Ideally, we wouldn't ever marshal the same message twice. When we call `MarshalStable`, we get back a bag of data that ought to be usable for a `POST` request too - it's a valid serialization of the request message. If it's too big for a `GET`, we should try compressing it; if it's still too big, we should hand off the serialized-and-compressed data to the regular `connectMarshaler`. Can we just add a method to `connectMarshaler` to make this work?

akshayjshah (commented):
Looks very good so far! Left a few suggestions on `connectGetMarshaler`, where I think we can do a little better performance-wise.

Also, let's expose the HTTP version in use, the HTTP method, and query parameters (if applicable) on the `Peer`. That should make it easy to require GETs with middleware, inspect any query params added by middle boxes, etc.

option.go
jchadwick-buf: Done. c98f933e4190e34e8ace8fa9955ee6c9728b0fc0

option.go
jchadwick-buf: I chose `WithHTTPGet`, although I'm not really certain what the best approach is. 596851957872611bdcf249a139c456302a1ac374

protocol_connect.go
jchadwick-buf: Unfortunately, accomplishing all of this is essentially impossible with the way marshalers are laid out today. We need to do things that are request-specific for Get, but `connectUnaryMarshaler` is used for both requests and responses. Moreover, the control flow of `connectUnaryMarshaler.Marshal` is subtley different from what we want for HTTP `GET`, so it's hard to extract out the commonality. I wound up mostly duplicating the `connectUnaryMarshaler.Marshal` function. I'm open to suggestions here, but I suspect getting this to be any cleaner will require a deeper change. d270803c63ee6c7961e2c752022e5f656e6eaf1d

akshayjshah (approved):
Looks great! Some small nits, but no need for more review after they're addressed. Let's get this merged, then take a look at the exported API for all of this and see how we feel. I think we _may_ have gotten ourselves into a position where we've got a bit too much magic (esp with the codec interface upcast).

akshayjshah:
Ooops, think we might have missed this part of the previous code review:

> Also, let's expose the HTTP version in use, the HTTP method, and query parameters (if applicable) on the Peer. That should make it easy to require GETs with middleware, inspect any query params added by middle boxes, etc.

Can you add that here and tag me when it's ready?

jchadwick-buf:
Discussed offline: Due to the order interceptors run in, currently it is not possible to propagate this information to interceptors. We added Query to Peer, but it is currently only useful for handlers. More work will be done later before merging to main.

connect.go
```diff
@@ -288,9 +288,13 @@ type Spec struct {
 // On both the client and the server, Protocol is the RPC protocol in use.
 // Currently, it's either [ProtocolConnect], [ProtocolGRPC], or
 // [ProtocolGRPCWeb], but additional protocols may be added in the future.
+//
+// Query contains the query parameters for the request. For the server, this
+// will reflect the actual query parameters sent. For the client, it is unset.
 type Peer struct {
 	Addr     string
 	Protocol string
+	Query    url.Values
```
akshayjshah, line 297: Let's also add a `// server-only` to this field.

akshayjshah (approved):

connect.go
jchadwick-buf: FYI: I followed this onto the next PR, since I did merge this back into get-support preemptively. Sorry!

PR #475, opened 2023-03-17
Connect Get: Add serverside compression support

jchadwick-buf:
This was missing from the original server PR.

Q: Should we do anything for the accept header?

### DIFF
--- a/client_ext_test.go
+++ b/client_ext_test.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


akshayjshah (approved):
LGTM.

I don't think we need to do anything for the accept header (at least, not in this stack of PRs). Like gRPC, the Connect protocol doesn't really support content-type negotiation - the response always uses the same encoding as the request. If we want to change that, let's do it in a separate stack of PRs and address both the `POST` and `GET` paths.

jchadwick-buf:
Oops, I forgot to merge this. No, I think for now we don't really need to worry about Accept.

PR #478, opened 2023-03-17
Connect HTTP Get support

jchadwick-buf:


idempotency_level.go
```diff
@@ -0,0 +1,68 @@
...
+	// given call is "idempotent", such that multiple instances of the same
+	// request to this procedure would have the same side-effects as a single
+	// request. This is equivalent to [RFC 9110 § 9.2.2] "idempotent" methods.
+	// This level is a subset of the previous level. This idempotency level is
+	// appropriate for any procedure that is safe to retry multiple times
+	// and be guaranteed that the response and side-effects will not be altered
+	// as a result of multiple attempts, for example, entity deletion requests.
+	//
+	// [RFC 9110 § 9.2.2]: https://www.rfc-editor.org/rfc/rfc9110.html#section-9.2.2
+	IdempotencyIdempotent IdempotencyLevel = 2
```
bufdev, line 55: Just a note here - it feels like unknown < idempotent < noSideEffects in terms of strength of information about the RPC's behavior, but the ordering is unknown -> noSideEffects -> idempotent. Likely not, but worth breaking with descriptor.proto on this one?

akshayjshah: I'm indifferent - a logical ordering would be nice, but matching `descriptor.proto` is also nice (and it keeps the generated code a bit shorter).

codec.go
```diff
@@ -97,6 +143,27 @@ func (c *protoJSONCodec) Unmarshal(binary []byte, message any) error {
...
 }
 
+func (c *protoJSONCodec) MarshalStable(message any) ([]byte, error) {
+	// protojson does not offer a "deterministic" field ordering, but fields
+	// are still ordered consistently by their index. However, protojson can
+	// output inconsistent whitespace for some reason, therefore it is
+	// suggested to use a formatter to ensure consistent formatting.
+	// https://github.com/golang/protobuf/issues/1373
+	compactedJSON := new(bytes.Buffer)
+	messageJSON, err := c.Marshal(message)
```
akshayjshah, line 153: Minor nit: can we do `c.Marshal` first, so that we can appropriately size the buffer and potentially skip some slice-growth costs?

akshayjshah:
Overall, I really like the way this has shaped up! Connect's exported APIs don't change much:

1. We introduce the `IdempotencyLevel` enum and `WithIdempotency` option. Even if we didn't want to support GETs, I think we'd want these changes to make automatic retry middleware work sanely.
2. We introduce the `WithHTTPGet()` client option. I think this is as simple as opting into GETs can be.
3. For users who run into URL size limits, we also add the `WithHTTPGetMaxURLSize` client option. I see the importance of this, but it's also a little esoteric; I wonder whether we ought to start by keeping it unexported and opening an issue for feedback.

jchadwick-buf:
> 3. For users who run into URL size limits, we also add the `WithHTTPGetMaxURLSize` client option. I see the importance of this, but it's also a little esoteric; I wonder whether we ought to start by keeping it unexported and opening an issue for feedback.

Maybe we should re-think not having a default set here in that case; but it does put us in a hard spot, because as a user you wouldn't get to choose the fallback behavior. If we make it an error, the end user will probably need two clients just to have a graceful fallback. But if we make it fallback to POST, the end user can't even detect that its a GET or POST until after its finished.

Ultimately, most people would probably not have too much of an issue with either behavior, but neither seem particularly good as the *only* possible behavior.

codec.go
jchadwick-buf: Honestly, it seems unnecessary to allocate even that: we should have no issues re-using the same buffer. Reading the source code of `json.Compact` I can see that it is safe, since it literally can only write less than the input, and as such could never run over itself.

c2015a9e1e70c7d68604301082b1935bf2bbbc1c

protocol_connect.go
```diff
@@ -144,6 +180,38 @@ func (h *connectHandler) NewConn(
...
+			contentType,
+		)
+	}
+
+	codec := h.Codecs.Get(codecName)
+	// The codec can be nil in the GET request case; that's okay: when failed
+	// is non-nil, codec is never used.
+	if failed == nil && codec == nil {
+		failed = errorf(CodeInvalidArgument, "invalid message encoding: %q", codecName)
+	}
```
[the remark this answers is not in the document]
jchadwick-buf: Acknowledged, but I think maybe this could use more thought: semantically speaking, 415 feels like something that refers to the `POST` body or at least the headers for the `POST` body. Is it semantically correct to return a 415 for this? If you would still prefer that, I'm curious what you think would be a good way to plumb this error through such that we can respond specially for it.

codec.go
akshayjshah: Nice, that's even better.

protocol_connect.go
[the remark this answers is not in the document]
akshayjshah: I agree that a 415 feels semantically odd - whether it's appropriate depends on how liberally you want to read, ["The format problem might be due to...inspecting the data directly."](https://www.rfc-editor.org/rfc/rfc9110.html#name-415-unsupported-media-type) At best, we're on the border of what's acceptable.

But if we don't return a 415 here, we're now into territory where error handling differs based on protocol: gRPC servers should _always_ return a 415 (which clients should interpret as `CodeInternal`) if they don't understand the serialization scheme, but Connect servers may return a 415 or a 400 (with an explicit `CodeInvalidArgument`) depending on HTTP method. This waters down the "change protocols with a config flag, no other code changes required" story.

I think we may be able to do return 415s by having each protocol implement a `CanHandleGetCodec(query http.Header) bool` method, and have `handler.go` return a 415 if it fails. We could call that method in the switch in `handler.go` that selects the `protocolHandler`.

akshayjshah:
> Maybe we should re-think not having a default set here in that case; but it does put us in a hard spot, because as a user you wouldn't get to choose the fallback behavior. If we make it an error, the end user will probably need two clients just to have a graceful fallback. But if we make it fallback to POST, the end user can't even detect that its a GET or POST until after its finished.
> Ultimately, most people would probably not have too much of an issue with either behavior, but neither seem particularly good as the only possible behavior.

Eh, I think the option shape we have is the best way to support the full breadth of options that I suspect we'll eventually need. However, people always surprise me with how not-picky they are; a small part of me wonders if `WithHTTPGetMaxURLSize` will be unused for years.

All that said, I'm fine keeping it exported. I feel more strongly about the consistent error handling for sure.

protocol_connect.go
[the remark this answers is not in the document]
jchadwick-buf: I agree especially to keep semantics the same across GET/POST it would probably be necessary to handle this case the same, otherwise it creates a headache for translation. I gave a shot at it. da71ecab4b5cf49a49ca0ef192f3bac82e80c93b

protocol.go
```diff
@@ -99,6 +99,9 @@ type protocolHandler interface {
 	// request's context, a nil cancellation function, and a nil error.
 	SetTimeout(*http.Request) (context.Context, context.CancelFunc, error)
 
+	// CanHandlePayload returns true if the protocol can handle an HTTP request.
```
akshayjshah, line 102: nit: In the docs, can you mention that this is called _after_ we've already validated that the HTTP method is supported?

akshayjshah (approved):

akshayjshah:
Oh, please hold off on merging this until we chat next :)

I think we should (at least) cut a separate release with everything already on `main`, so that this is the only change in its release. We should probably also have a specification published and have the Envoy filter ready to PR.

client_ext_test.go
```diff
@@ -132,15 +132,6 @@ func TestClientPeer(t *testing.T) {
...
 			connect.WithHTTPGet(),
-			connect.WithHTTPGetMaxURLSize(256, false),
-			connect.WithSendGzip(),
-		)
-	})
-	t.Run("connect+get fallback", func(t *testing.T) {
-		t.Parallel()
-		run(t,
-			connect.WithHTTPGet(),
-			connect.WithHTTPGetMaxURLSize(1, true),
```
jchadwick-buf: Unfortunately, exporting this out into an internal test is infeasible: as far as I can tell, there's no reasonable way to import the generated code from a given protobuf inside of the connect-go package without causing an import cycle, and moving the test to outside of the connect-go package would again prevent us from accessing the newly unexported option.

option.go
```diff
@@ -278,10 +278,14 @@ func WithHTTPGet() ClientOption {
...
 // By default, Connect-protocol clients with GET requests enabled may send a
 // URL of any size.
-func WithHTTPGetMaxURLSize(bytes int, fallback bool) ClientOption {
+func withHTTPGetMaxURLSize(bytes int, fallback bool) ClientOption {
 	return &getURLMaxBytes{Max: bytes, Fallback: fallback}
 }
 
+// TODO(jchadwick-buf): Remove this once we've determined whether this option
+// should be exported or removed.
+var _ = withHTTPGetMaxURLSize(0, false)
```
jchadwick-buf, line 287: This is ugly, but otherwise the linter (correctly) determines that the option is unused.

client_ext_test.go
akshayjshah: We ought to be able to avoid cycles by skipping the code generated by protoc-gen-connect-go and using NewHandler and NewClient directly. This is one of Connect's benefits vs grpc-go - generics make generated code a convenience, not a necessity.

jchadwick-buf: 911951962ab73a5731fdd802144e68d4dd3108a6

akshayjshah:
LGTM with the test added back.

protocol.go
jchadwick-buf: Oops, done in 5b79b761420b1460854696c6ce2ad9901545f9f7.

bufdev (approved):

idempotency_level.go
akshayjshah: Talked in person, agreed to keep it as-is.

Issue #471, opened 2023-03-03
Generate constants for full method names


pkwarren:
**Is your feature request related to a problem? Please describe.**
It would be convenient in some cases to have access to the full method name for each method. Previous attempts to add this data required separate plugins: https://github.com/bufbuild/plugins/issues/264

**Describe the solution you'd like**
Add the full method name as a constant in the generated .go code.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

bufdev:
I'd like to make sure this is necessary before we potentially pollute the generated code.

akshayjshah:
> I'd like to make sure this is necessary before we potentially pollute the generated code.

Ack. This comes up pretty often (#326, #383, Slack, in person with https://github.com/cuebox) - the most common use case is exempting some RPCs from logging or auth middleware without hard-coding a string literal. The [best available workaround](https://github.com/bufbuild/connect-go/issues/326#issuecomment-1354089305) is not very good, so IMO this is worthwhile.

bufdev:
Ok

Issue #470, opened 2023-03-03
Update protoc-gen-connect-go to preserve more comments


pkwarren:
**Is your feature request related to a problem? Please describe.**
When converting .proto files to .go files, some comments for the file and the package are lost. A recent update to grpc-go fixed a similar problem, and protobuf-go has addressed this for some time.

**Describe the solution you'd like**
Preserve more comments from the original .proto file in the generated .go source.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

akshayjshah:
@VoyTechnology we've already merged support for this, so your issue should be solved in the next release!

PR #480, opened 2023-03-18
Improve comments & add procedure consts to generated code

### LINKED ISSUE #326: HTTP middleware can't identify RPC paths
When using `net/http` middleware to wrap handlers and collect logs or metrics, it's nice to be able to parse out the protobuf package name, service name, and procedure name. However, it's potentially expensive to do this without an allowlist of known RPC endpoints: any script kiddie can generate zillions of invalid URLs, which may put a _lot_ of load on observability systems (especially systems that treat each unique combination of tags as a timeseries).

@mattrobenolt brought this to our attention. Previously, he used grpc-go's generated `ServiceDesc` to build an allowlist. (This isn't what the gRPC folks want users to do with `ServiceDesc`, but that's beside the point.)

Today, users have a few options to solve this problem:
* Emit logs & metrics from a connect interceptor. At that point, URLs have already been validated and `Spec.Procedure` is available. However, users may end up with one log line from HTTP middleware and another from a connect interceptor.
* Use the [generated service name constants](https://github.com/bufbuild/connect-go/blob/main/internal/gen/connect/ping/v1/pingv1connect/ping.connect.go#L39) to build a partial allowlist. This only validates the `package.service` portion of the URL, but that's likely enough to protect from most scanning attempts.
* Write a separate plugin to generate an exact allowlist. This works, but is a bit of a pain.

If this is a common pain point, we could make this a bit easier by generating more code.

akshayjshah:
This PR improves `protoc-gen-connect-go`'s generated code in two ways:

* First, it propagates comments on the protobuf syntax or package statements to
  the generated code. This matches protoc-gen-go's behavior, with one small
  improvement: comments attached to the protobuf package statement are also
  attached to the Go package declaration (vs. leaving an intervening blank
  line). This lets package-level protobuf comments also serve as package-level
  GoDoc.
- Second, it generates constants for each RPC's fully-qualified name. This lets
  users easily customize the behavior of interceptors or `net/http` middleware
  on a per-procedure basis. This has come up repeatedly, beginning shortly
  after release, so we ought to provide a satisfactory answer.
  `protoc-gen-go-grpc` began doing something similar in the most recent
  release.

Fixes #326, #470, and #471.


cmd/protoc-gen-connect-go/main.go
```diff
@@ -410,13 +459,8 @@ func serverSignatureParams(g *protogen.GeneratedFile, method *protogen.Method, n
...
 
-func procedureName(method *protogen.Method) string {
-	return fmt.Sprintf(
-		"/%s.%s/%s",
-		method.Parent.Desc.ParentFile().Package(),
-		method.Parent.Desc.Name(),
-		method.Desc.Name(),
-	)
+func procedureConstName(m *protogen.Method) string {
+	return fmt.Sprintf("%s%sProcedure", m.Parent.Desc.Name(), m.Desc.Name())
```
bufdev, line 463: On face value, naming this "Procedure" feels weird to me - Protobuf doesn't have any concept of procedure. I now realize connect does, but I don't know, must have missed that too. I just worry people are going to not grok that this is just a path

cmd/protoc-gen-connect-go/main.go
```diff
@@ -144,12 +168,37 @@ func generatePreamble(g *protogen.GeneratedFile, file *protogen.File) {
...
+	g.P(")")
+	g.P()
+
+	if numMethods == 0 {
+		return
+	}
+	wrapComments(g, "These constants are the fully-qualified names of the RPCs defined in this package. ",
+		"They're exposed at runtime as Spec.Procedure and as the final two segments of the HTTP route.")
+	g.P("//")
+	wrapComments(g, "Note that these are different from the fully-qualified method names used by ",
```
bufdev, line 189: I don't love this either, but I don't know what to do

cmd/protoc-gen-connect-go/main.go
```diff
@@ -68,6 +69,11 @@ const (
 	usage = "See https://connect.build/docs/go/getting-started to learn how to use this plugin.\n\nFlags:\n  -h, --help\tPrint this help and exit.\n      --version\tPrint the version and exit."
 
 	commentWidth = 97 // leave room for "// "
+
+	// To propagate top-level comments, we need the field number of the syntax
+	// declaration and the package name in the file descriptor.
+	protoSyntaxFieldNum  = 12
```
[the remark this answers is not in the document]
akshayjshah: Huh - are you sure? I did a fair bit of digging to choose the correct term here. From our own grammar:

```
FieldDecl = [ FieldCardinality ] TypeName FieldName equals FieldNumber
            [ CompactOptions ] semicolon .
```

and 

> Fields in a message are identified both by name and by number. The number, also called the "tag", is used in the binary format, for a more compact on-the-wire representation.

protobuf.dev calls them "field numbers". The code in protocolbuffers/protobuf nearly always uses "field number" or "tag field number" - from what I saw, "tag number" appears in only one test.

Are you sure that "tag number" is common and/or correct?

cmd/protoc-gen-connect-go/main.go
akshayjshah: Complain to the gRPC people? We're effectively stuck with this for compatibility. We can skip documenting the inconvenient fact that it's not the same as protobuf fully-qualified method names for some inscrutable reason, but that seems like a pretty significant footgun.

cmd/protoc-gen-connect-go/main.go
akshayjshah: If this is a sticking point, let's discuss it in person.

cmd/protoc-gen-connect-go/main.go
```diff
@@ -144,12 +168,37 @@ func generatePreamble(g *protogen.GeneratedFile, file *protogen.File) {
...
+	if numMethods == 0 {
+		return
+	}
+	wrapComments(g, "These constants are the fully-qualified names of the RPCs defined in this package. ",
+		"They're exposed at runtime as Spec.Procedure and as the final two segments of the HTTP route.")
+	g.P("//")
+	wrapComments(g, "Note that these are different from the fully-qualified method names used by ",
+		"google.golang.org/protobuf/reflect/protoreflect. To convert from these constants to ",
+		"reflection-formatted method names, remove the leading slash and convert the ",
+		"remaining slash to a period.")
```
pkwarren, line 192: Instead of documenting this in the generated code, is it worth adding a utility function to convert a procedure to a protoreflect.FullName? That seems like a useful method and we could simplify the generated code.

internal/gen/connect/ping/v1/pingv1connect/ping.connect.go
```diff
@@ -12,10 +12,15 @@
...
 
+// The canonical location for this file is
+// https://github.com/bufbuild/connect-go/blob/main/internal/proto/connect/ping/v1/ping.proto.
+
 // Code generated by protoc-gen-connect-go. DO NOT EDIT.
 //
 // Source: connect/ping/v1/ping.proto
 
+// The connect.ping.v1 package contains an echo service designed to test the
+// connect-go implementation.
```
pkwarren, line 23: It looks like the go code adds a trailing space here before the package statement.

internal/gen/connect/ping/v1/ping.pb.go
```diff
@@ -12,12 +12,18 @@
...
+// The canonical location for this file is
+// https://github.com/bufbuild/connect-go/blob/main/internal/proto/connect/ping/v1/ping.proto.
+
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.28.1
 // 	protoc        (unknown)
 // source: connect/ping/v1/ping.proto
 
+// The connect.ping.v1 package contains an echo service designed to test the
```
pkwarren, line 24: I'm guessing nothing has changed to the generated code between v1.28.1 and v1.30.0 (besides the version number), but it might be good to confirm the comment handling hasn't changed.

internal/gen/connect/ping/v1/pingv1connect/ping.connect.go
```diff
@@ -12,10 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// The canonical location for this file is
```
pkwarren, line 15: Is one benefit now for people not using our license header tools is that the license in the .proto file will be included?

pkwarren (commented):
Changes look great - I'm not great at naming so will let y'all figure out the best name for the constants.

internal/gen/connect/ping/v1/pingv1connect/ping.connect.go
pkwarren: It does look like that is the result - here's a run from the plugins repo from two images exported w/ `buf export`: https://gist.github.com/pkwarren/0628b6e2412d7f6483e59c23344d9079

cmd/protoc-gen-connect-go/main.go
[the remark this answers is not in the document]
timostamm: In the binary encoding, a tag is the result of combining field number and wire type via `(field_number << 3) | wire_type`. See [Go](https://pkg.go.dev/google.golang.org/protobuf/encoding/protowire#EncodeTag), [Java](https://github.com/protocolbuffers/protobuf/blob/35ffb665e2f3828853333713e1e4c9f59aedcff8/java/core/src/main/java/com/google/protobuf/CodedOutputStream.java#L228-L230), and the [encoding docs](https://protobuf.dev/programming-guides/encoding/#structure) for example. 

I'd expect "field number" here, not "tag" or "tag number". Some user-facing docs on protobuf.dev use "tag", but FieldNum seems appropriate to me here.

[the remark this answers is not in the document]
bufdev: Well, I guess I've been saying something incorrect for a decade.

bufdev (approved):

cmd/protoc-gen-connect-go/main.go
akshayjshah: Discussed in person, Peter's ok with this as-is.

cmd/protoc-gen-connect-go/main.go
akshayjshah: I'm hoping to minimize the number of places that the connect-go runtime couples closely to the protobuf runtime. I suspect that the _most common_ reason people would make this translation is to get a method descriptor, so I'm planning to just expose that directly in a subsequent PR.

internal/gen/connect/ping/v1/pingv1connect/ping.connect.go
akshayjshah: Yeah. I'm not sure why - the commits over there don't clarify. It looks like this is perhaps an artifact of reusing the same function for the comments on the protobuf package and the comments on the protobuf syntax. For us, it seems nice to produce Go code with package docs.

internal/gen/connect/ping/v1/ping.pb.go
akshayjshah: Good point, I'll update.

internal/gen/connect/ping/v1/pingv1connect/ping.connect.go
akshayjshah: Yep! This seemed like a good thing to me - do you think it's ok?

cmd/protoc-gen-connect-go/main.go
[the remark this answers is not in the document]
akshayjshah: Good catch. `protogen` already handles all this for us, so I think we can just switch to using the `GoName` attributes.

internal/gen/connect/ping/v1/ping.pb.go
akshayjshah: No changes in 1.30. I'm going to hold off on updating our `go.mod` - I'll open an issue discussing whether we ought to be aggressively bumping library versions here, since that effectively _forces_ users to update.

pkwarren (approved):

VoyTechnology:
Any estimates as to when this change will land in a release?

akshayjshah:
Likely Monday!

PR #482, opened 2023-03-21
Rewrite HTTP Get options

jchadwick-buf:
* No longer a requirement to specify URL size limit
* Separate URL size limit option adds fallback parameter


client_ext_test.go
```diff
@@ -128,7 +128,11 @@ func TestClientPeer(t *testing.T) {
 	})
 	t.Run("connect+get", func(t *testing.T) {
 		t.Parallel()
-		run(t, connect.WithHTTPGet(256), connect.WithSendGzip())
+		run(t,
+			connect.WithHTTPGet(connect.GetPolicyEnforce),
+			connect.WithHTTPGetMaxURLSize(256),
+			connect.WithSendGzip(),
+		)
```
akshayjshah, line 135: Can we also add a set of tests that exercise the get->post fallback?

option.go
```diff
@@ -247,25 +247,29 @@ func WithIdempotency(idempotencyLevel IdempotencyLevel) Option {
...
+}
+
+// WithHTTPGetMaxURLSize sets the maximum allowable URL length for Get requests;
+// this is useful as most user agents, middleboxes/proxies, and servers have
+// limits on the allowable length of a URL. For example, Apache and Nginx limit
+// the size of a request line to around 8 KiB, meaning that maximum length of a
+// URL is a bit smaller than this.
+//
+// If a URL would be longer than the configured maximum value here, the request
+// will be sent as an HTTP POST instead.
```
akshayjshah, line 268: This isn't true anymore - we should instead say that the fallback behavior is controlled by the GetPolicy.

protocol_connect.go
```diff
@@ -887,13 +889,14 @@ func (m *connectUnaryMarshaler) write(data []byte) *Error {
...
 	connectUnaryMarshaler
+	getPolicy      GetPolicy
 	getURLMaxBytes int
 	stableCodec    stableCodec
 	duplexCall     *duplexHTTPCall
 }
 
 func (m *connectUnaryRequestMarshaler) Marshal(message any) *Error {
-	if m.stableCodec != nil && m.getURLMaxBytes > 0 {
+	if m.stableCodec != nil && m.getPolicy != GetPolicyNever {
```
akshayjshah, line 899: Would it be better for us to return an error if the user has `GetPolicyEnforce` set and no stable codec? Otherwise we're falling back to POST, which seems bad. `CodeInternal` seems like the right severity level for misconfiguration like that.

protocol_connect.go
```diff
@@ -903,25 +906,28 @@ func (m *connectUnaryRequestMarshaler) marshalWithGet(message any) *Error {
 	// TODO(jchadwick-buf): This function is mostly a superset of
 	// connectUnaryMarshaler.Marshal. This should be reconciled at some point.
 	if message == nil {
-		if m.writeWithGet(nil, false) {
-			return nil
-		}
-		return m.write(nil)
+		return m.writeWithGet(nil)
```
akshayjshah, line 909: I believe that this panics. Can we get some test coverage around this code path?

akshayjshah (commented):
I have some line-level comments, but looking at this diff overall it seems that we may have overcomplicated the exported API. (That's probably my fault.)

We want a simple way for clients to opt into GETs. We also want a way to cap the size of GETs, with the default being "send any size request." Finally, we want clients to be able to _require_ GETs, with no POST fallback (primarily for tests).

I think we can get away with just `WithHTTPGet()` and `WithMaxGetURLSize(bytes int, fallback bool)`. This is super-simple to opt into, and defaults to unlimited URL size. Clients can configure a max size if necessary, and they can indicate whether they'd like to fall back to POST or not when they exceed the max. I think colocating the max and fallback will also clarify the docs and guide the user a bit more effectively.

akshayjshah (approved):
LGTM - thanks for bearing with me through some irritating changes of direction, @jchadwick-buf. I pushed some copy edits and Go nit fixes, along with two small-but-substantive changes:

- In https://github.com/bufbuild/connect-go/pull/482/commits/a552844cfe87e6963e4e41443f14e8da791fe350, I fixed `buildGetURL` so that it base64-encodes all compressed messages (since they're almost certainly binary).
- In https://github.com/bufbuild/connect-go/pull/482/commits/432894f323c0b7801532c764e67dbeb3d21b0010, I had unary Connect-protocol handlers automatically set the Vary response header (since the response content changes depending on the client's Accept-Encoding). This felt like a nice quality-of-life improvement - if users want to enable HTTP caching, they should now only need to worry about the request headers they're using in their application logic.

If the commits I've added look good to you, this PR is ready to merge.

protocol_connect.go
```diff
@@ -899,15 +899,15 @@ type connectUnaryRequestMarshaler struct {
...
 func (m *connectUnaryRequestMarshaler) Marshal(message any) *Error {
-	if m.enableGet {
-		if m.stableCodec == nil && !m.getUseFallback {
-			return errorf(CodeInternal, "codec %s doesn't support stable marshal; cam't use get", m.codec.Name())
-		}
-		if m.stableCodec != nil {
-			return m.marshalWithGet(message)
-		}
+	if !m.enableGet {
+		return m.connectUnaryMarshaler.Marshal(message)
```
jchadwick-buf: I think this is broken: it has return paths that don't return a value.

jchadwick-buf:
@akshayjshah I wound up adding a couple more changes:

- Added a case of the test to make sure the nil client send branch is covered.
- Fixed our use of base64; in the confusion of trying to clean up the base64 code, I accidentally forgot that the headers use standard base64, not URL-safe base64. This commit might be of interest for style check; maybe the resulting code in header.go should finally be split into its own file.

Otherwise, it's good to go from my end.

akshayjshah:
Nice final catches, John! The base64 dialect is a big one :)

> This commit might be of interest for style check; maybe the resulting code in header.go should finally be split into its own file.

Good point. They're only used in the connect protocol implementation, so I moved the query param stuff to the bottom of `protocol_connect.go`.

PR #483, opened 2023-03-21
Add timeout support for Get requests

jchadwick-buf:


akshayjshah (approved):

protocol_connect.go
```diff
@@ -53,6 +53,7 @@ const (
 	connectUnaryMessageQueryParameter     = "msg"
 	connectUnaryBase64QueryParameter      = "b64"
 	connectUnaryCompressionQueryParameter = "cmp"
+	connectUnaryTimeoutQueryParameter     = "t"
 	connectUnaryAPIQueryParameter         = "api"
```
akshayjshah, line 57: Reviewed this API briefly with Peter. He'd prefer longer query parameter names: "encoding", "message", "base64", "compression", and "timeout". I think he'd probably prefer "connect=v1" instead of "api=connectv1" too. We've got enough params here that I tend to agree - we've got kilobytes of space to work with, and making the overall URL more readable to people who don't know the protocol is valuable.

protocol_connect.go
```diff
@@ -970,6 +979,8 @@ func (m *connectUnaryRequestMarshaler) buildGetURL(data []byte, compressed bool)
 }
 
 func (m *connectUnaryRequestMarshaler) writeWithGet(url *url.URL) *Error {
+	delete(m.header, connectHeaderProtocolVersion)
```
akshayjshah, line 982: This  seems like a good fix to retain, even though we don't want to convert timeouts to query params.

akshayjshah (changes requested):

PR #493, opened 2023-04-06
Fix WithCompression to match docs; add tests

### LINKED ISSUE #492: Doc inaccuracy related to handler option WithCompression
The [doc comment](https://github.com/bufbuild/connect-go/blob/main/option.go#L121) for `WithCompression` states:
```
Calling WithCompression with an empty name or nil constructors is a no-op.
```

However, tracing through the code, this is not true. The constructors are never checked for nil. Instead, any existing compressor previously registered with the same name is blindly overwritten with an invalid one. It is invalid because if _either_ constructor is nil and ends up needing to be used, it will cause a nil dereference panic in the `compressionPool` [here](https://github.com/bufbuild/connect-go/blob/main/compression.go#L70) and [here](https://github.com/bufbuild/connect-go/blob/main/compression.go#L73).

Perhaps this should instead behave like `WithAcceptCompression`? In that option, the constructors _are_ checked. But if they are both nil, it effectively unregisters the named compression algorithm (instead of being a no-op).

jhump:
Fixes #492.


compression.go
```diff
@@ -65,6 +65,9 @@ func newCompressionPool(
 	newDecompressor func() Decompressor,
 	newCompressor func() Compressor,
 ) *compressionPool {
+	if newDecompressor == nil && newCompressor == nil {
```
akshayjshah, line 68: Perhaps OR instead of AND?

compression_test.go
```diff
@@ -54,3 +54,110 @@ func TestAcceptEncodingOrdering(t *testing.T) {
...
+		checkPools(t, config)
+	})
+	t.Run("WithCompression-nil-ctors-noop", func(t *testing.T) {
+		t.Parallel()
+		opts := []HandlerOption{WithCompression("foo", nil, nil)}
+		config := newHandlerConfig(testProc, opts)
+		assert.Equal(t, config.CompressionNames, []string{compressionGzip})
+		checkPools(t, config)
+	})
+	t.Run("WithCompression-nil-ctors-does-not-unregister", func(t *testing.T) {
```
jhump, line 156: I did not actually change the behavior from the way it was documented. I merely fix it so that it behaved as documented. Is that the right way to go? Or do you think it would be better to make it behave like `WithAcceptCompression`, where a compressor nil but nil constructors _unregisters_ the given name?

akshayjshah (approved):
One small suggestion, otherwise LGTM :shipit:

compression.go
jhump: AND matches the test that replaces it [here](https://github.com/bufbuild/connect-go/pull/493/files#diff-086a37ef11c8fd9a2516991339db79c207741ac98dd89c8215d31a8bead2814fL51).

But I definitely see that OR is much safer, in terms of avoiding potential runtime panics. However it is likely to produce inexplicible behavior later, when the compressor they tried to register just doesn't work. One being nil and the other not seems like a clear programmer error; would it be acceptable to instead fail hard and fast: panic immediately if one is nil and the other is not?

akshayjshah: Hm, that's fair. I'd personally prefer a panic right here, but that's hard-banned by the internal style guide. Let's keep it as is.

compression_test.go
akshayjshah: I think it'd be better if we behaved like `WithAcceptCompression`.

jhump: Done. See d5ecb751878dc193584208858a648cca4bff259f

PR #494, opened 2023-04-07
Add APIs to make and handle conditional GETs

akshayjshah:
If servers want clients to cache responses to HTTP GETs, they'll also
need a mechanism to return HTTP 304s. Unfortunately, we can't introduce
a new status code without breaking our promise that clients can switch
protocols with just a configuration flag, so we need a sentinel error
instead.

### DIFF
--- a/protocol_connect.go
+++ b/protocol_connect.go
@@ -518,7 +518,12 @@ func (cc *connectUnaryClientConn) validateResponse(response *http.Response) *Err
 			cc.compressionPools.CommaSeparatedNames(),
 		)
 	}
-	if response.StatusCode != http.StatusOK {
+	if response.StatusCode == http.StatusNotModified && cc.Spec().IdempotencyLevel == IdempotencyNoSideEffects {
+		serverErr := NewWireError(CodeUnknown, errNotModifiedClient)
+		// RFC 9110 doesn't allow trailers on 304s, so we only need to include headers.
+		serverErr.meta = cc.responseHeader.Clone()
+		return serverErr
+	} else if response.StatusCode != http.StatusOK {
 		unmarshaler := connectUnaryUnmarshaler{
 			reader:          response.Body,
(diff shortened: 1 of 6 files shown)


protocol_connect.go
```diff
@@ -518,7 +518,12 @@ func (cc *connectUnaryClientConn) validateResponse(response *http.Response) *Err
 			cc.compressionPools.CommaSeparatedNames(),
 		)
 	}
-	if response.StatusCode != http.StatusOK {
+	if response.StatusCode == http.StatusNotModified && cc.Spec().IdempotencyLevel == IdempotencyNoSideEffects {
```
jhump, line 521: Is it possible for the handler to detect whether the request was a GET vs a POST? Would that possibly be a better check here than just looking if the method is side-effect-free?

jhump (approved):
One question, otherwise LGTM.

protocol_connect.go
jhump: @jchadwick-buf, do you know the answer to above? I think either way, we can probably go ahead and merge this PR. If the answer is "no, there isn't yet a way for a handler to detect if current request was GET vs POST", then that is something that can be tacked on later.

jchadwick-buf (approved):

protocol_connect.go
akshayjshah: Unfortunately, no - currently, there's no way for _clients_ to easily detect whether the request was a GET or a POST. (The GET/POST decision happens after interceptors run, so there wasn't any need for such an API before.) We could add a way to detect GETs directly in a later PR - it's feasible, but would be a bit gross.

jhump: I was just talking about handlers, not clients.

jchadwick-buf: There's not a great way to tell with handlers, though as a hack it is possible to inspect the `Query` value on `Peer`.

PR #495, opened 2023-04-08
Implement fmt.Stringer interface for StreamType

svrana:
Following the example of `Code`, implement `fmt.Stringer` for `StreamType`. This makes it easier to use stream types in logs, traces, metric tags, and the like.

### DIFF
--- a/connect.go
+++ b/connect.go
@@ -26,6 +26,7 @@ package connect
 
 import (
 	"errors"
+	"fmt"
 	"io"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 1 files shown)


svrana:
I started porting/writing a logging interceptor at github.com/svrana/go-connect-middleware and it'd be nice to be able output the StreamType as a string as is done in the grpc version of this code. Course I could do that in my own project, but thought it might generally be useful.

bufdev:
I'm not sure we want to do this for this single enum - we'll discuss if we want to expose `fmt.Stringer` across enums. It adds to the API surface, so we want to make sure this is something we want to do.

connect.go
```diff
@@ -53,6 +54,20 @@ const (
...
+	case StreamTypeUnary:
+		return "unary"
+	case StreamTypeClient:
+		return "client"
+	case StreamTypeServer:
+		return "server"
+	case StreamTypeBidi:
+		return "bidi"
+	}
+	return fmt.Sprintf("stream_%d", s)
```
bufdev, line 68: If we do do this, generally we've just done `strconv.Itoa(int(s))` for these cases

akshayjshah:
@svrana I'm going to reopen this, tack on a few commits, and get this merged in :)

akshayjshah (approved):
This looks good to me. Thanks for following the example of `Code` so closely, @svrana!

PR #498, opened 2023-04-21
Prefix comments with correct identifier name

cuishuang:
To match typical Go style (even for unexported identifiers), the first word of most documentation comments should be the identifier name.


akshayjshah (approved):
Thanks for the fixes!

Issue #477, opened 2023-03-17
JSON unmarshaling by default errors on unknown field inputs.


koblas:
**Is your feature request related to a problem? Please describe.**
I'm porting a bunch of infrastructure from twirp to connect and one of the things I just got caught by was the JSON Codec for Service Unmarshaling by default signals an error for fields present in JSON not present in the protobuf description.

**Describe the solution you'd like**
Either:
1. By default use an options with discard set true (e.g.`options := protojson.UnmarshalOptions{DiscardUnknown: true}`) by default.
2. Have a way to configure options into the standard Codecs -- maybe as simple as `connect.WithJsonDiscardUnknown()`.

**Describe alternatives you've considered**
My current solution was to cut-n-paste out the JSON codec and provide it back in as an option to my service handler.

```
	_, api := apiv1connect.NewUserServiceHandler(
		user.NewUserServer(config, opts...),
		connect.WithCodec(bufcutil.NewJsonCodec()),
	)
```

**Additional context**

Nope.

akshayjshah:
This is pretty reasonable - we ought to make this sort of customization more straightforward.

We're definitely not going to make discarding unknown fields the default behavior. It doesn't seem clearly better than the current behavior, it's a significant change post-1.0, and it doesn't match the default for `protojson`.

Tentatively, I think it's best not to special-case this one toggle. If we're going to do anything, I'd prefer `WithJSONMarshalOptions` and `WithJSONUnmarshalOptions`. If you'd like to open a PR to add those two options, I'm happy to take a look. If not, we'll get back to this but it may take a little while - we're focused on adding support for HTTP GETs right now.

akshayjshah:
After thinking about this a bit more, I'd prefer to leave this configurability to third-party packages. From the discussion on #500:

> In general, I'm reluctant to expose APIs from `connect-go` that don't _require_ access to significant chunks of the package internals. Over time, this keeps the surface area of `connect-go` as small and approachable as possible.

> The `protojson` codec is so small that I don't think this API carries its weight. It's relatively easy, and just as functional, to provide this configurability from a separate package that replicates the JSON marshaling logic. I've done just that, so you can use  [`go.akshayshah.org/connectproto`](https://pkg.go.dev/go.akshayshah.org/connectproto) today. (Please do keep in mind that this is a small personal repo rather than an official Buf project!)

> We're sorting out the best way to make third-party packages like these more discoverable, but hopefully this unblocks you in the short term.

[...]

PR #500, opened 2023-04-23
[#477] add WithProtoJSONOptions() expose protojson marshalling options

natemurthy:
r? @akshayjshah

I'd like to merge this fork from @greg-montoux / @xxgreg  into the main repository. I'm experiencing the same challenges customizing my `protojson` marshalling/unmarshalling options as proposed in: Issue https://github.com/bufbuild/connect-go/issues/477

In my specific case, I have a large body of legacy JSON APIs maintained with `lower_snake_case` and the default `camelCase` marshaling is throwing some consistency curveballs my way.

I think this option method should be available to everyone as soon as convenience allows. My alternative is to maintain a fork this repository.

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -120,7 +120,9 @@ func (c *protoBinaryCodec) IsBinary() bool {
 }
 
 type protoJSONCodec struct {
-	name string
+	name             string
+	marshalOptions   protojson.MarshalOptions
+	unmarshalOptions protojson.UnmarshalOptions
 }
 
 var _ Codec = (*protoJSONCodec)(nil)
@@ -132,8 +134,7 @@ func (c *protoJSONCodec) Marshal(message any) ([]byte, error) {
 	if !ok {
 		return nil, errNotProto(message)
 	}
-	var options protojson.MarshalOptions
-	return options.Marshal(protoMessage)
+	return c.marshalOptions.Marshal(protoMessage)
 }
 
 func (c *protoJSONCodec) Unmarshal(binary []byte, message any) error {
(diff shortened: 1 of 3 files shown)


xxgreg:
(Replying from my other GH account)

Another API worth considering is to export a constructor for [protoJSONCodec](https://github.com/bufbuild/connect-go/blob/main/codec.go#L122) and allow setting the options there.

This would look something like:
```
mopts := protojson.MarshalOptions{
...
}
uopts := protojson.UnMarshalOptions{
...
}
cn.NewClient(..., cn.WithCodec(cn.NewProtoJSONCodec(mopts, upots))
```

It looks like the workaround in the meantime is to fork the `protoJSONCodec` struct, which isn't too bad either.

akshayjshah:
In general, I'm reluctant to expose APIs from `connect-go` that don't _require_ access to significant chunks of the package internals. Over time, this keeps the surface area of `connect-go` as small and approachable as possible.

The `protojson` codec is so small that I don't think this API carries its weight. It's relatively easy, and just as functional, to provide this configurability from a separate package that replicates the JSON marshaling logic. I've done just that, so you can use  [`go.akshayshah.org/connectproto`](https://pkg.go.dev/go.akshayshah.org/connectproto) today. (Please do keep in mind that this is a small personal repo rather than an official Buf project!)

We're sorting out the best way to make third-party packages like these more discoverable, but hopefully this unblocks you in the short term. I'll leave this PR open for a bit in case there's more discussion, but I don't intend to merge it.

akshayjshah (changes requested):
See above.

natemurthy:
I ended up going with @akshayjshah suggestion above, thanks everyone!

PR #502, opened 2023-05-04
Expose request method of unary requests to clients and server handlers

jhump:
In order for #494 (sentinel error to return "304 not modified" status) to be safe to use, the handler actually needs to know whether the incoming request used a POST vs GET. It's only safe to use that error for GET requests.

I haven't added tests yet because I wanted to maybe get a little feedback first no whether these additions are acceptable. The primary concern is about adding another unexported method to `AnyRequest`. Also, this is very targeted to only unary RPCs (nothing in the streaming flows ever attempts to expose this since they should always be POST anyway). But I didn't know if we should try to push more of this down further into the abstractions (though that's a bit tricky since the abstractions are exported interfaces `StreamingClientConn` and `StreamingHandlerConn`...).

Anyhow, if this idea seems work-able, then I'll finish this PR out with some tests.

In addition to making the request method available to server handlers, this also makes it available to clients -- when the unary RPC completes, the caller can inspect the request's method to see if it was sent via GET or POST.

WDYT?

### DIFF
--- a/client.go
+++ b/client.go
@@ -77,6 +77,11 @@ func NewClient[Req, Res any](httpClient HTTPClient, url string, options ...Clien
 	unarySpec := config.newSpec(StreamTypeUnary)
 	unaryFunc := UnaryFunc(func(ctx context.Context, request AnyRequest) (AnyResponse, error) {
 		conn := client.protocolClient.NewConn(ctx, unarySpec, request.Header())
+		if hasRequestMethod, ok := conn.(clientConnWithRequestMethod); ok {
+			hasRequestMethod.onSetMethod(func(method string) {
+				request.setRequestMethod(method)
+			})
+		}
 		// Send always returns an io.EOF unless the error is from the client-side.
 		// We want the user to continue to call Receive in those cases to get the
 		// full error from the server-side.
--- a/connect.go
+++ b/connect.go
@@ -135,6 +135,7 @@ type Request[T any] struct {
 	spec   Spec
 	peer   Peer
 	header http.Header
+	method string
 }
 
 // NewRequest wraps a generated request message.
@@ -172,8 +173,26 @@ func (r *Request[_]) Header() http.Header {
 	return r.header
 }
 
+// Method returns the HTTP method for this request. This is nearly always
+// POST, but side-effect-free RPCs could be made via a GET.
+//
+// On a newly created request, via NewRequest, this will return "POST" and
+// only changes to return to "GET" after it is actually sent to a server
+// using GET as the method.
+func (r *Request[_]) Method() string {
+	if r.method == "" {
+		return http.MethodPost
+	}
+	return r.method
+}
+
 // internalOnly implements AnyRequest.
 func (r *Request[_]) internalOnly() {}
+
+// setRequestMethod sets the request method to the given value.
+func (r *Request[_]) setRequestMethod(method string) {
+	r.method = method
+}
 
 // AnyRequest is the common method set of every [Request], regardless of type
 // parameter. It's used in unary interceptors.
@@ -190,8 +209,10 @@ type AnyRequest interface {
 	Spec() Spec
 	Peer() Peer
 	Header() http.Header
+	Method() string
 
 	internalOnly()
+	setRequestMethod(string)
 }
 
 // Response is a wrapper around a generated response message. It provides
@@ -313,6 +334,16 @@ type handlerConnCloser interface {
 	StreamingHandlerConn
 
 	Close(error) error
+}
+
+type handlerConnWithRequestMethod interface {
+	StreamingHandlerConn
+	getMethod() string
+}
+
+type clientConnWithRequestMethod interface {
+	StreamingClientConn
+	onSetMethod(fn func(string))
 }
 
 // receiveUnaryResponse unmarshals a message from a StreamingClientConn, then
--- a/protocol_connect.go
+++ b/protocol_connect.go
@@ -499,6 +499,10 @@ func (cc *connectUnaryClientConn) CloseResponse() error {
 	return cc.duplexCall.CloseRead()
 }
 
+func (cc *connectUnaryClientConn) onSetMethod(fn func(string)) {
+	cc.duplexCall.onSetMethod = fn
+}
+
 func (cc *connectUnaryClientConn) validateResponse(response *http.Response) *Error {
 	for k, v := range response.Header {
 		if !strings.HasPrefix(k, connectUnaryTrailerPrefix) {
@@ -719,6 +723,10 @@ func (hc *connectUnaryHandlerConn) Close(err error) error {
 	return hc.request.Body.Close()
 }
 
+func (hc *connectUnaryHandlerConn) getMethod() string {
+	return hc.request.Method
+}
+
 func (hc *connectUnaryHandlerConn) writeResponseHeader(err error) {
 	header := hc.responseWriter.Header()
 	if hc.request.Method == http.MethodGet {
@@ -928,7 +936,7 @@ type connectUnaryRequestMarshaler struct {
 func (m *connectUnaryRequestMarshaler) Marshal(message any) *Error {
 	if m.enableGet {
 		if m.stableCodec == nil && !m.getUseFallback {
-			return errorf(CodeInternal, "codec %s doesn't support stable marshal; cam't use get", m.codec.Name())
+			return errorf(CodeInternal, "codec %s doesn't support stable marshal; can't use get", m.codec.Name())
 		}
 		if m.stableCodec != nil {
 			return m.marshalWithGet(message)
--- a/duplex_http_call.go
+++ b/duplex_http_call.go
@@ -33,6 +33,7 @@ type duplexHTTPCall struct {
 	ctx              context.Context
 	httpClient       HTTPClient
 	streamType       StreamType
+	onSetMethod      func(method string)
 	validateResponse func(*http.Response) *Error
 
 	// We'll use a pipe as the request body. We hand the read side of the pipe to
@@ -150,6 +151,9 @@ func (d *duplexHTTPCall) URL() *url.URL {
 // SetMethod changes the method of the request before it is sent.
 func (d *duplexHTTPCall) SetMethod(method string) {
 	d.request.Method = method
+	if d.onSetMethod != nil {
+		d.onSetMethod(method)
+	}
 }
 
 // Read from the response body. Returns the first error passed to SetError.
--- a/handler.go
+++ b/handler.go
@@ -67,11 +67,16 @@ func NewUnaryHandler[Req, Res any](
 		if err := conn.Receive(&msg); err != nil {
 			return err
 		}
+		method := http.MethodPost
+		if hasRequestMethod, ok := conn.(handlerConnWithRequestMethod); ok {
+			method = hasRequestMethod.getMethod()
+		}
 		request := &Request[Req]{
 			Msg:    &msg,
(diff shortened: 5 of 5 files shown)


protocol_connect.go
```diff
@@ -928,7 +936,7 @@ type connectUnaryRequestMarshaler struct {
 func (m *connectUnaryRequestMarshaler) Marshal(message any) *Error {
 	if m.enableGet {
 		if m.stableCodec == nil && !m.getUseFallback {
-			return errorf(CodeInternal, "codec %s doesn't support stable marshal; cam't use get", m.codec.Name())
+			return errorf(CodeInternal, "codec %s doesn't support stable marshal; can't use get", m.codec.Name())
```
jchadwick-buf, line 939: Oops, nice catch.

client.go
```diff
@@ -77,6 +77,11 @@ func NewClient[Req, Res any](httpClient HTTPClient, url string, options ...Clien
 	unarySpec := config.newSpec(StreamTypeUnary)
 	unaryFunc := UnaryFunc(func(ctx context.Context, request AnyRequest) (AnyResponse, error) {
 		conn := client.protocolClient.NewConn(ctx, unarySpec, request.Header())
+		if hasRequestMethod, ok := conn.(clientConnWithRequestMethod); ok {
+			hasRequestMethod.onSetMethod(func(method string) {
+				request.setRequestMethod(method)
```
jchadwick-buf, line 82: nit: Couldn't this just be `hasRequestMethod.onSetMethod(request.setRequestMethod)`?

connect.go
```diff
@@ -172,9 +173,27 @@ func (r *Request[_]) Header() http.Header {
 	return r.header
 }
 
+// Method returns the HTTP method for this request. This is nearly always
+// POST, but side-effect-free RPCs could be made via a GET.
+//
+// On a newly created request, via NewRequest, this will return "POST" and
+// only changes to return to "GET" after it is actually sent to a server
+// using GET as the method.
+func (r *Request[_]) Method() string {
```
jchadwick-buf, line 182: The `Request` struct is also used for handlers, but I do not think `r.method` is being set by handlers in this PR. I think we either need to do that or document that this won't be set for handlers?

jchadwick-buf (approved):
I think the general approach will work. I'm not thrilled by it, but when I tried to do this for the client I ran into the same thing. It's not great since obviously the earliest point you can know a request was going to be a GET request is _after_ it's sent, and there's really no elegant way to implement this with how connect-go is currently architected. Therefore, I don't think there's much reasonable I can do other than give my general approval, albeit with the note that it is a bit begrudged :) At the very least, I think the API portion of this is good and what I think most people would expect, and that's the part we can't ever break, so that's the important bit.
I hope we can clean this up a bit later down the road by reorganizing the code to make it a bit less awkward, but there's no sense in trying to do it right now in my opinion. So yes. LGTM, with some caveats in the comments.

connect.go
jchadwick-buf: Somehow I missed the actual place where this gets set, nevermind.

client.go
jhump: Yes indeed, thanks for catching that!

jhump:
I didn't realize that `Query` was added to `Peer` (vs. adding it to `Request`, alongside `Header()` which seems like the closest attribute we expose.

So I guess the question -- potentially for @akshayjshah -- is whether this attribute should be added to `Peer` or to `Request`. I personally think it makes more sense in `Request` (and the fact that it's irrelevant for streaming RPCs, which are always `POST`, means we don't need any analog in the stream interfaces). But, perhaps it could be viewed as more consistent to put it next to the query param (though I see that is only ever populated for handlers, so putting it there might be more confusing...).

timostamm (commented):
Seems obvious to me that users need a way to distinguish GET requests in order to implement conditional requests. They shouldn't just look at `If-*` request headers. 

The tests suggest:

https://github.com/bufbuild/connect-go/blob/e04e4bd94c1d91031cdb2de84a50be3c003b275f/error_not_modified_example_test.go#L48

It does not seem very intuitive. `Method() string` looks more discoverable to me, but I am wondering whether a `Cacheable() bool` might be a better fit since we are in RPC land, not HTTP land.

jchadwick-buf:
I think I'm in favor of `Method`. Even though we're in RPC land, the abstraction is somewhat necessarily leaky here; we expose the concept of HTTP headers and trailers inside handlers, and obviously when setting up the RPC server there's no question that it is an HTTP handler. The reason we want to know the `Method` in handlers is primarily so that we can apply logic that is specific to HTTP GET anyways, and I think leaking the abstraction is the best option for that, since it gives the user the most flexibility and least confusion.
Certainly this PR is better than the `if len(req.Peer().Query) > 0` hack, because at least with the way Connect handlers are generated, you could easily formulate a request that gets the wrong answer.

As for `Peer` vs `Request` - My knee-jerk thought is that it probably belongs in `Request`. To be honest, I'm not sure that `Query` _really_ belongs in `Peer` either. I think I get what we were going for, but honestly, to me, the word "peer" implies that it's details about the peer itself, and to me the method/query are details about the request specifically. The way that we wound up with _other_ details is because _before_ a connection is made, a Client might have details in `Peer` that are taken from the URL rather than being representative of details about the peer itself, since there isn't a connection yet. But that's just my personal gut feeling.

From a user ergonomics standpoint, I definitely think calling `req.Method()` would be a more comfortable and obvious result than `req.Peer().Method` and even `req.Cacheable()`. Frankly, I think `req.Query()` would be more comfortable and obvious, too. Even if we didn't have a great justification for it, I think it would be nice to prioritize giving the user an obvious API that they can figure out without needing an example, when possible.

Another less-valid reason for why not to use `Peer` is one I experienced while trying to do it: it's very easy to wind up with data races when adding things to `Peer` that aren't known right away.

jhump:
I'm not opposed to a more semantic method, such as `Cacheable`. My only concern is if there are (now or in the future) other implications of GET. Then it might be awkward to ask if the request is cacheable when we are actually intending to ask about some other property that happens to be true when method == "GET". And it would also be awkward to expand the API surface area with more accessors for those other properties. ¯\\\_(ツ)\_/¯

I'd like to defer to @akshayjshah who can hopefully take a look at this early next week.

jchadwick-buf:
This might become more urgent to figure out if we're going to want to resolve the TODO in https://github.com/bufbuild/connect-es/pull/624 before merging, for similar functionality in connect-es. It'd be nice to have functional parity and unity on the API design front.

akshayjshah (commented):
I'm on board with exposing this on `Request` rather than `Peer`. I'd prefer to directly expose the HTTP verb rather than something semantic like `Cacheable` - let's stick as close to HTTP as we can.

In retrospect, I should probably have put the query params onto `Request` too. We can add a passthrough `Request.Query()` later on if there's demand - it doesn't seem likely to be used enough to double the API surface immediately.

@jhump When you add tests, can you please make sure to nail down how the new method behaves in interceptors? I'm particularly concerned that the current behavior will be a little misleading for client-side interceptors.

connect.go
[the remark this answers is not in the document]
jhump: > Do we expect clients to have much use for this in application code?

Like you mention, could be useful for metrics or logs. It's just to provide some observability of under-the-hood mechanics.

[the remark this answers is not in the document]
jhump: Changed to `HTTPMethod()` and also updated the behavior and docs (and verified in tests) that it returns empty string until the method is actually known (after RPC is sent).

connect.go
```diff
@@ -315,6 +336,16 @@ type handlerConnCloser interface {
...
 
+type handlerConnWithRequestMethod interface {
+	StreamingHandlerConn
+	getMethod() string
+}
+
+type clientConnWithRequestMethod interface {
+	StreamingClientConn
+	onSetMethod(fn func(string))
+}
```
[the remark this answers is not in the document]
jhump: I redid this. This is now using an anonymous interface as suggested for the handler side. For the client-side, to implement the new semantics (always blank until after RPC is sent, and set after), we need to know when the RPC is sent for all client conn impls. So now they all implement it, and there's no need for a type assertion.

jchadwick-buf:
Yeah, don't worry about that, GitHub Actions is having trouble again today. I noticed the same. https://www.githubstatus.com/

jhump:
@akshayjshah, ping, I was hoping for an LGTM from you before merging.

akshayjshah:
Will look from the plane today! I'm in airport purgatory RN.

interceptor_ext_test.go
```diff
@@ -137,6 +155,14 @@ func TestOnionOrderingEndToEnd(t *testing.T) {
...
 	assert.Equal(t, sum, 55)
 	assert.Nil(t, responses.Close())
+
+	// make sure the interceptors were invoked again
+	assert.Equal(t, int32(2), client1.Load())
+	assert.Equal(t, int32(2), client2.Load())
+	assert.Equal(t, int32(2), client3.Load())
+	assert.Equal(t, int32(2), handler1.Load())
+	assert.Equal(t, int32(2), handler2.Load())
+	assert.Equal(t, int32(2), handler3.Load())
```
akshayjshah, line 165: :+1: appreciate the improvement here!

interceptor_ext_test.go
```diff
@@ -268,3 +348,46 @@ func (cc *headerInspectingClientConn) Receive(msg any) error {
...
+	return func(ctx context.Context, req connect.AnyRequest) (connect.AnyResponse, error) {
+		h.count.Add(1)
+		if h.client {
+			// should be blank until after we make request
+			if req.HTTPMethod() != "" {
+				return nil, fmt.Errorf("expected blank HTTP method but instead got %q", req.HTTPMethod())
+			}
+		} else {
+			// server interceptors see method from the start
+			if req.HTTPMethod() != http.MethodPost {
```
akshayjshah, line 367: Should we loosen this up? For servers, we just need to know that the method isn't the empty string - depending on protocol, it's fine for it to be POST or GET.

interceptor_ext_test.go
```diff
@@ -268,3 +348,46 @@ func (cc *headerInspectingClientConn) Receive(msg any) error {
...
+				return nil, fmt.Errorf("expected HTTP method %s but instead got %q", http.MethodPost, req.HTTPMethod())
+			}
+		}
+		resp, err := unaryFunc(ctx, req)
+		if req.HTTPMethod() != http.MethodPost {
```
akshayjshah, line 372: Same here about looser assertions.

protocol.go
```diff
@@ -145,7 +145,14 @@ type protocolClient interface {
...
 	// methods return before the Receiver's Receive or Close methods are called.
-	NewConn(context.Context, Spec, http.Header) StreamingClientConn
+	NewConn(context.Context, Spec, http.Header) streamingClientConn
+}
+
+// streamingClientConn extends StreamingClientConn with a method for registering
+// a hook when the HTTP request is actually sent.
+type streamingClientConn interface {
+	StreamingClientConn
+	onRequestSend(fn func(*http.Request))
```
akshayjshah, line 155: style guide nit: blank line between embedded and named fields, please :)

akshayjshah (approved):
LGTM - some small nits and suggestions, but no need for more review from me.

protocol.go
jhump: Done

interceptor_ext_test.go
akshayjshah: OK, makes sense to me!

PR #504, opened 2023-05-10
Move asMaxBytesError to error.go

bufdev:
Extreme nit: since other error functions (including `as*` functions) are in `error.go`, it seems like it might be worth putting it in there to avoid an extra file.


akshayjshah:
Sure :) This used to be the Go 1.19+ variant, with a different version conditionally compiled for Go 1.18 or lower.

akshayjshah (approved):

PR #509, opened 2023-05-17
Expose HTTP method in unary handlers

akshayjshah:
Fixes #502 to expose the HTTP method in handlers.

### DIFF
--- a/protocol.go
+++ b/protocol.go
@@ -181,6 +181,13 @@ func (hc *errorTranslatingHandlerConnCloser) Close(err error) error {
 	return hc.fromWire(closeErr)
 }
 
+func (hc *errorTranslatingHandlerConnCloser) getHTTPMethod() string {
+	if methoder, ok := hc.handlerConnCloser.(interface{ getHTTPMethod() string }); ok {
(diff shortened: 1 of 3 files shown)


protocol.go
```diff
@@ -181,6 +181,13 @@ func (hc *errorTranslatingHandlerConnCloser) Close(err error) error {
 	return hc.fromWire(closeErr)
 }
 
+func (hc *errorTranslatingHandlerConnCloser) getHTTPMethod() string {
+	if methoder, ok := hc.handlerConnCloser.(interface{ getHTTPMethod() string }); ok {
+		return methoder.getHTTPMethod()
+	}
+	return http.MethodPost
+}
+
```
jhump, line 190: I think this is unused.

jhump: No, never mind, I see.

jhump (approved):
Well, shoot. I now see my error -- I had a test verifying `GET` only on the client side and failed to add something server side for this. Sorry about that.

Thanks for fixing!

jchadwick-buf (approved):
Great catch Akshay! Very sneaky issue.

PR #510, opened 2023-05-18
Add Request.Query(), deprecate Peer.Query

jchadwick-buf:
I know this idea came up briefly while we were adding `HTTPMethod`, WDYT? I don't want to suggest adding interfaces too hastily, but I think this is the more obvious place to find this functionality for the end user.


akshayjshah (commented):
I agree that this is better - and I do wish I'd done it this way to begin with - but I'm unsure how often users will want access to the raw query parameters. Clients can't set them, and I'm not sure how often servers will want to read them. I'm a little reluctant to merge this and add more API surface without being pretty sure that a fair number of users will run into trouble with this.

What do you think of leaving this PR open in the "changes requested" state and seeing if anyone pipes up on it?

jchadwick-buf:
Sure, makes sense to me.

akshayjshah (changes requested):
Blocking until/unless we have some stronger signal of user need. If you were confused by these APIs and want this PR merged, throw a 👍🏽 on the PR description!

akshayjshah:
This hasn't gotten much traction in a few months; I'm not sure it's worth the effort of rebasing the change on top of the rewritten commit history. @jchadwick-buf, do you want to keep it open or should we close this?

jchadwick-buf:
The changes weren't really that complicated, so it can always be redone later. Moving to close for now.

I still believe it is definitely the right way to go in the long-term, though. At the very least, we should add `Query()` to `Request` just to be congruent.

PR #515, opened 2023-05-30
Always close response body in CloseRead

pkwarren:
If we fail to discard bytes, we should still close the response body.


jchadwick-buf (approved):
This seems straight-forwardly correct to me, LGTM and good catch.

PR #519, opened 2023-06-08
Use switch based HTTP router rather than ServeMux

mattrobenolt:
So this one is a bit spicy, but `switch` statements in Go are pretty good and there have been recent optimizations to make them faster. I can only assume as the language progresses that switch statements will gain more optimizations at compile time.

The stdlib http.ServeMux does a bunch of extra things that we don't need here like handling redirecting to another matching endpoint with a trailing slash, etc. Due to the path cleaning and map access (with a lock), using the http.ServeMux is 10x slower than a handcrafted switch statement.

Given that you can't craft a switch statement at runtime, it makes sense that the default mux has to be backed by a map, but, good news is, we are already using codegen!

I've attached a benchmark simply to compare a http.ServeMux vs a switch based handler. I don't think it's too valuable to retain this, just here so we have something to compare numbers to.

Time for data and pprofs.

```
goos: darwin
goarch: arm64
pkg: github.com/bufbuild/connect-go
BenchmarkMux
BenchmarkMux/net/http
BenchmarkMux/net/http-10         	12484843	        95.20 ns/op	       0 B/op	       0 allocs/op
BenchmarkMux/switch
BenchmarkMux/switch-10           	100000000	        10.32 ns/op	       0 B/op	       0 allocs/op
PASS
```

So at least the good part is, we don't have any change in allocations, but we can see a massive reduction of ~90% in CPU time just to match a route.

Here's the pprofs:

Before:
image

We can see that most of the time is spent in... stuff we don't need for our use case.

After:
image

Can't get much cleaner than some `memeqbody`.

As for the rest, I'm not sure fully what is needed to be done to regenerate all the existing connect.go files within here, etc.


mattrobenolt:
I guess as an alternative would be to return a bit more structured `&fooServer` like we do for `&fooClient` that implements `ServeHTTP` rather than this `HandlerFunc`.

The nice part is the contract is we only return an `http.Handler`, so we have options. I don't have a strong opinion on which implementation. Just eyeballing, twirp does a `&fooServer` with a `ServeHTTP` and a `switch` for routes.

mattrobenolt:
> I think that this LGTM. There are some small changes to (undocumented) behavior here: while I don't think the default behavior would do things like automatically redirect to drop the slash, it does clean the path (that is, resolve .. and . elements.)

Ok, so I thought about this, and frankly, I think those behaviors are outside the spec of Connect, granted some of it can take some creative liberties on interpreting some things there.

The spec explicitly doesn't allow a trailing slash, for one, specifically the Path here: https://connect.build/docs/protocol#unary-request

Connect is also.. mostly POST (I know we just now support GETs) and POST can't be redirected with what the Mux spits back. The ServeMux spits back a 301 redirect, which explicitly prohibits a POST body from being re-sent. vs if it did a 308 redirect, then it'd at least be permitted.

Traversing `.` and `..` I feel is just... silly behavior, and also not declared in the spec of the protocol. It's not generically an HTTP server's responsibility to traverse these paths like this and it's, according to HTTP, also just valid to have a URL such as `example.com/../../..` and the path just simply be `/../../..` and serve content without an expectation of "directory" traversal.

I think overall, since we tend to, or at least attempt to, control both the server and clients, I'd prefer to optimize for the common case and not these edge cases.

> I checked and it looks like these switch statements tend to compile down to a binary search of string lengths. 

From go1.19 release notes:

> The compiler now uses a [jump table](https://en.wikipedia.org/wiki/Branch_table) to implement large integer and string switch statements. Performance improvements for the switch statement vary but can be on the order of 20% faster. (GOARCH=amd64 and GOARCH=arm64 only)

So I'm not entirely sure what all cases hit that, but seems pretty nice to hit a jump table. I think it's also fair to assume potential future compiler optimizations can make this better.

akshayjshah (approved):
Pushed a commit to remove the mux vs switch microbenchmark; otherwise LGTM!

PR #520, opened 2023-06-09
Lint for explicit import aliases

zchee:
Add `connect` and `pkngv1` import aliases for packages whose name doesn't match.

similar https://github.com/bufbuild/knit-go/pull/22


akshayjshah (approved):
Works for me - thanks for the contribution, @zchee!

PR #522, opened 2023-06-14
Allow clients to set Host in headers

### LINKED ISSUE #513: Provide the Host value to handlers
I wanted to use the `Host` field value of the `http.Request` in my application.
Since Go does not set `Host` header in `http.Header`, connect-go handlers cannot
access this information.

ref: https://cs.opensource.google/go/go/+/refs/tags/go1.20.4:src/net/http/request.go;l=157-158

It would be great if `connect.Request` provided the Host field value in the `http.Request`.
For example,

```go
func (MyHandler) DoSomething(ctx context.Context, req *connect.Request[pb.Request]) (*connect.Response[pb.Response], error) {
    fmt.Println(req.Host())
}
```

My current workaround is to wrap the HTTP handler returned by `NewXXX` generated function like:

```go
	mux := http.NewServeMux()
	p, h := myserviceconnect.NewMyServiceHandler(handler)
	h2 := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		r.Header.Set("Host", r.Host)
		h.ServeHTTP(w, r)
	})
	mux.Handle(p, h2)
```

This way, I can access the Host value from handlers with `req.Header().Get("Host")`.

emcfarlane:
Allows Host to set the request.Host on the client and promotes request.Host back to a Header in the handler.

Fixes #513

### DIFF
--- a/connect_ext_test.go
+++ b/connect_ext_test.go
@@ -521,6 +521,33 @@ func TestHeaderBasic(t *testing.T) {
 	assert.Equal(t, response.Header().Get(key), hval)
 }
 
+func TestHeaderHost(t *testing.T) {
+	t.Parallel()
+	const (
+		key  = "Host"
+		cval = "buf.build"
+	)
+
+	pingServer := &pluggablePingServer{
+		ping: func(_ context.Context, request *connect.Request[pingv1.PingRequest]) (*connect.Response[pingv1.PingResponse], error) {
+			assert.Equal(t, request.Header().Get(key), cval)
+			response := connect.NewResponse(&pingv1.PingResponse{})
+			return response, nil
+		},
+	}
+	mux := http.NewServeMux()
+	mux.Handle(pingv1connect.NewPingServiceHandler(pingServer))
+	server := httptest.NewServer(mux)
+	defer server.Close()
+
+	client := pingv1connect.NewPingServiceClient(server.Client(), server.URL)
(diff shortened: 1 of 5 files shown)


connect_ext_test.go
```diff
@@ -521,6 +521,33 @@ func TestHeaderBasic(t *testing.T) {
...
+			response := connect.NewResponse(&pingv1.PingResponse{})
+			return response, nil
+		},
+	}
+	mux := http.NewServeMux()
+	mux.Handle(pingv1connect.NewPingServiceHandler(pingServer))
+	server := httptest.NewServer(mux)
+	defer server.Close()
+
+	client := pingv1connect.NewPingServiceClient(server.Client(), server.URL)
```
akshayjshah, line 543: Can you try also testing this flow using the `connect.WithGRPC()` and `connect.WithGRPCWeb()` client options? I suspect that'll change how you approach fixing the bug 😉

akshayjshah (changes requested):
Congrats on your first PR as an official Buf employee, @emcfarlane! There's a little work still to do, but I left a suggestion on how to get started.

connect_ext_test.go
emcfarlane: Thanks for the quick review! I improved the test case and moved the logic. Still not sure if this is the right position for it.

akshayjshah (approved):
LGTM!

PR #525, opened 2023-06-20
Cache mapping of methods to protocol handlers

emcfarlane:
Remove allocation of an array for each request by mapping the handlers by methods. Follows pattern of caching allow method and content types.

```
                │  allocs/op  │  allocs/op   vs base           │
Connect/unary-8   247.0 ± ∞ ¹   243.0 ± ∞ ¹  ~ (p=1.000 n=1) ²
```


akshayjshah (approved):
Nice!

PR #526, opened 2023-06-20
Add WithCORS handler to support CORS

emcfarlane:
Implements a new handler option WithCORS to add support for CORS preflight and headers. AllowedMethods are taken from the protocol handlers and custom headers are exposed through the config option CORS.

Implemented with rs/cors.

Fixes https://github.com/bufbuild/connect-go/issues/499

### DIFF
--- a/cors.go
+++ b/cors.go
@@ -1,0 +1,77 @@
+// Copyright 2021-2023 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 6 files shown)


cors.go
```diff
@@ -0,0 +1,115 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package connect
+
+import (
+	"net/http"
+
+	"github.com/rs/cors"
```
mattrobenolt, line 20: Not sure how we feel about bringing in dependencies, but given the usage here is pretty straightforward, we can/should easily be able to implement the subset of features we need to expose imo without bringing in a dependency. CORS is relatively simple to implement.

emcfarlane:
Closing to add CORS support outside the connect-go package.

PR #527, opened 2023-06-21
New grpcadapter package to transcode connect and gRPC-Web

emcfarlane:
Creates a new package `grpcadapter` (`github.com/connect-go/grpcadapter`). Create a handler `grpcadapter.NewHandler` that can transcode connect and gRPC-web protocols to gRPC handlers. Allows gRPC go servers to use connect by wrapping [`grpc.Server`](https://pkg.go.dev/google.golang.org/grpc#Server.ServeHTTP). 

```go
	grpcServer := grpc.NewServer()
	mux := grpcadapter.NewHandler(grpcServer,
		grpcadapter.WithReadMaxBuffer(1024*1024),
		grpcadapter.WithWriteMaxBuffer(1024*1024),
	)
	http.ListenAndServe(":8080", h2c.NewHandler(mux, &http2.Server{}))
```

Unary connect requests are buffered and obey the limits set via connect options. For streams and gRPC-web these options are ignored and control is passed to the underlying gRPC handler.

Connect-crosstest is used to add more coverage: https://github.com/bufbuild/connect-crosstest/pull/478

Closes https://github.com/bufbuild/connect-go/issues/334


grpc_transcode.go
```diff
@@ -0,0 +1,602 @@
...
+
+func (w *connectResponseWriter) Flush() {
+	if w.isStreaming() {
+		flushResponseWriter(w.ResponseWriter)
+	}
+}
+
+// GRPCHandler translates connect and gRPC-web to a gRPC request for
+// use with gRPC handlers.
+func GRPCHandler(h http.Handler, options ...HandlerOption) http.Handler {
```
jhump, line 581: The use of `HandlerOption` here feels a bit misleading since most options don't apply.

Perhaps instead we need a new `MiddlewareOption` interface that extends (embeds) `HandlerOption` and use it just for the options it cares about (which seems to be max read amount and whether the connect version header is required). One trick is that max read is used for both handler and client -- so we might need yet another, like `OptionAny` that implements both `Option` and `MiddlewareOption`.

jhump (commented):
I need to check this branch out locally to really take a look. But here are a couple of initial comments.

protocol_grpc.go
```diff
@@ -45,10 +45,11 @@ const (
 	grpcTimeoutMaxHours = math.MaxInt64 / int64(time.Hour) // how many hours fit into a time.Duration?
 	grpcMaxTimeoutChars = 8                                // from gRPC protocol
 
-	grpcContentTypeDefault    = "application/grpc"
-	grpcWebContentTypeDefault = "application/grpc-web"
-	grpcContentTypePrefix     = grpcContentTypeDefault + "+"
-	grpcWebContentTypePrefix  = grpcWebContentTypeDefault + "+"
+	grpcContentTypeDefault        = "application/grpc"
+	grpcWebContentTypeDefault     = "application/grpc-web"
+	grpcWebTextContentTypeDefault = "application/grpc-web-text"
```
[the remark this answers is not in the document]
emcfarlane: Removed grpc-web-text support.

jhump:
I know I suggested `MiddlewareOption` and `OptionAny`, but looking more close, they don't quite feel right. For one, the actual function `GRPCHandler` doesn't have the word "middleware" in it, so it's unclear what `MiddlewareOption` applies to. Also, middleware is a bit vague since it could do anything -- and, in fact, interceptors are a form of middleware.

Let's tweak this a bit:
1. Rename `GRPCHandler` to `NewGRPCAdapter`, so its name is more inline with other functions used to instantiate RPC bits and also includes the word "adapter". I think this is more clear than "middleware".
2. Rename `MiddlewareOption` to `AdapterOption`.
3. I'm on the fence about the name `OptionAny`. I suggested that, having "Option" at the start, since it's like `Option`, in that it fills multiple roles and can be used in multiple contexts. But it looks kinds of funny and is inconsistent with the other names. Maybe `AnyOption` instead? (Not sure that's really much better, but it at least _looks_ more consistent...)

I'm still going through the code, but I also see that you are still pushing commits. Are those commits from doing cross tests with it? Should I wait for the outcome of that, in case you have to make any material changes to get the tests all green?

emcfarlane:
@jhump yep found some issues using crosstest: https://github.com/bufbuild/connect-crosstest/pull/478 These are parsing for me locally but the CI just failed so need to look into it. 
I like the `AdapterOption` and `NewGRPCAdapter` rename, will switch to that. 
I've been using the term transcode for testing, do we want to call it something else?

emcfarlane:
If we could break the Option style configuration we could have better documentation for what the specific config does. Either through accepting a struct:
```go
type GRPCAdapterConfig struct {
    UnaryReadMaxSize int
    UnaryWriteMaxSize int
}
func NewGRPCAdapter(h http.Handler, config GRPCAdapterConfig) http.Handler
```
Or through the adapter being configurable struct like http.Server:
```go
type GRPCAdapter struct {
    Handler http.Handler
    UnaryReadMaxSize int64
    UnaryWriteMaxSize int64
    // private
}
func (a *GRPCAdapter) ServeHTTP(w http.ResponseWriter, r *http.Request)
```

jhump:
> Or could define new types: GRPCAdapterOption with two options WithGRPCAdaterMaxReadSize and WithGRPCAdatperMaxWriteSize that don't have any overlap with current options.

I was actually thinking of this, too. And taking it a step further: moving the adapter into a sub-package, with its own options. That would be more clear I think. As to whether we stick with option-function pattern vs. options-struct pattern, I don't have as much of an opinion on.

grpc_adapter.go
```diff
@@ -0,0 +1,668 @@
...
+			config:     config,
+		}
+	}
+
+	// Handle GET request with query parameters
+	if request.Method == http.MethodGet {
+		generateGetBody()
+	} else {
+		generatePostBody()
+	}
```
jhump, line 373: Why define these as lambdas instead of just inlining here? Or maybe move GET to its own top-level named function (and pass it params instead of closing over params) so you can do something more like:
```go
if request.Method == http.MethodGet {
    translateConnectGetRequestToGRPC(...)
    return
}
// inline handling of POST here
```

grpc_adapter.go
```diff
@@ -0,0 +1,668 @@
...
+	if w.isStreaming() {
+		if !w.wroteHeader {
+			w.writeHeader(http.StatusOK)
+		}
+		flushResponseWriter(w.ResponseWriter)
+		return
+	}
+}
+
+// NewGRPCAdapter returns a new http.Handler that transcodes connect and
```
jhump, line 609: Style-wise, exported elements are supposed to appear before unexported elements. This has things in the opposite order. Even outside of our [style guide](https://www.notion.so/bufbuild/Go-Code-Standards-ff40382e6b104c1097a017f9565d54db#c6aeeed92adf40e189d7323ba0ccd11d), I find it personally easier to read to see the most important parts -- the entry to the story -- at the top, and then the supporting details below.

grpc_adapter.go
```diff
@@ -0,0 +1,668 @@
...
+	return nil
+}
+
+// bufferedEnvelopeReader buffers a message to wrap it in an envelope.
+type bufferedEnvelopeReader struct {
+	io.ReadCloser
+
+	config     *grpcAdapterConfig
+	buf        bytes.Buffer
+	buffered   bool
```
jhump, line 222: On first read, this was a bit confusing. I understand it now, but I think `initialized` might be a better word since the buffering is an initialization step.

grpc_adapter.go
```diff
@@ -0,0 +1,668 @@
...
+		r.buffered = true
+	}
+	if r.err != nil {
+		return 0, r.err
+	}
+	return r.buf.Read(data)
+}
+
+// readAll reads from r until an error or the limit is reached.
+func readAll(data []byte, reader io.Reader, limit int64) ([]byte, error) {
```
jhump, line 269: I think this could be simplified and leverage the std lib a bit more
1. If there is no limit, simply return `io.ReadAll`
2. If there is a limit, create an `io.LimitReader` with the limit == `limit`. Then call `io.ReadAll`. If the resulting data length == limit, call `reader.Read` to get one byte, and if it's not EOF, then the incoming data is too large.

grpc_adapter.go
```diff
@@ -0,0 +1,668 @@
...
+	if w.header == nil {
+		w.header = make(http.Header)
+	}
+	return w.header
+}
+
+// isStreaming checks via the content type if the response is streaming.
+func (w *connectResponseWriter) isStreaming() bool {
+	return w.typ == connectStreamingContentTypeDefault
+}
```
jhump, line 420: I think the implementation would be a little more straight-forward to split this into two response writers -- one for the unary protocol and one for the streaming protocol.

grpc_adapter.go
```diff
@@ -0,0 +1,668 @@
...
+			delHeaderCanonical(responseHeader, headerContentLength)
+			delHeaderCanonical(responseHeader, connectUnaryHeaderCompression)
+			setHeaderCanonical(responseHeader, headerContentType, connectUnaryContentTypeJSON)
+			_ = w.config.ErrorWriter.writeConnectUnary(w.ResponseWriter, err)
+		}
+	}
+	flushResponseWriter(w.ResponseWriter)
+}
+
+func (w *connectResponseWriter) Flush() {
```
jhump, line 599: To support Go 1.20's `ResponseController`, this type (and all impls of `ResponseWriter`) should also expose an `Unwrap` method:
```go
func (w *connectResponseWriter) Unwrap() http.ResponseWriter {
    return w.ResponseWriter
}
```
Ideally, we'd actually _check_ whether the underlying writer is a `http.Flusher` first, and then return a writer impl _without_ that method if it doesn't. As is, we just silently ignore this case in `flushResponseWriter`, but ideally the caller should be the one to decide what to do if the writer does not support flush.

In this case, [gRPC **requires** a writer that supports flush](https://github.com/grpc/grpc-go/blob/master/internal/transport/handler_server.go#L72-L76), so we should do the same: we should check for this at the top (in `NewGRPCAdapter`), instead of silently ignoring it when called.

grpc_adapter.go
```diff
@@ -0,0 +1,668 @@
...
+	"io"
+	"math"
+	"net/http"
+	"net/textproto"
+	"strconv"
+	"strings"
+)
+
+type grpcAdapterConfig struct {
+	Codec         Codec
```
jhump, line 32: This is not needed. GRPC trailers _always_ use the proto format for encoding the status. So just use `proto.Marshal` and `proto.Unmarshal` instead of this, which (to my eyes) implies that the trailer statuc codec could be customized (which is not true with GRPC).

jhump (commented):
Looks like the only real need for shared code is the set of header constants and the `setHeaderCanonical` and `getHeaderCanonical` helpers. If we want to move this to a sub-package, those would be trivial to refactor into an internal package.

grpc_adapter.go
emcfarlane: Was a linting issues, should have split out sooner.

grpc_adapter.go
emcfarlane: I've renamed this to:
```go
	buf            bytes.Buffer
	isInitialized  bool
	initializedErr error
	isCompressed   bool
```

grpc_adapter.go
emcfarlane: Added `ResponseController` interface!

For ConnectUnary Flushing isn't required and we want to avoid flushing the writer to capture trailers and re-encode as headers. I've added flusher checking to the other methods.

grpc_adapter.go
emcfarlane: It's part of the current internal func `grpcErrorFromTrailer`.

grpc_adapter.go
emcfarlane: Thanks, agree.

grpc_adapter.go
akshayjshah: If we're factoring this out into a separate package or repo, we don't need this flexibility. It's there in the main code path in case the user prefers a different implementation of protobuf serialization (vtproto, csproto, etc.).

In retrospect, using user-supplied codecs here may have been a bad idea - it complicates custom codecs a fair bit, since they need to handle both application-specific request/responses (where users have wide latitude to do unusual things) and protocol internals (where users have very little freedom to change behavior).

emcfarlane: Separated it into package `grpcadapter` with go import: github.com/connect-go/grpc-adapter.

grpcadapter/transport.go
```diff
@@ -0,0 +1,143 @@
...
+		i += 2
+	}
+	return out.String()
+}
+
+// The gRPC wire protocol specifies that errors should be serialized using the
+// binary Protobuf format, even if the messages in the request/response stream
+// use a different codec. Consequently, this function needs a Protobuf codec to
+// unmarshal error information in the headers.
+func grpcErrorFromTrailer(trailer http.Header) *connect.Error {
```
emcfarlane, line 102: Copied from `protocol_grpc.go`

grpcadapter/transport.go
```diff
@@ -0,0 +1,143 @@
...
+	stringReader := strings.NewReader(data)
+	if len(data)%4 != 0 {
+		// Data definitely isn't padded.
+		return base64.NewDecoder(base64.RawURLEncoding, stringReader)
+	}
+	// Data is padded, or no padding was necessary.
+	return base64.NewDecoder(base64.URLEncoding, stringReader)
+}
+
+func queryValueReader(data string, base64Encoded bool) io.Reader {
```
emcfarlane, line 59: Copied from `protocol_connect.go`.

grpcadapter/adapter.go
```diff
@@ -0,0 +1,770 @@
...
+		if contentEncoding := getHeaderCanonical(header, connectUnaryHeaderCompression); len(contentEncoding) > 0 {
+			isCompressed = contentEncoding != "identity"
+			delHeaderCanonical(header, connectUnaryHeaderCompression)
+			setHeaderCanonical(header, grpcHeaderCompression, contentEncoding)
+		}
+		request.Body = &bufferedEnvelopeReader{
+			ReadCloser:   request.Body,
+			isCompressed: isCompressed,
+			maxBytes:     readMaxBytes,
+		}
```
akshayjshah, line 471: If we have a `Content-Length` header, can we skip the buffering? It seems reasonable to trust the header.

grpcadapter/adapter.go
```diff
@@ -0,0 +1,770 @@
...
+			setHeaderCanonical(header, grpcHeaderCompression, contentEncoding)
+		}
+
+		msg := query.Get(connectUnaryMessageQueryParameter)
+		msgReader := queryValueReader(msg, query.Get(connectUnaryBase64QueryParameter) == "1")
+		request.Body = &bufferedEnvelopeReader{
+			ReadCloser:   io.NopCloser(msgReader),
+			isCompressed: isCompressed,
+			maxBytes:     readMaxBytes,
+		}
```
akshayjshah, line 459: Why do we need another buffer here? We ought to be able to calculate the payload size (or just eagerly un-base64 it and check the size).

grpcadapter/adapter.go
```diff
@@ -0,0 +1,770 @@
...
+			continue
+		}
+		if rawKey != key {
+			delete(header, rawKey)
+		}
+		header[key] = values
+	}
+	if statusCode != http.StatusOK {
+		w.ResponseWriter.WriteHeader(statusCode)
+	}
```
akshayjshah, line 534: Why not call the underlying ResponseWriter's WriteHeader method for 200s?

protocol.go
```diff
@@ -35,10 +35,11 @@ const (
...
 const (
-	headerContentType = "Content-Type"
-	headerHost        = "Host"
-	headerUserAgent   = "User-Agent"
-	headerTrailer     = "Trailer"
+	headerContentType   = "Content-Type"
+	headerContentLength = "Content-Length"
+	headerHost          = "Host"
+	headerUserAgent     = "User-Agent"
+	headerTrailer       = "Trailer"
```
akshayjshah, line 42: Do we need these changes?

protocol_connect.go
```diff
@@ -46,9 +46,10 @@ const (
 
 	connectFlagEnvelopeEndStream = 0b00000010
 
-	connectUnaryContentTypePrefix     = "application/"
-	connectUnaryContentTypeJSON       = connectUnaryContentTypePrefix + "json"
-	connectStreamingContentTypePrefix = "application/connect+"
+	connectUnaryContentTypePrefix      = "application/"
+	connectUnaryContentTypeJSON        = connectUnaryContentTypePrefix + "json"
+	connectStreamingContentTypeDefault = "application/connect"
+	connectStreamingContentTypePrefix  = "application/connect+"
```
akshayjshah, line 52: Do we need these changes?

akshayjshah (commented):
I _really_ like the direction here - the user-facing API is super nice and clean! It's also much more decoupled from the internals of `connect-go` than I expected, which is great - I'd love to keep this PR open until we've got `github.com/connectrpc` ready to go, and then put this code into a separate repository.

A few notes on the code:

- Please follow the Go style guide. (Search Notion for "Code Standards.")
- Can we use the content-length header to buffer less when handling unary Connect RPCs?

grpcadapter/adapter.go
```diff
@@ -0,0 +1,770 @@
...
+		for _, element := range valElements {
+			pos := strings.IndexByte(element, ';')
+			if pos != -1 {
+				element = element[:pos]
+			}
+			if strings.ToLower(element) == "trailers" {
+				return // "trailers" is already present
+			}
+		}
+	}
```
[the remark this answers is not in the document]
emcfarlane: I didn't know this package existed!

grpcadapter/adapter.go
```diff
@@ -0,0 +1,770 @@
...
+	return http.HandlerFunc(func(responseWriter http.ResponseWriter, request *http.Request) {
+		method := request.Method
+		header := request.Header
+		contentType := getHeaderCanonical(header, headerContentType)
+
+		if _, ok := responseWriter.(http.Flusher); !ok {
+			msg := "gRPC requires a ResponseWriter supporting http.Flusher"
+			http.Error(responseWriter, msg, http.StatusInternalServerError)
+			return
+		}
```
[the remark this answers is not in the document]
emcfarlane: Not sure I understand what you mean here. I can create a response controller but how does one test without calling the `FlushError()` method?

grpcadapter/adapter.go
emcfarlane: Mostly to move error handling to `io.Reader`. Simplifies path between GET and POST.

grpcadapter/adapter.go
emcfarlane: Can skip, but simpler to ignore. Will look at adding.

grpcadapter/adapter.go
```diff
@@ -0,0 +1,770 @@
...
+
+func (w *grpcWebResponseWriter) writeHeaders() {
+	header := w.Header()
+
+	isTrailer := map[string]bool{
+		// Force GRPC status values to be trailers.
+		grpcHeaderStatus:  true,
+		grpcHeaderDetails: true,
+		grpcHeaderMessage: true,
+	}
```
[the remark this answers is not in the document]
emcfarlane: 😬 I like my bool maps https://gist.github.com/davecheney/3be245c92b61e5045f75#gistcomment-2671632 it's slower and memory diff is minimal for structs. Plus nicer to do `if isTrailer[val] {}` then `if _, ok := isTrailer[val]; ok {}` in this small case setting. Will match style though!

grpcadapter/adapter.go
```diff
@@ -0,0 +1,770 @@
...
+
+func (w *grpcWebResponseWriter) writeTrailer() error {
+	header := w.Header()
+	trailer := make(http.Header, len(header)-len(w.seenHeaders))
+	for key, values := range header {
+		if w.seenHeaders[key] {
+			continue
+		}
+		key = strings.TrimPrefix(key, http.TrailerPrefix)
+		// https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md#protocol-differences-vs-grpc-over-http2
```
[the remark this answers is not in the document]
emcfarlane: I just read through that PR the other day haha

grpcadapter/adapter.go
emcfarlane: Fine for streaming, not for unary. Fixed.

grpcadapter/adapter.go
```diff
@@ -0,0 +1,770 @@
...
+func (w *connectUnaryResponseWriter) Flush() {}
+
+func (w *connectUnaryResponseWriter) Unwrap() http.ResponseWriter {
+	return w.ResponseWriter
+}
+
+func flushResponseWriter(responseWriter http.ResponseWriter) {
+	if flusher, ok := responseWriter.(http.Flusher); ok {
+		flusher.Flush()
+	}
```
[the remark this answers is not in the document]
emcfarlane: Removed the check.

grpcadapter/adapter.go
```diff
@@ -0,0 +1,770 @@
...
+}
+
+func newGRPCWebResponseWriter(responseWriter http.ResponseWriter, contentType string) *grpcWebResponseWriter {
+	return &grpcWebResponseWriter{
+		ResponseWriter: responseWriter,
+		contentType:    contentType,
+	}
+}
+
+func (w *grpcWebResponseWriter) writeHeaders() {
```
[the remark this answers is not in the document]
emcfarlane: I think I've resolved this.

grpcadapter/adapter.go
```diff
@@ -0,0 +1,770 @@
...
+		}
+		key = strings.TrimPrefix(key, http.TrailerPrefix)
+		// https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md#protocol-differences-vs-grpc-over-http2
+		trailer[strings.ToLower(key)] = values
+	}
+
+	var buf bytes.Buffer
+	if err := trailer.Write(&buf); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
emcfarlane: Replicated buffer pool.

timruffles:
What's the status of this feature, and any indication on which version it'll appear in?

akshayjshah:
We're splitting it out into a separate project, which we'll release under the https://github.com/connectrpc organization. It's being actively worked on by @emcfarlane and @jhump, so it's coming as soon as we get something really nice ready!

timruffles:
That's great. Do you know if it will support existing unary server interceptors? We have a lot of existing interceptors we'd like to avoid having to rewrite or maintain 2x copies of.

timruffles:
Is there somewhere we could follow along the development of the forthcoming package? We're getting close to an internal decision point about technology, and having this available would really help. The number of internal existing golang gRPC services we have implemented means it wouldn't make sense to use connect without something like this.

timruffles:
Also - we'd be very happy to beta-test! 🥳

emcfarlane:
Hey @timruffles, beta testing would be great! Will get something to you soon.

PR #529, opened 2023-06-22
Export max URL size option

jhump:
Was this unexported on purpose? This seems like a very valuable option for anyone that wants to actually use HTTP GET requests. Also, it was fully specified as if it was exported, even though it was unexported and used by a single test case.


emcfarlane (approved):

akshayjshah (approved):
LGTM - if we're already running into a need for this, let's export it.

PR #530, opened 2023-06-26
CORS object to support web servers

emcfarlane:
Adds a new method `CORSDefault()` which returns a `CORS` object to help setup CORS with Connect and supported protocol headers. Can be used as:

```go
corsSettings := connect.CORSDefaults()
cors.New(cors.Options{
	AllowedOrigins: []string{"https://example.com"},
	AllowedMethods: corsSettings.AllowedMethods,
	AllowedHeaders: corsSettings.AllowedHeaders,
	ExposedHeaders: append(corsSettings.ExposedHeaders, "Trailer-Custom-Response-Trailer"),
	MaxAge:         1024,
})
```

Supports https://github.com/bufbuild/connect-go/issues/499

### DIFF
--- a/cors.go
+++ b/cors.go
@@ -1,0 +1,85 @@
+// Copyright 2021-2023 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package connect
+
+import (
+	"net/http"
+)
+
+// CORS config to help implement Cross-Origin Resource Sharing (CORS) header support.
+type CORS struct {
+	// AllowedMethods that scripts running in the browser are permitted to use.
+	//
+	// To support cross-domain requests with the protocols supported by Connect,
+	// these headers fields must be included in the preflight response header
+	// Access-Control-Allow-Methods.
+	AllowedMethods []string
+
+	// AllowedHeader fields that scripts running in the browser are permitted
+	// to send.
+	//
+	// To support cross-domain requests with the protocols supported by Connect,
+	// these field names must be included in the preflight response header
+	// Access-Control-Allow-Headers.
+	//
+	// Make sure to include any application-specific headers your browser client
+	// may send.
+	AllowedHeaders []string
+
+	// ExposedHeader fields that scripts running the browser are permitted
+	// to see.
+	//
+	// To support cross-domain requests with the protocols supported by Connect,
+	// these field names must be included in header Access-Control-Expose-Headers
+	// of the actual response.
+	//
+	// Make sure to include any application-specific headers your browser client
+	// should see. If your application uses trailers, they will be sent as header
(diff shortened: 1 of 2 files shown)


cors.go
```diff
@@ -0,0 +1,85 @@
...
+	return CORS{
+		AllowedMethods: []string{http.MethodGet, http.MethodPost},
+		AllowedHeaders: []string{
+			headerContentType,
+			connectHeaderProtocolVersion,
+			connectHeaderTimeout,
+			connectUnaryHeaderCompression,           // Unused in web browsers, but added for future-proofing
+			connectUnaryHeaderAcceptCompression,     // Unused in web browsers, but added for future-proofing
+			connectStreamingHeaderCompression,       // Unused in web browsers, but added for future-proofing
+			connectStreamingHeaderAcceptCompression, // Unused in web browsers, but added for future-proofing
```
akshayjshah, line 71: The streaming headers aren't used by web browsers, but the unary headers absolutely are :)

cors.go
```diff
@@ -0,0 +1,85 @@
...
+			grpcHeaderTimeout,                       // Used for gRPC and gRPC-web
+			grpcHeaderMessageType,                   // Unused in web browsers, but added for future-proofing
```
akshayjshah, line 73: This header is weird. It's part of the gRPC protocol specification, but it seems to be unused _everywhere_ - nothing web-specific. I'd document it like so: "unused in practice, but part of gRPC protocol"

cors.go
```diff
@@ -0,0 +1,85 @@
...
+			grpcWebHeaderXGrpcWeb,                   // Used for gRPC-web
+			grpcWebHeaderXUserAgent,                 // Used for gRPC-web
+		},
+		ExposedHeaders: []string{
+			connectUnaryHeaderCompression,     // Unused in web browsers, but added for future-proofing
+			connectStreamingHeaderCompression, // Unused in web browsers, but added for future-proofing
+			grpcHeaderStatus,                  // Required for gRPC-web
+			grpcHeaderMessage,                 // Required for gRPC-web
+			grpcHeaderDetails,                 // Error details in gRPC, gRPC-web
```
akshayjshah, line 82: In general, I'd remove references to gRPC from the comments and focus on gRPC-Web. CORS is only relevant for browser support, and gRPC is unusable from browsers.

cors.go
```diff
@@ -0,0 +1,85 @@
...
+	// protocols.
+	ExposedHeaders []string
+}
+
+// CORSDefault returns a new CORS config object with the default required
+// configuration for Connect and supported protocols.
+//
+// Make sure to include application-specific headers that your application
+// uses as well in addition to these default headers.
+func CORSDefault() CORS {
```
akshayjshah, line 61: Think this'll fit better with the rest of the package as `func NewDefaultCORS` or `func NewCORSDefault`.

cors.go
```diff
@@ -0,0 +1,85 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package connect
+
+import (
+	"net/http"
+)
+
+// CORS config to help implement Cross-Origin Resource Sharing (CORS) header support.
```
akshayjshah, line 21: Can we integrate this a little more closely with the HTML docs? Maybe we should include a link to the docs right now, and plan to update the docs to use this code once it's included in a release.

akshayjshah (changes requested):
I like the shape of this PR, but really want it to live in the new Github org in a separate package. CORS isn't a super-common concern in Go, and this functionality doesn't depend on anything except a few string constants. Let's try to keep the main Connect package as lean as possible so it's not utterly unmanageable in 5y.

emcfarlane:
Closing, will reopen as a separate package.

PR #532, opened 2023-06-26
Simplify code when pooled buffers aren't returned

pkwarren:
In some error handling cases, we're retrieving a value from the sync.Pool and explicitly not returning it. Instead, we might simplify things and just allocate a buffer (since we already know the size) and use it and not put extra work on the pool to garbage collect objects that are retrieved but never returned.

### DIFF
--- a/envelope.go
+++ b/envelope.go
@@ -171,14 +171,11 @@ func (r *envelopeReader) Unmarshal(message any) *Error {
 		// stream. Save the message for protocol-specific code to process and
 		// return a sentinel error. Since we've deferred functions to return env's
 		// underlying buffer to a pool, we need to keep a copy.
+		copiedData := make([]byte, data.Len())
+		copy(copiedData, data.Bytes())
 		r.last = envelope{
-			Data:  r.bufferPool.Get(),
+			Data:  bytes.NewBuffer(copiedData),
 			Flags: env.Flags,
-		}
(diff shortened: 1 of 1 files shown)


envelope.go
```diff
@@ -171,15 +171,12 @@ func (r *envelopeReader) Unmarshal(message any) *Error {
 		// stream. Save the message for protocol-specific code to process and
 		// return a sentinel error. Since we've deferred functions to return env's
 		// underlying buffer to a pool, we need to keep a copy.
+		copiedData := make([]byte, data.Len())
```
pkwarren, line 174: Just a small optimization after re-reading our buffer pool logic.

jchadwick-buf (approved):
Thinking about it, this seems logical enough to me. If the buffer is never returned, then it's effectively just deferring the allocation to later in most cases. Right? So this is not really making the GC/allocation load any worse, it's just changing when the alloc inevitably happens. (Maybe in some cases that is _still_ desired?)

In addition, I think this particular make+copy pattern is optimized by gc and makes it marginally faster than other options for copying data. Though, it could be getting pessimized by the function call to `data.Bytes()` happening during the `copy`. Given that this is in a hot path, it might just be worth looking into.

akshayjshah (approved):
Makes sense to me too!

PR #533, opened 2023-06-27
Capture unexpected io.EOF errors as io.ErrUnexpectedEOF

emcfarlane:
On [Unmarshal](https://github.com/bufbuild/connect-go/blob/929c254ef86fbd5a131f0740fff25971fc65befa/envelope.go#L235-L239) we return `io.EOF` if there is no envelope header. The `errSpecialEnvelope` wraps `io.EOF` and is used in client code to check for [stream errors](https://github.com/bufbuild/connect-go/blob/929c254ef86fbd5a131f0740fff25971fc65befa/client_stream.go#L135-L137). To distinguish between them we can instead check on `Receive` streams if the error is not a known `errSpecialEnvelope`  and return an `io.ErrUnexpectedEOF`.

Fixes https://github.com/bufbuild/connect-go/issues/397

### DIFF
--- a/connect_ext_test.go
+++ b/connect_ext_test.go
@@ -2079,6 +2079,46 @@ func TestHandlerReturnsNilResponse(t *testing.T) {
 	assert.Equal(t, panics, 2)
 }
 
+func TestStreamUnexpectedEOF(t *testing.T) {
+	t.Parallel()
+	mux := http.NewServeMux()
+	mux.HandleFunc("/connect.ping.v1.PingService/CountUp", func(responseWriter http.ResponseWriter, request *http.Request) {
+		io.Copy(io.Discard, request.Body) //nolint:errcheck
+		header := responseWriter.Header()
+		header["content-type"] = []string{"application/connect+json"}
+		responseWriter.WriteHeader(http.StatusOK)
+		head := [5]byte{0, 0, 0, 0, 0}
+		payload := []byte(`{"number": 42}`)
+		binary.BigEndian.PutUint32(head[1:], uint32(len(payload)))
+		responseWriter.Write(head[:]) //nolint:errcheck
+		responseWriter.Write(payload) //nolint:errcheck
+	})
+	server := httptest.NewUnstartedServer(mux)
+	server.EnableHTTP2 = true
+	server.StartTLS()
+	t.Cleanup(server.Close)
+
+	client := pingv1connect.NewPingServiceClient(
+		server.Client(),
+		server.URL,
+		connect.WithProtoJSON(),
+	)
+	t.Run("stream_unexpected_eof", func(t *testing.T) {
+		t.Parallel()
+		const upTo = 2
+		request := connect.NewRequest(&pingv1.CountUpRequest{Number: upTo})
+		request.Header().Set(clientHeader, headerValue)
+		stream, err := client.CountUp(context.Background(), request)
+		assert.Nil(t, err)
+		for stream.Receive() {
+			assert.Equal(t, stream.Msg().Number, 42)
+		}
+		assert.NotNil(t, stream.Err())
+		assert.Equal(t, connect.CodeOf(stream.Err()), connect.CodeUnknown)
+		assert.True(t, errors.Is(stream.Err(), io.ErrUnexpectedEOF))
+	})
+}
+
 // TestBlankImportCodeGeneration tests that services.connect.go is generated with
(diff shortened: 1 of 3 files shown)


jchadwick-buf (approved):
LGTM!

Do we have test cases today that cover other early-EOF scenarios? e.g. what about if we stop part-way through a frame?

connect_ext_test.go
```diff
@@ -2079,6 +2079,46 @@ func TestHandlerReturnsNilResponse(t *testing.T) {
 	assert.Equal(t, panics, 2)
 }
 
+func TestStreamUnexpectedEOF(t *testing.T) {
+	t.Parallel()
+	mux := http.NewServeMux()
+	mux.HandleFunc("/connect.ping.v1.PingService/CountUp", func(responseWriter http.ResponseWriter, request *http.Request) {
+		io.Copy(io.Discard, request.Body) //nolint:errcheck
+		header := responseWriter.Header()
+		header["content-type"] = []string{"application/connect+json"}
```
jchadwick-buf, line 2088: nit: Doesn't practically matter, but this is cleaner in my opinion.
```suggestion
		header.Set("Content-Type", "application/connect+json")
```

emcfarlane:
@jchadwick-buf we do handle partial frames, this seems like an issue with request streams being able to close without a flag message but response streams should always have a final flagged end response.

connect_ext_test.go
```diff
@@ -2079,6 +2080,108 @@ func TestHandlerReturnsNilResponse(t *testing.T) {
 	assert.Equal(t, panics, 2)
 }
 
+func TestStreamUnexpectedEOF(t *testing.T) {
+	t.Parallel()
+	testcases := map[string]http.HandlerFunc{
+		"stream_unexpected_eof": func(responseWriter http.ResponseWriter, request *http.Request) {
+			_, _ = io.Copy(io.Discard, request.Body)
```
jchadwick-buf, line 2087: Thanks for adding additional tests!
Sorry, but just one last thought: this could be de-duplicated a bit by moving the first four lines of each handler into the `HandleFunc` call before calling `testcase`. Then, since `payload` and `head` never actually change, it should also be safe to move them above. Then it'd look a bit more like this:

```go
	payload := []byte(`{"number": 42}`)
	head := [5]byte{}
	binary.BigEndian.PutUint32(head[1:], uint32(len(payload)))

	testcases := map[string]http.HandlerFunc{
		"stream_unexpected_eof": func(responseWriter http.ResponseWriter, request *http.Request) {
			_, _ = responseWriter.Write(head[:])
			_, _ = responseWriter.Write(payload)
		},
		"stream_partial_payload": func(responseWriter http.ResponseWriter, request *http.Request) {
			_, _ = responseWriter.Write(head[:])
			_, _ = responseWriter.Write(payload[:len(payload)-1])
		},
		"stream_partial_frame": func(responseWriter http.ResponseWriter, request *http.Request) {
			_, _ = responseWriter.Write(head[:4])
		},
	}
```

WDYT?

jchadwick-buf (approved):

connect_ext_test.go
emcfarlane: Could group the handler and test logic together?
```go
	testcases := []struct {
		name    string
		handler http.HandlerFunc
		runner  func(t *testing.T)
	}{{
		name: "stream_unexpected_eof",
		handler: func(responseWriter http.ResponseWriter, request *http.Request) {
			_, _ = responseWriter.Write(head[:])
			_, _ = responseWriter.Write(payload)
		},
		runner: func(t *testing.T) {
			const upTo = 2
			request := connect.NewRequest(&pingv1.CountUpRequest{Number: upTo})
			request.Header().Set("Test-Case", t.Name())

			stream, err := client.CountUp(context.Background(), request)
			assert.Nil(t, err)
			for stream.Receive() {
				assert.Equal(t, stream.Msg().Number, 42)
			}
			assert.NotNil(t, stream.Err())
			assert.Equal(t, connect.CodeOf(stream.Err()), connect.CodeUnknown)
			assert.True(t, errors.Is(stream.Err(), io.ErrUnexpectedEOF))
		},
	}}
```

akshayjshah:
Is the error code the same for all three RPC protocols? I think we'll end up with `CodeInternal` for gRPC and `CodeUnknown` for Connect. Can we expand the test coverage to include all three protocols?

PR #536, opened 2023-06-27
Drain stream and error on trailing data

emcfarlane:
For connect and gRPC-web on receiving end of stream payloads ensure no extra data is written by draining the reader and erroring on more data.

Extension of https://github.com/bufbuild/connect-go/pull/533

Fixes https://github.com/bufbuild/connect-go/issues/427

### DIFF
--- a/connect_ext_test.go
+++ b/connect_ext_test.go
@@ -2259,12 +2259,9 @@ func TestStreamUnexpectedEOF(t *testing.T) {
 			request.Header().Set("Test-Case", t.Name())
 			stream, err := client.CountUp(context.Background(), request)
 			assert.Nil(t, err)
-			for i := 0; stream.Receive() && i < upTo; i += 1 {
+			for i := 0; stream.Receive() && i < upTo; i++ {
 				assert.Equal(t, stream.Msg().Number, 42)
 			}
-			t.Log(stream.Err())
-			closeErr := stream.Close()
-			t.Log(closeErr)
 			assert.NotNil(t, stream.Err())
 			assert.Equal(t, connect.CodeOf(stream.Err()), testcase.expectCode)
 			assert.Equal(t, stream.Err().Error(), testcase.expectMsg)
--- a/connect_ext_test.go
+++ b/connect_ext_test.go
@@ -2138,6 +2138,20 @@ func TestStreamUnexpectedEOF(t *testing.T) {
 		},
 		expectCode: connect.CodeInvalidArgument,
 		expectMsg:  "invalid_argument: protocol error: incomplete envelope: unexpected EOF",
+	}, {
+		name: "stream_excess_eof",
+		handler: func(responseWriter http.ResponseWriter, request *http.Request) {
+			_, _ = responseWriter.Write(head[:])
+			_, _ = responseWriter.Write(payload)
+			// Write EOF
+			_, _ = responseWriter.Write([]byte{2, 0, 0, 0, 2})
+			_, _ = responseWriter.Write([]byte("{}"))
+			// Excess payload
+			_, _ = responseWriter.Write(head[:])
+			_, _ = responseWriter.Write(payload)
+		},
+		expectCode: connect.CodeUnknown,
+		expectMsg:  fmt.Sprintf("unknown: corrupt response: %d extra bytes after end of stream", len(payload)+len(head)),
 	}}
 	for _, testcase := range testcases {
 		testcaseMux[t.Name()+"/"+testcase.name] = testcase.handler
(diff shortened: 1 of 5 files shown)


protocol_grpc.go
```diff
@@ -326,7 +326,7 @@ func (g *grpcClient) NewConn(
 	} else {
 		conn.readTrailers = func(_ *grpcUnmarshaler, call *duplexHTTPCall) http.Header {
 			// To access HTTP trailers, we need to read the body to EOF.
-			_ = discard(call)
+			_, _ = discard(call)
```
emcfarlane, line 329: Should gRPC error here?

emcfarlane: Removed.

jchadwick-buf (approved):
Seems reasonable enough to me!

emcfarlane:
Block on https://github.com/bufbuild/connect-go/pull/539 need to expand these testcases for grpc and grpc-web.

protocol_grpc.go
```diff
@@ -325,8 +325,6 @@ func (g *grpcClient) NewConn(
 		}
 	} else {
 		conn.readTrailers = func(_ *grpcUnmarshaler, call *duplexHTTPCall) http.Header {
-			// To access HTTP trailers, we need to read the body to EOF.
-			_ = discard(call)
 			return call.ResponseTrailer()
 		}
```
akshayjshah, line 329: I'm not sure that it's safe to remove this: if the client is using the gRPC protocol, the code you've added to `envelope.go` won't run (because end-of-stream data isn't sent in the body). I think we'll want the `readTrailers` function to also return an error, so we can error out here if the client hasn't read the whole stream or there are I/O errors.

Should all become clearer once you've got tests in place though :)

connect_ext_test.go
```diff
@@ -2138,6 +2138,20 @@ func TestStreamUnexpectedEOF(t *testing.T) {
...
+			_, _ = responseWriter.Write(payload)
+			// Write EOF
+			_, _ = responseWriter.Write([]byte{2, 0, 0, 0, 2})
+			_, _ = responseWriter.Write([]byte("{}"))
+			// Excess payload
+			_, _ = responseWriter.Write(head[:])
+			_, _ = responseWriter.Write(payload)
+		},
+		expectCode: connect.CodeUnknown,
+		expectMsg:  fmt.Sprintf("unknown: corrupt response: %d extra bytes after end of stream", len(payload)+len(head)),
```
akshayjshah, line 2154: As in #539, we'll want tests that cover the same behavior for clients using the gRPC and gRPC-Web protocols. Clients should see the same code and similar (ideally identical) messages regardless of RPC protocol in use.

emcfarlane: Set to internal status code.

protocol_grpc.go
emcfarlane: Reverted, hard to write a test case for it so removed from the excess trailing data test case.

connect_ext_test.go
```diff
@@ -2138,6 +2138,20 @@ func TestStreamUnexpectedEOF(t *testing.T) {
 		},
 		expectCode: connect.CodeInvalidArgument,
 		expectMsg:  "invalid_argument: protocol error: incomplete envelope: unexpected EOF",
+	}, {
+		name: "stream_excess_eof",
+		handler: func(responseWriter http.ResponseWriter, request *http.Request) {
+			_, _ = responseWriter.Write(head[:])
+			_, _ = responseWriter.Write(payload)
+			// Write EOF
+			_, _ = responseWriter.Write([]byte{2, 0, 0, 0, 2})
```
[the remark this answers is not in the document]
emcfarlane: I've used bit shifts to clarify.

connect_ext_test.go
```diff
@@ -2205,6 +2205,42 @@ func TestStreamUnexpectedEOF(t *testing.T) {
...
+			trailer := http.Header{"grpc-status": []string{"0"}}
+			_ = trailer.Write(&buf)
+			var head [5]byte
+			head[0] = 1 << 7
+			binary.BigEndian.PutUint32(head[1:], uint32(buf.Len()))
+			_, _ = responseWriter.Write(head[:])
+			_, _ = responseWriter.Write(buf.Bytes())
+			// Excess payload
+			_, _ = responseWriter.Write(head[:])
+			_, _ = responseWriter.Write(payload)
```
akshayjshah, line 2240: Please check all these errors, asserting that they're `nil`.

akshayjshah (approved):
One test nit, otherwise LGTM! Please add the error checks, but then this is good to land from my perspective.

PR #539, opened 2023-06-29
Surface io.EOF errors as io.ErrUnexpectedEOF for gRPC

emcfarlane:
Cover the gRPC and gRPC-Web protocols in the tests for `io.ErrUnexpectedEOF`, and ensure that error codes and observable `errors.Is` behavior is consistent for all 3 supported RPC protocols.

### DIFF
--- a/connect_ext_test.go
+++ b/connect_ext_test.go
@@ -704,7 +704,7 @@ func TestGRPCMissingTrailersError(t *testing.T) {
 		var connectErr *connect.Error
 		ok := errors.As(err, &connectErr)
 		assert.True(t, ok)
-		assert.Equal(t, connectErr.Code(), connect.CodeInternal)
+		assert.Equal(t, connectErr.Code(), connect.CodeUnknown)
 		assert.True(
 			t,
 			strings.HasSuffix(connectErr.Message(), "protocol error: no Grpc-Status trailer"),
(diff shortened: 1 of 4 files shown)


connect_ext_test.go
```diff
@@ -2093,47 +2093,114 @@ func TestStreamUnexpectedEOF(t *testing.T) {
...
+		name:    "grpc-web_missing_end",
+		options: []connect.ClientOption{connect.WithProtoJSON(), connect.WithGRPCWeb()},
+		handler: func(responseWriter http.ResponseWriter, _ *http.Request) {
+			header := responseWriter.Header()
+			header.Set("Content-Type", "application/grpc-web+json")
+			_, _ = responseWriter.Write(head[:])
+			_, _ = responseWriter.Write(payload)
+		},
+		expectCode: connect.CodeInternal,
+		expectMsg:  "internal: gRPC protocol error: no Grpc-Status trailer",
```
emcfarlane, line 2144: Should this be `internal: gRPC-Web protocol error: no Grpc-Status trailer`?

jchadwick-buf: I believe the reason that it isn't is because internally gRPC and gRPC-web are nearly the same codepath, so the message just comes from the same line of code.

emcfarlane: Yeah, I'm not sure if it's worth changing but wanted to flag. Will resolve for now.

akshayjshah: Probably fine to change to "internal: protocol error: no Grpc-Status trailer"

akshayjshah (approved):
Looks great! Thanks for making all the protocols behave consistently, and sorry for being a bit of a stickler.

PR #542, opened 2023-07-05
Set GetBody on unary requests to enable retries

emcfarlane:
Buffers writes on unary requests to a buffer set for the `request.GetBody`.

TODO: recalim the buffer on close. Need to ensure no more reads would be processed after. Currently discards the fetched buffer.

Closes https://github.com/bufbuild/connect-go/issues/541

### DIFF
--- a/duplex_http_call.go
+++ b/duplex_http_call.go
@@ -47,8 +47,10 @@ type duplexHTTPCall struct {
 	sendRequestOnce sync.Once
 	responseReady   chan struct{}
 	request         *http.Request
-	requestBodyBuf  *bytes.Buffer
 	response        *http.Response
+
+	requestBodyMu  sync.Mutex
+	requestBodyBuf *bytes.Buffer
 
 	errMu sync.Mutex
 	err   error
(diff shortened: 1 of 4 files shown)


emcfarlane:
Closing, I don't think this locking strategy is safe. Need the writes and reads to the buffer to be synced with copying the buffer. We could delay all writes until close, then having a static buffer call `GetBody` with a `bytes.NewReader`.

duplex_http_call.go
```diff
@@ -265,6 +276,19 @@ func (d *duplexHTTPCall) makeRequest() {
 	if d.onRequestSend != nil {
 		d.onRequestSend(d.request)
 	}
+	// If we're sending a unary request, we need to buffer the request body so
+	// that we can send it again if we need to retry.
+	if d.streamType == StreamTypeUnary {
```
jhump, line 281: The exact same error could happen for streaming calls, too. So I don't think we necessarily want to exclude other kinds of RPCs.

Also, I don't think we want to do so much copying. I think we could instead stash entire `*bytes.Buffer` instances and save them (or defer releasing them back to buffer pool) until we decide we've written too much or the operation completes successfully.

Looking through the code, I think a fix to this problem might be rather intrusive, and we'd want to make sure any buffering isn't impacting the performance in a material way. Perhaps it's worth some discussion in the ticket (or a small design doc in Notion)?

PR #546, opened 2023-07-10
Separate Envelope writing from marshalling

emcfarlane:
A new API for working with envelope messages. Envelopes are built up within a single buffer. We can now use a single write call and pass a bytes.Buffer as the full payload.

With this change marshallers no longer write to the stream. Control is given to the protocols connection handler. This allows the protocols implementation to easily customize the behaviour. However, it does duplicates some of the flow. Overall I think it's simpler to duplicate control flow into the connection handler to allow for protocol specific features. Changes:
- Connect unary now use `marshal` to gain `marshalAppend` codecs: https://github.com/bufbuild/connect-go/blob/6183cbf6ff174060dbdc494095ed7934dee4bc7d/protocol_connect.go#L907
- No longer need to partially build envelopeWriters, create the envelope and write it instead: https://github.com/bufbuild/connect-go/blob/6183cbf6ff174060dbdc494095ed7934dee4bc7d/error_writer.go#L133-L138
- `errSpecialEnvelope` is no longer needed, protocols check for end flags directly: https://github.com/bufbuild/connect-go/blob/6183cbf6ff174060dbdc494095ed7934dee4bc7d/envelope.go#L28
- gRPC read trailer functions can be moved into the `Receive()` functions: https://github.com/bufbuild/connect-go/blob/6183cbf6ff174060dbdc494095ed7934dee4bc7d/protocol_grpc.go#L323
-  Easier to avoid leaking abstractions from one protocol to the other. Error messages are set in the handler: https://github.com/bufbuild/connect-go/blob/6183cbf6ff174060dbdc494095ed7934dee4bc7d/envelope.go#L190
- Small reduction in allocs/op:
```diff
- BenchmarkConnect/unary-8                     583           2035062 ns/op         5448544 B/op        232 allocs/op
+ BenchmarkConnect/unary-8                     580           2054063 ns/op         5190420 B/op        221 allocs/op
```

### DIFF
--- a/connect.go
+++ b/connect.go
@@ -364,7 +364,7 @@ func receiveUnaryResponse[T any](conn StreamingClientConn) (*Response[T], error)
 	// In a well-formed stream, the response message may be followed by a block
 	// of in-stream trailers or HTTP trailers. To ensure that we receive the
 	// trailers, try to read another message from the stream.
-	if err := conn.Receive(new(T)); err == nil {
+	if err := conn.Receive(&msg); err == nil {
 		return nil, NewError(CodeUnknown, errors.New("unary stream has multiple messages"))
 	} else if err != nil && !errors.Is(err, io.EOF) {
 		return nil, NewError(CodeUnknown, err)
--- a/envelope.go
+++ b/envelope.go
@@ -19,269 +19,176 @@ import (
 	"encoding/binary"
 	"errors"
 	"io"
+	"math"
 )
 
 // flagEnvelopeCompressed indicates that the data is compressed. It has the
 // same meaning in the gRPC-Web, gRPC-HTTP2, and Connect protocols.
 const flagEnvelopeCompressed = 0b00000001
 
-var errSpecialEnvelope = errorf(
-	CodeUnknown,
-	"final message has protocol-specific flags: %w",
-	// User code checks for end of stream with errors.Is(err, io.EOF).
-	io.EOF,
-)
-
-// envelope is a block of arbitrary bytes wrapped in gRPC and Connect's framing
-// protocol.
-//
-// Each message is preceded by a 5-byte prefix. The first byte is a uint8 used
-// as a set of bitwise flags, and the remainder is a uint32 indicating the
-// message length. gRPC and Connect interpret the bitwise flags differently, so
-// envelope leaves their interpretation up to the caller.
-type envelope struct {
-	Data  *bytes.Buffer
-	Flags uint8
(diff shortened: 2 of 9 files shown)


connect.go
```diff
@@ -364,7 +364,7 @@ func receiveUnaryResponse[T any](conn StreamingClientConn) (*Response[T], error)
 	// In a well-formed stream, the response message may be followed by a block
 	// of in-stream trailers or HTTP trailers. To ensure that we receive the
 	// trailers, try to read another message from the stream.
-	if err := conn.Receive(new(T)); err == nil {
```
emcfarlane, line 367: I think we can remove this extra Receive call with this protocol drain fix: https://github.com/bufbuild/connect-go/pull/536

akshayjshah: Awesome. That'd also let us make the function non-generic, which would cut binary size a bit.

envelope.go
```diff
@@ -40,248 +38,168 @@ var errSpecialEnvelope = errorf(
...
+	Buffer *bytes.Buffer
 }
 
-func (e *envelope) IsSet(flag uint8) bool {
-	return e.Flags&flag == flag
+// makeEnvelope creates an envelope with a 5-byte prefix.
+func makeEnvelope(buffer *bytes.Buffer) envelope {
+	var head [5]byte
+	buffer.Write(head[:])
+	return envelope{Buffer: buffer}
```
akshayjshah, line 48: With these changes, is `envelope` doing enough work to keep around? It doesn't seem to provide much value as a wrapper on top of `bytes.Buffer`, and I don't love that the `unwrap` method isn't safe to call twice.

Would the rest of the repository be more clear if we instead had something like this (and managed the envelope and message buffer separately)?

```go
type envelope [5]byte

func (e *envelope) Flags() uint8 { ... }
func (e *envelope) SetFlags(uint8) { ... }
func (e *envelope) HasFlags(uint8) { ... } // check if flags are set
func (e *envelope) Size() int { ... } // read size from last 4 bytes
func (e *envelope) SetSize(int) { ... } // write size to last 4 bytes
```

akshayjshah (commented):
This is ... a lot to get through in one sitting 😅 Most important questions:

* Is the new `envelope` type the best we can do? (Asked inline on the type.)
* How can we break this change up into more easily-reviewed chunks? As is, it's _really_ difficult to substantively review.

emcfarlane:
Closing, will look at breaking apart!

PR #547, opened 2023-07-13
Track external tool versions in go.mod

pkwarren:
Add an internal/tools module with its own dependencies on buf and golangci-lint. Keep tools up to date automatically with dependabot. Update project to ignore GoLand/IDEA project files.


bufdev:
Fwiw, I'm not terribly in favor of this, I think the cost/benefit in terms of complication and updates is high vs any benefit we get is high

pkwarren:
> Fwiw, I'm not terribly in favor of this, I think the cost/benefit in terms of complication and updates is high vs any benefit we get is high

Ok - it is a pattern I've used mainly to make sure we remember to keep tools updated (since we were still on buf 1.18.0 / golangci-lint 1.52.x). Happy to close though.

bufdev:
I just wanted to provide my two cents - not strictly blocking, but I do think it adds a bunch of complication and I'm not a fan of the blank import style of tool imports - it happens to work right now for all of the tools we use because they're all in Golang, but doesn't work for others, and you have to do a separate module to make it work properly.

PR #556, opened 2023-07-28
Update demo service calls

akshayjshah:
Update the README to call the new demo service correctly.


jhump (approved):

akshayjshah:
@jhump Sadly, we should cut a new release so that this change gets picked up by pkg.go.dev. Can you cut a new patch release?

PR #564, opened 2023-08-15
Use googleapis/rpc status protos

emcfarlane:
Migrates internal status message for gRPC to googleapis/rpc. The googleapis have been refactored into a smaller [module](https://pkg.go.dev/google.golang.org/genproto/googleapis/rpc ) making this a much lighter dep!


bufdev (commented):
I'm not hard against, but I'm very skeptical of this - I don't trust the google.golang.org/genproto stuff to be reliable over time, and it's an extra dependency for connect-go. I don't know, I could be convinced either way.

emcfarlane:
Oh, I missed that. I'll close, makes sense to wait for LTS.

PR #566, opened 2023-08-15
Include Go 1.21.0 in CI

akshayjshah:
Include Go 1.21.0 in CI, and update `golangci-lint` to a version that supports
the new runtime.

As part of this change, we're committing to support for the last 3 versions of
Go. This ends up including a version of Go that no longer receives security
patches, but it matches AWS and some large companies' very slow upgrade cycles.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,8 @@ MAKEFLAGS += --warn-undefined-variables
 MAKEFLAGS += --no-builtin-rules
 MAKEFLAGS += --no-print-directory
 BIN := .tmp/bin
+export PATH := $(BIN):$(PATH)
+export GOBIN := $(abspath $(BIN))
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -7,6 +7,8 @@ MAKEFLAGS += --warn-undefined-variables
 MAKEFLAGS += --no-builtin-rules
 MAKEFLAGS += --no-print-directory
 BIN := .tmp/bin
+export PATH := $(BIN):$(PATH)
+export GOBIN := $(abspath $(BIN))
```
akshayjshah, line 11: In a subsequent PR, I'm planning to rip out our homegrown support for overriding the Go compiler in use. Now that 1.21 is out, users can set `GOTOOLCHAIN` without any special magic in this Makefile.

bufdev (approved):

PR #567, opened 2023-08-16
Drop homegrown $GO Make support

akshayjshah:
Now that Go 1.21 has `GOTOOLCHAIN`, we don't need this homegrown
mechanism anymore.


bufdev (approved):
Much nicer

PR #572, opened 2023-08-25
Set X-User-Agent in gRPC-Web clients

akshayjshah:
The gRPC-Web pseudo-specification isn't totally clear, but it sounds
like it requires setting `X-User-Agent` instead of `User-Agent` as a
part of the protocol. This makes sense in browsers, where JS _can't_ set
`User-Agent`, but it's very odd for backend clients. This PR proposes
splitting the difference and setting both headers.

### DIFF
--- a/connect_ext_test.go
+++ b/connect_ext_test.go
@@ -2020,6 +2020,31 @@ func TestAllowCustomUserAgent(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


timostamm (approved):
> This PR proposes splitting the difference and setting both headers.

Setting both headers seems to be the consequence for a non-browser client.

bufdev (approved):

PR #574, opened 2023-08-25
Eliminate preallocation for enveloped messages

akshayjshah:
For enveloped protocols (gRPC, gRPC-Web, and Connect streaming), this PR stops using the size from the envelope to preallocate buffers. This optimization rarely makes much difference in practice (because we're already pooling the buffers), and it's a potential abuse vector for malicious clients - they can trigger large allocations just by sending a 5-byte envelope.


jchadwick-buf (approved):
LGTM. I think this is the best choice for now.

PR #580, opened 2023-09-02
Try to read full prefix in envelope reader

njiang747:
I believe this fixes a bug where calling to `r.reader.Read` before all 5 bytes of the prefixes are ready causes an unnecessary `ErrUnexpectedEOF` (resulting in a client error `invalid_argument: protocol error: incomplete envelope: unexpected EOF`).

Switching to use `io.ReadFull` guarantees that the `envelopReader` will wait until the full prefixes is available before proceeding. Error handling should be identical to before due to the semantics of `io.ReadFull`. If fewer than 5 bytes are read and then `EOF` is hit, it will return `ErrUnexpectedEOF`. If any other error is hit, it will be returned as expected.

### DIFF
--- a/envelope.go
+++ b/envelope.go
@@ -226,7 +226,7 @@ func (r *envelopeReader) Unmarshal(message any) *Error {
 
 func (r *envelopeReader) Read(env *envelope) *Error {
 	prefixes := [5]byte{}
-	prefixBytesRead, err := r.reader.Read(prefixes[:])
+	prefixBytesRead, err := io.ReadFull(r.reader, prefixes[:])
 
 	switch {
 	case (err == nil || errors.Is(err, io.EOF)) &&
@@ -240,7 +240,7 @@ func (r *envelopeReader) Read(env *envelope) *Error {
 		// to the user so that they know that the stream has ended. We shouldn't
 		// add any alarming text about protocol errors, though.
 		return NewError(CodeUnknown, err)
-	case err != nil || prefixBytesRead < 5:
+	case err != nil:
 		// Something else has gone wrong - the stream didn't end cleanly.
 		if connectErr, ok := asError(err); ok {
 			return connectErr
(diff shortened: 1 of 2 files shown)


emcfarlane:
Hey @njiang747 , I think the fix is a great find! Want to add a test case in `envelope_test.go`?

```go
func TestEnvelope_read(t *testing.T) {
	t.Parallel()

	head := [5]byte{}
	payload := []byte(`{"number": 42}`)
	binary.BigEndian.PutUint32(head[1:], uint32(len(payload)))
	buf := &bytes.Buffer{}
	buf.Write(head[:])
	buf.Write(payload)

	t.Run("full", func(t *testing.T) {
		t.Parallel()
		env := &envelope{Data: &bytes.Buffer{}}
		rdr := envelopeReader{
			reader: bytes.NewReader(buf.Bytes()),
		}
		assert.Nil(t, rdr.Read(env))
		assert.Equal(t, payload, env.Data.Bytes())
	})
	t.Run("byteByByte", func(t *testing.T) {
		t.Parallel()
		env := &envelope{Data: &bytes.Buffer{}}
		rdr := envelopeReader{
			reader: byteByByteReader{
				reader: bytes.NewReader(buf.Bytes()),
			},
		}
		assert.Nil(t, rdr.Read(env))
		assert.Equal(t, payload, env.Data.Bytes())
	})
}

// byteByByteReader is test reader that reads a single byte at a time.
type byteByByteReader struct {
	reader io.ByteReader
}

func (b byteByByteReader) Read(data []byte) (int, error) {
	if len(data) == 0 {
		return 0, nil
	}
	next, err := b.reader.ReadByte()
	if err != nil {
		return 0, err
	}
	data[0] = next
	return 1, nil
}
```
This fails when removing your patch.

jhump (approved):

Issue #541, opened 2023-07-03
define Request.GetBody to avoid this error


pqn:
**Describe the bug**

In some of our production logs we've noticed the following client error:
```
unavailable: http2: Transport: cannot retry err [http2: Transport received Server's graceful shutdown GOAWAY] after Request.Body was written; define Request.GetBody to avoid this error
```

**To Reproduce**

We've been unable to reproduce this locally.

**Environment (please complete the following information):**
- `connect-go` version or commit: `v1.6.0`
- `go version`: `go version go1.20.5 linux/amd64`
- your complete `go.mod`: I can provide dependencies as needed

**Additional context**
- We are using the GRPC protocol, and the server is a Python gRPC server.
- It seems related to this issue: https://github.com/olivere/elastic/issues/1443
- One way this has been fixed in a public repo is here: https://github.com/stripe/stripe-go/pull/711

emcfarlane:
Hey @pqn! Thanks for the detailed breakdown. For unary requests we should be able to set GetBody on requests as we have the body in a `bytes.Buffer`. However we currently write to the body with a pipe: https://github.com/bufbuild/connect-go/blob/main/duplex_http_call.go#L80 
I'll try replicate this with testing and see if we can create a fix for unary requests. In the meantime you might be able to check for this error and retry on the client side. How often do you see this happening?

pqn:
In one log I'm seeing it maybe 4-5 times in an hour.

espadolini:
We see this error 10-40 times per minute in our prod infra out of roughly 2400 requests, so 0.4% to 1.6% of all requests are being retried, which isn't terrible but also not insignificant.

pqn:
I saw #543 was closed. Just wondering, is this being worked on or in the backlog right now? Thanks!

emcfarlane:
Hey @pqn, we are still aiming to fix this issue soon. The feedback on the PR was to focus on the unary changes first as this will have the largest impact. @mattrobenolt has some good suggestion on splitting out the unary flow to get some nice performance improvements too. Will be a much neater solution!

pqn:
Sounds good! Thanks for the update.

emcfarlane:
Closing, solved in #649

jhump:
Resolved in #649, but only for unary-request RPCs (which includes both unary RPCs and server-streaming RPCs). Implementing `GetBody` for streaming requests is more complicated, and not as high of a priority since unary operations are the vast majority of RPC usage.

[...]

PR #586, opened 2023-09-12
Separate envelope writing from marshalling

emcfarlane:
New mechanism for working with envelope messages. Creates new util functions like `writeAll`, `readAll` and `writeEnvelope`, `readEnvelope`. These are used directly by the protocol implementations to write and read messages from the stream. Each protocol can then handle specific flags independently and avoid the sentinel `errSpecialEnvelope` to eject from the envelope marshallers.

PR is a prelude to the fix for https://github.com/connectrpc/connect-go/issues/541 with a new methods on  `duplexHTTPCall` to write messages based on `SendEnvelope` `WriteEnvelope` and changing the buffer pool use ownership. Removing the envelope marshallers means can change the signature for the client and not the server.

- No longer need to partially build envelopeWriters, create the message and call `writeEnvelope `: https://github.com/bufbuild/connect-go/blob/6183cbf6ff174060dbdc494095ed7934dee4bc7d/error_writer.go#L133-L138
- `errSpecialEnvelope` is no longer needed, protocols check for end flags directly: https://github.com/bufbuild/connect-go/blob/6183cbf6ff174060dbdc494095ed7934dee4bc7d/envelope.go#L28
- gRPC read trailer functions can be moved into the Receive() functions: https://github.com/bufbuild/connect-go/blob/6183cbf6ff174060dbdc494095ed7934dee4bc7d/protocol_grpc.go#L323
- Easier to avoid leaking abstractions from one protocol to the other. Error messages are set in the protocol connection: https://github.com/bufbuild/connect-go/blob/6183cbf6ff174060dbdc494095ed7934dee4bc7d/envelope.go#L190
- Small reduction in allocs:
```diff
- BenchmarkConnect/unary-8                     609           1967440 ns/op         5840173 B/op        230 allocs/op
+ BenchmarkConnect/unary-8                     606           1986667 ns/op         5230730 B/op        218 allocs/op
```
Rework of https://github.com/connectrpc/connect-go/pull/546 . Removed `makeEnvelope` and use the method signature for enveloping.


protocol.go
```diff
@@ -40,7 +40,7 @@ const (
 	headerUserAgent   = "User-Agent"
 	headerTrailer     = "Trailer"
 
-	discardLimit = 1024 * 1024 * 4 // 4MiB
+	discardLimit = 1024 * 1024 * 8 // 8MiB
```
emcfarlane, line 43: Test was failing now that discard is used. I've upped this limit to avoid changing the test.

connect_ext_test.go
```diff
@@ -2190,8 +2190,8 @@ func TestStreamUnexpectedEOF(t *testing.T) {
 			_, err = responseWriter.Write(payload[:len(payload)-1])
 			assert.Nil(t, err)
 		},
-		expectCode: connect.CodeInvalidArgument,
-		expectMsg:  fmt.Sprintf("invalid_argument: protocol error: promised %d bytes in enveloped message, got %d bytes", len(payload), len(payload)-1),
+		expectCode: connect.CodeInternal,
+		expectMsg:  "internal: incomplete envelope: unexpected EOF",
```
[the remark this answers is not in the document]
akshayjshah: Can you check what `grpc-go` does here? I think I changed this to `CodeInvalidArgument` specifically to match them, but I may be wrong.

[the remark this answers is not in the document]
emcfarlane: Starts as `io.ErrUnexpectedEOF` and converted to `status.Error(codes.Internal, err.Error())`
- https://github.com/grpc/grpc-go/blob/b0a946cf0cf1adb094d7da9dbf2d054905702c73/rpc_util.go#L619
- https://github.com/grpc/grpc-go/blob/82a568ddbb02ff410a4f7c071ecf8b277992a251/rpc_util.go#L883
Oh I thought this was my wrong doing 😅

protocol_connect.go
```diff
@@ -1312,3 +1254,29 @@ func queryValueReader(data string, base64Encoded bool) io.Reader {
...
+		canonical := http.CanonicalHeaderKey(name)
+		if name != canonical {
+			delete(end.Trailer, name)
+			end.Trailer[canonical] = append(end.Trailer[canonical], value...)
+		}
+	}
+	return &end, nil
+}
+func connectMarshalEndStreamMessage(dst *bytes.Buffer, end *connectEndStreamMessage) *Error {
+	data, marshalErr := json.Marshal(end)
```
emcfarlane, line 1276: Will switch this to `json.NewEncoder`.

envelope.go
```diff
@@ -25,286 +25,177 @@ import (
...
-		return nil
-	case err != nil && errors.Is(err, io.EOF) && prefixBytesRead == 0:
-		// The stream ended cleanly. That's expected, but we need to propagate them
-		// to the user so that they know that the stream has ended. We shouldn't
-		// add any alarming text about protocol errors, though.
-		return NewError(CodeUnknown, err)
-	case err != nil || prefixBytesRead < 5:
+func readEnvelope(dst *bytes.Buffer, src io.Reader, readMaxBytes int) (uint8, *Error) {
+	prefix := [5]byte{}
+	if _, err := io.ReadFull(src, prefix[:]); err != nil {
```
emcfarlane, line 111: Incorporates the fix from: https://github.com/connectrpc/connect-go/pull/580

envelope.go
```diff
@@ -25,286 +25,177 @@ import (
...
+	raw, err := codec.Marshal(message)
 	if err != nil {
 		return errorf(CodeInternal, "marshal message: %w", err)
 	}
-	buffer := bytes.NewBuffer(raw)
-	// Put our new []byte into the pool for later reuse.
-	defer w.bufferPool.Put(buffer)
-	envelope := &envelope{Data: buffer}
-	return w.Write(envelope)
+	dst.Write(raw)
```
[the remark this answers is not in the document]
emcfarlane: Moved to replacing the buffer if it's too small.

envelope.go
jhump: Looks like that contributor never got any feedback. Maybe it would be better to _accept_ (and merge) #580 (so the author gets credit for it in release notes) and then rebase this branch?

emcfarlane: Yep I agree, it's a good fix.

connect_ext_test.go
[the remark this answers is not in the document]
jhump: What direction is this? Returning "internal" from the server, when the _client_ has sent an invalid stream, seems very wrong. Internal should be reserved for internal server errors, not errors in the client (akin to 5xx vs 4xx). You generally want exception reporting/alerting for internal server errors since they indicate a likely bug in your code; you do _not_ want any such alerting for client errors, because then a client can induce alerts in your system w/ carefully crafted (invalid) requests.

So I agree that this should be "internal" only in cases where it's the _server_ that sent the invalid message.

protocol.go
jhump: Which test? I think it might be better to change the test. While 8mb is not a big number, it can be bad combined with a slow client. And until we solve the "read timeout" thing (discussed in separate threads), we should be more conservative. 4mb is already generous since most RPC payloads are single digit kilobytes, so doubling this seems like the wrong play.

protocol.go
```diff
@@ -284,13 +284,22 @@ func isCommaOrSpace(c rune) bool {
...
-	lr := &io.LimitedReader{R: reader, N: discardLimit}
-	return io.Copy(io.Discard, lr)
+	wroteN, err := io.Copy(io.Discard, lreader)
+	if errors.Is(err, io.EOF) {
+		err = nil
+	}
+	if wroteN == limit {
+		// Ensure we error if we hit the limit.
+		err = io.ErrShortBuffer
+	}
```
jhump, line 301: Technically, the limit and the payload could be the same size, in which case this error would be spurious. You'd need to either set the limit reader to `discardLimit+1` or follow up with an attempt to read a single byte to decide on if this error is appropriate.

protocol_connect.go
```diff
@@ -441,15 +372,17 @@ func (c *connectClient) NewConn(
...
-	spec             Spec
-	peer             Peer
-	duplexCall       *duplexHTTPCall
-	compressionPools readOnlyCompressionPools
-	bufferPool       *bufferPool
-	marshaler        connectUnaryRequestMarshaler
-	unmarshaler      connectUnaryUnmarshaler
-	responseHeader   http.Header
-	responseTrailer  http.Header
+	*connectClient
```
jhump, line 375: Why embed? Does this add any additional exported symbols to `connectUnaryClientConn`? If so, that seems bad since a caller could type-assert to gain access to them.

jhump (commented):
Hey, Ed, sorry it took me so long to get you feedback. I've left numerous comments on the diff, but overall I'm really worried about merging a change like this.

This change does too much in a single branch. It's very large and dense. And it's hard to review because the changes involve lots of code motion and re-abstraction/de-abstraction. I think we'd need a very clear motivation to merge such a change, and I haven't heard or seen it. Maybe you can elaborate more on why you think this sort of pre-factoring is needed for the upcoming changes you are trying to make? Maybe there are are less intrusive ways to accomplish the same thing.

The changes in abstractions make the protocol implementations more verbose and more complicated. I'm not sure that's really desirable -- I know you stated the explicitness makes them easier to change in the future, but I think we could still do better towards code re-use/DRY-ness than what's in this branch. The change as is feels a lot like "moving spaghetti around on the plate", changing where certain things are done, but not in a way that is obviously an improvement to readability/maintainability.

Re-factoring changes are best when they are mostly mechanical. But this branch also has some subtle behavioral changes, which are definitely _not_ mechanical. That makes it harder to review. We should separate these aspects so that the behavioral changes can be reviewed/discussed separately.

I think we may need to chat about what you're really _needing_ to accomplish and come up with a less radical formulation. I hope that all makes sense.

emcfarlane:
@jhump thanks for the review. The issue I was trying to solve is the abstraction of the `evelopeWriter` over an `io.Writer` [here](https://github.com/connectrpc/connect-go/blob/4f411c45951b413e2a879fac7e79e11d2341ed71/envelope.go#L52).

Connect clients write to requests by exposing an [`io.Writer`](https://github.com/connectrpc/connect-go/blob/4f411c45951b413e2a879fac7e79e11d2341ed71/duplex_http_call.go#L96) which is passed to a unary or streaming call. For GetBody we want to pass around message buffers instead. This work was based on your [feedback](https://github.com/connectrpc/connect-go/pull/543#discussion_r1255771137) for removing the writer interfaces.

I still think the current envelope abstraction leads to more bugs and unnecessary complexity than it helps. These are the bugs/almost bugs I know of:
- FullPrefix read: https://github.com/connectrpc/connect-go/pull/580 we can add tests to the envlopers but would it's more difficult than removing them.
- Errors are marshalled in multiple locations: https://github.com/connectrpc/connect-go/pull/599#issuecomment-1745649279
- Connect GET requests aren't able to reuse code: https://github.com/connectrpc/connect-go/blob/4f411c45951b413e2a879fac7e79e11d2341ed71/protocol_connect.go#L969-L970
- Same issue here: https://github.com/connectrpc/connect-go/pull/561

I'll try address the feedback. Would be great to chat through a solution that's more surgical, less spaghetti.

jhump:
> Connect clients write to requests by exposing an [io.Writer](https://github.com/connectrpc/connect-go/blob/4f411c45951b413e2a879fac7e79e11d2341ed71/duplex_http_call.go#L96) which is passed to a unary or streaming call. For GetBody we want to pass around message buffers instead.

While `duplexHTTPCall` does expose a `Write` method, that's not the same thing in my mind as "exposing an `io.Writer`" to which Connect clients write requests. In fact, the whole connect client flow -- both unary and stream calls -- actually already speak in terms of messages. So I'm pretty sure that can be translated to message buffers in a much more surgical way.

> This work was based on your feedback for removing the writer interfaces.

What I was actually proposing in the comments on that previous PR was something much more surgical that just replaces the `io.Pipe` and the calls to its reader and writer.

Also, I thought we decided to punt on supporting `GetBody` for streaming calls for now and focus on unary, by separating it from the `duplexHTTPCall`. So I'm failing to see how a change like this is a needed pre-factor for the step of further special-casing unary calls.

> These are the bugs/almost bugs I know of:

All of those can be address by much smaller, targeted fixes instead of a large refactor that happens to touch on all of them at once. Small, targeted changes are easier write, review, and test and are inherently lower risk. While it can be difficult to fix something in a small, target way sometimes, like if a fix really _requires_ a significant architecture change, I don't think that is the case for any of these.

> Would be great to chat through a solution that's more surgical, less spaghetti.

I'll schedule some time to chat more.

emcfarlane:
>  So I'm failing to see how a change like this is a needed pre-factor for the step of further special-casing unary calls.

For connect, unary calls have access to the duplexWriter already so this isn't needed. For gRPC and gRPC-Web both envelope and go through the `envelopeWriter`. The changes I had to support `GetBody` would expose two methods on `duplexHTTPCall`:
```go
func (d *duplexHTTPCall) Send(buf *bytes.Buffer) error
func (d *duplexHTTPCall) SendEnvelope(buf *bytes.Buffer, flags uint8) error
```
Or something similar to this effect.

I'm happy to go with something better.

PR #594, opened 2023-09-18
Create memhttp package to debug flaky testcases

emcfarlane:
Creates two internal packages: `memhttp` and `memhttptest` based on https://github.com/akshayjshah/memhttp . We use this for testing connect's services.

Changes`httptest.Server` to use an in memory network. This makes testing more robust to ephemeral port issues under load. Two flaky test cases were now easily reproducible and fixed. Both occurred from races between starting the network request in a go routine and checking for a http2 stream and erroring if not. This can error on Send or Receive depending on how fast the request can complete.

On `duplexHTTPCall` I removed `SetError` to avoid overwriting the original response error. Now `BlockUntilResponseReady` returns an error from the response initialization.

See: https://github.com/golang/go/issues/14200

Fixes:
- `TestServer/http1/*/*/cumsum_cancel_before_send`: send can error on write message.
-  `TestBidiRequiresHTTP2`: send can `io.EOF` error on write, or succeed.
- `TestBidiOverHTTP1`: same as above.
-  `TestClientPeer/grpcweb`: stream request sends an invalidate payload invoking server error, races with interceptor.

Two places where errors raced:
-  https://github.com/connectrpc/connect-go/blob/525cf76ebd3c1c54f2cfd2cb4413c44b57539be9/duplex_http_call.go#L290
- https://github.com/connectrpc/connect-go/blob/525cf76ebd3c1c54f2cfd2cb4413c44b57539be9/handler.go#L184


emcfarlane:
Running many iterations locally to find more, need to fix another one:
```
--- FAIL: TestClientPeer (0.00s)
    --- FAIL: TestClientPeer/grpcweb (0.01s)
        client_ext_test.go:111:
            assertion:  assert.Nil
            got:        unknown: io: read/write on closed pipe
```

connect_ext_test.go
```diff
@@ -1540,9 +1494,7 @@ func TestStreamForServer(t *testing.T) {
 	newPingServer := func(pingServer pingv1connect.PingServiceHandler) (pingv1connect.PingServiceClient, *httptest.Server) {
```
jhump, line 1494: Looks like many call sites have been updated to ignore the returned server. Is that maybe _all_ of the call sites? If so, let's remove that return value so this function only returns `PingServiceClient`).

jhump (commented):
Looking good so far to me. Are you planning to fix that other flake you found in this PR? Or do you want this PR approved now so you can merge it and fix the other issue in a subsequent PR?

emcfarlane:
@jhump had to step out but hopefully can debug it now. Not sure what is calling close twice.

emcfarlane:
I left it as a TODO. I think it's related to `SetError` but not sure how it rises. I have other branches to rejig the duplex_http_call which should make this easier to debug. Will fix later.

emcfarlane:
Benchmarking increase in allocs, making the `net.Pipe` is ~8 and theres some more locking. Maybe better to use local network for bench. 
```patch
- BenchmarkConnect/unary-8                    5433           1981280 ns/op         5651212 B/op        229 allocs/op
+ BenchmarkConnect/unary-8                    5440           1980366 ns/op         5729820 B/op        237 allocs/op
```

jhump (approved):
> Maybe better to use local network for bench.

Yeah, I think so, too. The benchmark seems less useful if it's intentionally _different_ from real world conditions 🤷

internal/memhttp/memhttp.go
```diff
@@ -0,0 +1,140 @@
...
+		disableHTTP2:   cfg.DisableHTTP2,
+	}
+	server.goServe()
+	return server
+}
+
+// Transport returns a [http.RoundTripper] configured to use in-memory pipes
+// rather than TCP and talk HTTP/2 (if the server supports it).
+func (s *Server) Transport() http.RoundTripper {
+	if s.disableHTTP2 {
```
jhump, line 75: I think this should always return a transport that uses HTTP/2.

To test things that require HTTP 1.1 (like tests in this package to verify that bidi requests gracefully fail with 505 status), we can also have a `TransportHTTP1` method that returns an HTTP/1.1-only transport. I don't think we need a `ClientHTTP1` because a caller can always do this instead:
```go
client := svr.Client()
client.Transport = svr.TransportHTTP1()
```

We'd likely also need this (if/when we export this) for callers that want to do stuff with the server like web sockets (which isn't yet supported on HTTP/2 in the Go runtime or 3rd party web sockets packages).

compression_test.go
```diff
@@ -42,12 +43,10 @@ func TestAcceptEncodingOrdering(t *testing.T) {
 		w.WriteHeader(http.StatusOK)
 		called = true
 	})
-	server := httptest.NewServer(verify)
-	t.Cleanup(server.Close)
-
+	server := memhttptest.NewServer(t, verify, memhttp.WithoutHTTP2())
```
jhump, line 46: Why is this using `WithoutHTTP2` here (and throughout)? I think there are only two places in `connect_ext_test` that actually intend to force HTTP 1.1. Everything else should likely be fine using HTTP/2.

.golangci.yml
```diff
@@ -119,3 +119,12 @@ issues:
     - linters: [revive]
       text: "^if-return: "
       path: error_writer.go
+    # We want to set http.Server's logger
+    - linters: [forbidigo]
+      path: internal/memhttp
+    # We want to set http.Server's logger
+    - linters: [forbidigo]
+      path: internal/memhttp/memhttptest
```
jhump, line 127: I'm pretty sure `path` here is a regex, which means the above rule should already match this path. So I think this one might be redundant and can be removed.

internal/memhttp/option.go
```diff
@@ -0,0 +1,67 @@
...
+
+package memhttp
+
+import (
+	"log"
+	"time"
+)
+
+type config struct {
+	DisableTLS     bool
```
jhump, line 23: Unused

internal/memhttp/option.go
```diff
@@ -0,0 +1,67 @@
...
+	return optionFunc(func(cfg *config) {
+		for _, opt := range opts {
+			opt.apply(cfg)
+		}
+	})
+}
+
+// WithCleanupTimeout customizes the default five-second timeout for the
+// server's Cleanup method. It's most useful with the memhttptest subpackage.
+func WithCleanupTimeout(d time.Duration) Option {
```
jhump, line 56: If we move `config` into an internal package (accessible only to `memhttp` and `memhttptest` packages), then we could move this option into `memhttptest`, so that you can only use it if you import `memhttptest`, which I think makes the final remark in the Go docs unnecessary.

Further, we could make it so that it is only applied in `memhttptest` by having this value actually used in the `t.Cleanup` function, instead of in `svr.Shutdown`.

I think that leaves the API in `memhttp` a little more "pure", and the testing helpers are all isolated to the `memhttptest` package.

compression_test.go
emcfarlane: I was keeping previous behaviour. Will change.

internal/memhttp/option.go
emcfarlane: I've renamed the option to `WithShutdownTimeout`. It's used to set a timer on the context so simplifies users calling shutdown and guarantees the context get's cancelled. I think this is valuable normal server usage, removed the test desc.

internal/memhttp/listener.go
```diff
@@ -0,0 +1,97 @@
...
+
+// DialContext is the type expected by http.Transport.DialContext.
+func (l *MemoryListener) DialContext(ctx context.Context, network, addr string) (net.Conn, error) {
+	derr := func(err error) error {
+		return &net.OpError{
+			Op:  "dial",
+			Net: memoryAddr{}.Network(),
+			Err: err,
+		}
+	}
```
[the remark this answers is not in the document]
emcfarlane: Removed the lamda.

internal/memhttp/listener.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+type memoryAddr struct{}
+
+// Network implements net.Addr.
+func (memoryAddr) Network() string { return "memory" }
+
+// String implements io.Stringer, returning a value that matches the
+// certificates used by net/http/httptest.
+func (memoryAddr) String() string { return "example.com" }
```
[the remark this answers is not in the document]
emcfarlane: I've used `httptest.DefaultRemoteAddr`.

internal/memhttp/option.go
jhump: >  I think this is valuable normal server usage, removed the test desc.

I don't think that's the right call. For one, it is oddly inconsistent with `http.Server.Shutdown`: it's weird that we _change_ the context that the user supplies. Also, it puts the timeout that the method waits far from the callsite, which can impede readability. If the user wants a shutdown timeout, they add a timeout to the `ctx` right at the callsite.

The only value for it is in tests, where `Shutdown` is called transparently as part of test cleanup. In this case, the user doesn't have control over the callsite, so the duration to wait must be provided up-front when the cleanup action is configured.

emcfarlane: It doesn't modify the original context, just guarantees it always cancels. It's bounded to the limit but can always be reduced by the caller if desired. I thought it was a neat simplification but will revert.

emcfarlane: Wasn't sure how to restrict Cleanup only to the test server without more changes. So exposed Cleanup method on the main server.

jhump: Thank you for making those changes.

Restricting `Cleanup` only to the test server can be easily done by _removing_ it (and the associated field from `memhttp.Server`) and make the timeout _only_ used in the call to `t.Cleanup` in the test package.

The cleanup timeout is really only needed for that one place. If a test wants to explicitly stop the server before that cleanup happens, it can use the server's `Shutdown` and `Close` methods.

emcfarlane: How do you want the test server to get the `CleanupTimeout` on the `*internal.Config`? Options are currently built within `NewServer`.

jhump: Let's punt on it. I can put up a PR later to show what I mean and you and Akshay can see what you think at that time.

internal/memhttp/internal/config.go
```diff
@@ -0,0 +1,35 @@
+// Copyright 2021-2023 The Connect Authors
```
bufdev, line 1: This package would seem a tad overkill to me - is there a reason to not just make the types within here private to `internal/memhttp`?

emcfarlane: @jhump has a plan for restricting the `WithCleanupTimeout` to only the `memhttptest` package: https://github.com/connectrpc/connect-go/pull/594#discussion_r1353458824

jhump: Let's move it back to `memhttp` for now and simplify. I can propose an alternate formulation later and see what folks think of it.

emcfarlane:
```
=== CONT  TestBidiRequiresHTTP2
    connect_ext_test.go:786:
        assertion:      assert.Equal
        got:    unknown
        want:   unimplemented
```
Test is still flaky 😞, need to relook into this.

emcfarlane:
@jhump  behaviour in duplexHTTPCall has changed in the last 2 commits to address the flaky tests. I've removed SetError to avoid clobbering the error response from `makeRequest`. Now `BlockUntilResponseReady` reports the error and `Send` is closed with `io.EOF`. `ensureMakeRequest` is no longer concurrent safe consistent with the interface contract for ClientStreams that state Send or CloseRequest will race. Please re-review!

duplex_http_call.go
```diff
@@ -42,13 +42,11 @@ type duplexHTTPCall struct {
...
 	requestBodyWriter *io.PipeWriter
 
-	sendRequestOnce sync.Once
-	responseReady   chan struct{}
-	request         *http.Request
-	response        *http.Response
-
-	errMu sync.Mutex
-	err   error
+	requestSent   bool
```
emcfarlane, line 45: `sendRequestOnce` doesn't need to lock as `ensureMakeRequest` can't be called concurrently: https://github.com/connectrpc/connect-go/blob/b6d44ecfd51ddab5fa37fbbc7c3cbcece7940179/connect.go#L130-L134

duplex_http_call.go
```diff
@@ -42,13 +42,11 @@ type duplexHTTPCall struct {
...
+	responseReady sync.WaitGroup
```
emcfarlane, line 46: `responseReady` as a `sync.WaitGroup`.

client_ext_test.go
```diff
@@ -106,10 +106,8 @@ func TestClientPeer(t *testing.T) {
 		err = clientStream.Send(&pingv1.SumRequest{})
 		assert.Nil(t, err)
 		// server streaming
-		serverStream, err := client.CountUp(ctx, connect.NewRequest(&pingv1.CountUpRequest{}))
+		serverStream, err := client.CountUp(ctx, connect.NewRequest(&pingv1.CountUpRequest{Number: 1}))
```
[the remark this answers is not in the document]
emcfarlane: It causes the server to error with `invalid_argument: number must be positive: got 0`

duplex_http_call.go
emcfarlane: They are equivalent but `sync.WaitGroup` I thought was more idiomatic and avoids needing to allocate a channel. Can revert.

client_ext_test.go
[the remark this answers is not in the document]
emcfarlane: The test is checking in the interceptor that the peer address and peer protocol is set. It's testing each request style by sending an empty but valid request, however server stream is the only one that will cause the server handler to error. This can now report the server error instead of success. I don't think this is intended behaviour of the test.

duplex_http_call.go
emcfarlane: Reverted, changes were intended to reason about the racy behaviour but it's not a required fix.

connect_ext_test.go
```diff
@@ -1988,13 +1988,14 @@ func TestBidiOverHTTP1(t *testing.T) {
...
 	stream := client.CumSum(context.Background())
+	// Stream creates an async request, can error on Send or Receive.
 	if err := stream.Send(&pingv1.CumSumRequest{Number: 2}); err != nil {
 		assert.ErrorIs(t, err, io.EOF)
 	}
 	_, err := stream.Receive()
 	assert.NotNil(t, err)
 	assert.Equal(t, connect.CodeOf(err), connect.CodeUnknown)
-	assert.Equal(t, err.Error(), "unknown: HTTP status 505 HTTP Version Not Supported")
+	assert.True(t, strings.HasSuffix(err.Error(), "HTTP status 505 HTTP Version Not Supported"))
```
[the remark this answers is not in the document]
emcfarlane: There's no `ErrorContains` so kept the old behaviour.

protocol_connect.go
```diff
@@ -478,20 +478,22 @@ func (cc *connectUnaryClientConn) CloseRequest() error {
...
-	cc.duplexCall.BlockUntilResponseReady()
+	_ = cc.duplexCall.BlockUntilResponseReady()
 	return cc.responseHeader
 }
 
 func (cc *connectUnaryClientConn) ResponseTrailer() http.Header {
-	cc.duplexCall.BlockUntilResponseReady()
+	_ = cc.duplexCall.BlockUntilResponseReady()
 	return cc.responseTrailer
 }
```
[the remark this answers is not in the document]
emcfarlane: Headers are merged into the request/response messages. Presumably to hide header data from the user.

.golangci.yml
```diff
@@ -119,3 +119,9 @@ issues:
     - linters: [revive]
       text: "^if-return: "
       path: error_writer.go
+    # We want to set http.Server's logger
+    - linters: [forbidigo]
+      path: internal/memhttp
```
akshayjshah, line 124: Can we restrict this bypass to just logger-related failures? I'd still like to forbid panic, FIXME, etc. in `memhttp`. I think the right config key is `text:` (we use it in the revive bypass a little further above).

connect_ext_test.go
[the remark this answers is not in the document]
jhump: Weird. There is a `require.ErrorContains`. I wonder why they don't have symmetry in the `assert` package :/

Edit: er, wait. This module is using its own assert package instead of "stretchr/testify". We could always add such a function 🤷

duplex_http_call.go
```diff
@@ -42,13 +42,11 @@ type duplexHTTPCall struct {
...
+	request       *http.Request
+	response      *http.Response
+	responseErr   error
```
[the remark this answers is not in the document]
jhump: Hmm, that isn't quite what I meant. Now you've also moved it away from the `request` and `response` fields. Doesn't `responseReady` also protect those fields? In other words, those fields can be mutated by the goroutine that _sends_ the request and blocks for the reponse, so they can't safely be read unless awaiting `responseReady`, right?

I was expecting something more like so:
```go
	sendRequestOnce sync.Once
	
	// responseReady is closed when the response is ready or when the request
	// fails. Any error on request initialisation will be set on the
	// responseErr. There's always a response if responseErr is nil.
	responseReady chan struct{}
	request       *http.Request
	response      *http.Response
	responseErr   error
```

client_ext_test.go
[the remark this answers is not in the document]
emcfarlane: Not setting `Number` will cause the following only for grpcweb:
```
--- FAIL: TestClientPeer (0.00s)
    --- FAIL: TestClientPeer/grpcweb (0.07s)
        client_ext_test.go:112:
            assertion:  assert.Nil
            got:        unknown: http2: response body closed
```
- `Close` to fail with [`http2: response body closed`](https://github.com/golang/net/blob/48a597731ce9fbe97fed1733da5e808233b6cdcb/http2/transport.go#L2603) [here](https://github.com/connectrpc/connect-go/blob/d9b61e93e4e7e8c2a4c6be7a31fee7c33e7dc9a3/client_ext_test.go#L114)
- Errors on [`discard`](https://github.com/connectrpc/connect-go/blob/23dc2dae9871464ff47521548c5a29f41596ef5b/duplex_http_call.go#L180) which usually receives an `io.EOF`

The grpcweb error response is a header only response, need to investigate why this error is returned on a no body.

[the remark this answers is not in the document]
emcfarlane: Now flaky with: 
```
            client_ext_test.go:113:
                assertion:      assert.Nil
                got:    unknown: io: read/write on closed pipe
```
Discard is reading from a closed response which is causing an ` io.ErrClosedPipe`. This is probably due to the new memhttp. Need to investigate.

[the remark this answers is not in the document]
jhump: > Now flaky with:

By "now", do you mean after the last commit you pushed, which removed calls to `response.Body.Close()`?

Also, in that latest commit, who is the caller that is now responsible for calling `Close`? Is this done when `StreamClientConn.CloseResponse()` is called? Is there a chance some code path is _failing_ to call that? (In particular, I'm suspicious why this would only happen for gRPC-Web and not others.)

> Discard is reading from a closed response which is causing an `io.ErrClosedPipe`.

It seems like `discard` should be resilient against `io.ErrClosedPipe` and replace it with `io.EOF` when observed, no?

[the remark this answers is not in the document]
emcfarlane: It's due to the change in closing the pipe writer on Reader instead of Writer. `net.Pipe` will close both halves but the error is given on the opposite end. So closing the Reader made the Writer return nice `io.EOF` errors but the `Reader` to receive `io.ErrClosedPipe`. HTTP2 libraries set the request error to the response [here](https://github.com/golang/net/blob/4c7a5b64f145becd6fbd7d22f7ad4e9b891ce43a/http2/transport.go#L1326)

I'll need to write more testing around closing, and create an issue for testing with different latencies.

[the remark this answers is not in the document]
emcfarlane: HTTP2 sets the error on the response then [here](https://github.com/golang/net/blob/4c7a5b64f145becd6fbd7d22f7ad4e9b891ce43a/http2/transport.go#L1578)

duplex_http_call.go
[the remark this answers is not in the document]
emcfarlane: Fixed, but moved `request` under `sendRequestOnce` as `responseReady` doesn't block access to `request`.

duplex_http_call.go
```diff
@@ -175,11 +174,13 @@ func (d *duplexHTTPCall) Read(data []byte) (int, error) {
...
 func (d *duplexHTTPCall) CloseRead() error {
-	d.BlockUntilResponseReady()
+	_ = d.BlockUntilResponseReady()
 	if d.response == nil {
 		return nil
 	}
-	if _, err := discard(d.response.Body); err != nil {
+	if _, err := discard(d.response.Body); err != nil &&
+		!errors.Is(err, context.Canceled) &&
+		!errors.Is(err, context.DeadlineExceeded) {
```
emcfarlane, line 183: I've added these error checks to make `CloseResponse` behaviour more consistent. Otherwise Close does race with context cancel on errors causing either nil or a context error. Maybe we can revisit the discard behaviour in another PR.

client_ext_test.go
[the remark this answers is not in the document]
emcfarlane: Discard should never see `io.ErrClosedPipe`, however I've added context error checks to avoid issues with cancelation errors bubbling up to `CloseResponse()` errors.

Working and passing with `-count=100` locally, can't see any more flaky test issues. Could you please re-review.

duplex_http_call.go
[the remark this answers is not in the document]
jhump: >  `responseReady` doesn't block access to `request`

I'm afraid that's not right. `makeRequest`, which is run in a separate goroutine, can mutate `request`: https://github.com/connectrpc/connect-go/blob/main/duplex_http_call.go#L262
So the only way to safely access `request` is to await `responseReady`.

Basically, `request` and `response` are both "owned" by `makeRequest`, and it's not safe to access them until after `responseReady` completes.

connect_ext_test.go
```diff
@@ -183,14 +183,28 @@ func TestServer(t *testing.T) {
...
 		})
+		t.Run("count_up_cancel_after_first_response", func(t *testing.T) {
+			ctx, cancel := context.WithCancel(context.Background())
+			request := connect.NewRequest(&pingv1.CountUpRequest{Number: 5})
+			request.Header().Set(clientHeader, headerValue)
+			stream, err := client.CountUp(ctx, request)
+			assert.Nil(t, err)
+			assert.True(t, stream.Receive())
+			cancel()
+			assert.False(t, stream.Receive())
```
jhump, line 203: This seems racy and could possibly cause timing-sensitive test flakes.

The issue is that `stream.Receive` will try to synchronously read from the response body. However the `cancel` call above is totally async (the request will notice the cancellation and try to then cancel the HTTP operation, but it is effectively happening in the background, asynchronously.

Perhaps better is to call `Receive` in a loop until it fails with a cancelled error, instead of assuming the very next call will fail. If you get EOF after all 5 messages and no cancelled error, then the test case can fail. _Or_ you could insert reasonable delay to make sure background cancellation has had a chance to execute, like `time.Sleep(100*time.Millisecond)`.

duplex_http_call.go
```diff
@@ -257,15 +263,15 @@ func (d *duplexHTTPCall) makeRequest() {
 			err = NewError(CodeUnavailable, err)
 		}
 		d.responseErr = err
-		d.requestBodyReader.CloseWithError(io.EOF)
+		d.requestBodyWriter.Close()
```
jhump, line 266: Why is this changed? Since it was using `CloseWithError` and not simply `Close`, this would not be the source of `io.ErrPipeClosed` issues. In fact, I believe this will cause attempts to _send_ on the broken stream to now return `io.ErrPipeClosed` instead of `io.EOF`.

In order to control the error received by the write half (senders), you must close the read half (which is what it was doing before), and vice versa.

connect_ext_test.go
emcfarlane: The client explicitly looks at context cancellation before calling Receive, behaviour is good to keep that we always do a context check before `Receive()`: https://github.com/connectrpc/connect-go/blob/d9b61e93e4e7e8c2a4c6be7a31fee7c33e7dc9a3/duplex_http_call.go#L166

duplex_http_call.go
emcfarlane: `Close` is just calling `CloseWithError(nil)` which then gets set to `io.EOF`: https://cs.opensource.google/go/go/+/refs/tags/go1.21.3:src/io/pipe.go;l=100
It's always the opposite side of the pipe that gets the error with the current side getting: `io.ErrClosedPipe`.

emcfarlane: `SetError` called close on read: https://github.com/connectrpc/connect-go/blob/d9b61e93e4e7e8c2a4c6be7a31fee7c33e7dc9a3/duplex_http_call.go#L236
Was my mistake switching to `Writer`, it's now fixed in the latest PR.

jhump: > the current side getting: `io.ErrClosedPipe`
> `SetError` called close on read:

Exactly. By closing the writer here, we will induce `io.ErrClosedPipe` in calls to send (which is what used the writer). The old behavior, closing the reader, is better since it results EOF instead.

> Was my mistake switching to Writer, it's now fixed in the latest PR.

I don't follow this. I'm looking at refreshed diff and it's still closing the writer.

emcfarlane: Sorry, I confused myself and switched the ordering. My previous comments don't align.

Closing the write side makes `Write` on the pipe error with `io.ErrClosedPipe`. Connect checks for this when sending: https://github.com/connectrpc/connect-go/blob/d9b61e93e4e7e8c2a4c6be7a31fee7c33e7dc9a3/duplex_http_call.go#L106
However, on the read side we need to return `io.EOF` to avoid HTTP2 setting the `io.ErrClosedPipe` as the error for the response body: https://github.com/connectrpc/connect-go/pull/594#discussion_r1372148665

emcfarlane: `CloseWrite` behaviour was different to `SetError`: https://github.com/connectrpc/connect-go/blob/d9b61e93e4e7e8c2a4c6be7a31fee7c33e7dc9a3/duplex_http_call.go#L133
https://github.com/connectrpc/connect-go/blob/d9b61e93e4e7e8c2a4c6be7a31fee7c33e7dc9a3/duplex_http_call.go#L236

jhump: Got it. Thank you for explaining!

jhump (approved):

akshayjshah (approved):

PR #596, opened 2023-09-26
Cleanup gRPC util funcs for timeout and percent encoding

emcfarlane:
Small cleanup on gRPC util funcs. Removes timeout related init func and improves percent decoder. 

Fix for timeout value length: `grpcEncodeTimeout("45s")` was `45000m` but should be `45000000u`. 8 char limit should only be applied to the value not the value and unit. 

```diff
-BenchmarkGRPCPercentEncoding-8           3277959               347.2 ns/op            64 B/op          2 allocs/op
-BenchmarkGRPCPercentDecoding-8          17811729                65.90 ns/op           32 B/op          1 allocs/op
-BenchmarkGRPCTimeoutEncoding-8          15360220                78.53 ns/op           40 B/op          4 allocs/op
+BenchmarkGRPCPercentEncoding-8          18958806                64.16 ns/op           32 B/op          1 allocs/op
+BenchmarkGRPCPercentDecoding-8          22806225                53.29 ns/op           16 B/op          1 allocs/op
+BenchmarkGRPCTimeoutEncoding-8          29902784                40.86 ns/op           24 B/op          2 allocs/op
```


protocol_grpc.go
```diff
@@ -795,19 +770,48 @@ func grpcParseTimeout(timeout string) (time.Duration, error) {
...
+	if d := div(timeout, time.Second); d <= grpcTimeoutMaxValue {
+		return strconv.FormatInt(d, 10) + "S"
+	}
+	if d := div(timeout, time.Minute); d <= grpcTimeoutMaxValue {
+		return strconv.FormatInt(d, 10) + "M"
+	}
+	return strconv.FormatInt(div(timeout, time.Hour), 10) + "H"
+}
+
+func grpcTimeoutUnitLookup(unit string) time.Duration {
```
mattrobenolt, line 798: As a bit of a nit, we we explicitly use a `byte` here instead of a `string`, the switch statement gets optimized a bit more since the comparison is just with a uint8.

```
$ go test -v main_test.go -bench=. -benchmem
goos: darwin
goarch: arm64
BenchmarkString
BenchmarkString/hit
BenchmarkString/hit-10         	1000000000	         0.7803 ns/op	       0 B/op	       0 allocs/op
BenchmarkString/miss
BenchmarkString/miss-10        	768537562	         1.553 ns/op	       0 B/op	       0 allocs/op
BenchmarkByte
BenchmarkByte/hit
BenchmarkByte/hit-10           	1000000000	         0.3882 ns/op	       0 B/op	       0 allocs/op
BenchmarkByte/miss
BenchmarkByte/miss-10          	1000000000	         0.3111 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	command-line-arguments	3.126s
```

```go
// $ cat main_test.go
package main_test

import (
	"testing"
	"time"
)

var sink time.Duration

func grpcTimeoutUnitLookupString(unit string) time.Duration {
	switch unit {
	case "n":
		return time.Nanosecond
	case "u":
		return time.Microsecond
	case "m":
		return time.Millisecond
	case "S":
		return time.Second
	case "M":
		return time.Minute
	case "H":
		return time.Hour
	default:
		return 0
	}
}

func grpcTimeoutUnitLookupByte(unit byte) time.Duration {
	switch unit {
	case 'n':
		return time.Nanosecond
	case 'u':
		return time.Microsecond
	case 'm':
		return time.Millisecond
	case 'S':
		return time.Second
	case 'M':
		return time.Minute
	case 'H':
		return time.Hour
	default:
		return 0
	}
}

func BenchmarkString(b *testing.B) {
	b.Run("hit", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			sink = grpcTimeoutUnitLookupString("H")
		}
	})
	b.Run("miss", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			sink = grpcTimeoutUnitLookupString("x")
		}
	})
}

func BenchmarkByte(b *testing.B) {
	b.Run("hit", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			sink = grpcTimeoutUnitLookupByte('H')
		}
	})
	b.Run("miss", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			sink = grpcTimeoutUnitLookupByte('x')
		}
	})
}
```

https://godbolt.org/z/eh9arWedv

mattrobenolt:
Side note, in general, I'd be happy to help figure out a workflow to get us setup with [`benchcmp`](https://pkg.go.dev/golang.org/x/tools/cmd/benchcmp) as a bit more of a standardized way to show results here. Big fan of benchcmp.

protocol_grpc.go
emcfarlane: You had me a `byte`. Switched!

emcfarlane:
@mattrobenolt that'd be nice!

protocol_grpc.go
```diff
@@ -795,19 +770,48 @@ func grpcParseTimeout(timeout string) (time.Duration, error) {
 	return time.Duration(num) * unit, nil
 }
 
-func grpcEncodeTimeout(timeout time.Duration) (string, error) {
+func grpcEncodeTimeout(timeout time.Duration) string {
+	div := func(d time.Duration, r time.Duration) int64 {
```
mattrobenolt, line 774: fwiw https://pkg.go.dev/time#Duration.Truncate :)

protocol_grpc.go
```diff
@@ -887,61 +891,76 @@ func grpcStatusFromError(err error) *statusv1.Status {
...
-	for i := offset; i < len(encoded); i++ {
-		c := encoded[i]
-		if c != '%' || i+2 >= len(encoded) {
-			out.WriteByte(c)
-			continue
+	out.Grow(len(input) - 2*percentCount)
+	for i := 0; i < len(input); i++ {
+		switch input[i] {
+		case '%':
+			out.WriteByte(unhex(input[i+1])<<4 | unhex(input[i+2]))
```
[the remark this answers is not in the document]
emcfarlane: I can change the function sig to return a (string, error) and validate it's a hex value. This was lazily ignoring it.

protocol_grpc.go
```diff
@@ -887,61 +891,76 @@ func grpcStatusFromError(err error) *statusv1.Status {
...
+	// We need to escape some characters, so we'll need to allocate a new string.
 	var out strings.Builder
-	out.Grow(2 * len(msg))
-	out.WriteString(msg[:offset])
-	for i := offset; i < len(msg); i++ {
-		c := msg[i]
-		if c < ' ' || c > '~' || c == '%' {
-			fmt.Fprintf(&out, "%%%02X", c)
-			continue
+	out.Grow(len(msg) + 2*hexCount)
```
mattrobenolt, line 905: As another thing to follow up on, it might be opportunistic to leverage a buffer pool here to avoid the allocation. Maybe not worth it though since I'm not sure how often this path gets hit, and if the messages would be roughly the same size or wildly all over the place.

emcfarlane: `strings.Builder` is used to avoid the allocation from a buffer pool to a string.

protocol_grpc.go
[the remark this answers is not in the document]
mattrobenolt: I think that's up to you, I don't entirely know all of the context here or how this might be exploited or do something bad with a NULL byte or things outside of the expected range. I just can see that's possible here if the values input are something odd and not actually hex.

but I think it's probably worth checking that they are hex digits.

Is there any escape sequence where `%%` would be valid? To say, expect a `%` literal?

protocol_grpc.go
mattrobenolt: Oh fair, you'd need to allocate anyways to convert a []byte or something into a string since it outlives the lifetime. :)

protocol_grpc.go
```diff
@@ -795,19 +770,48 @@ func grpcParseTimeout(timeout string) (time.Duration, error) {
...
 	if timeout <= 0 {
-		return "0n", nil
+		return "0n"
 	}
-	for _, pair := range grpcTimeoutUnits {
-		digits := strconv.FormatInt(int64(timeout/pair.size), 10 /* base */)
-		if len(digits) < grpcMaxTimeoutChars {
-			return digits + string(pair.char), nil
-		}
+	if d := div(timeout, time.Nanosecond); d <= grpcTimeoutMaxValue {
```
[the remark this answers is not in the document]
emcfarlane: No diff in speed but it is cleaner.

protocol_grpc.go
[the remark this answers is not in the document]
emcfarlane: '%' will be escaped and represented as hex.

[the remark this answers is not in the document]
emcfarlane: Moved to err sig.

akshayjshah:
> Side note, in general, I'd be happy to help figure out a workflow to get us setup with [benchcmp](https://pkg.go.dev/golang.org/x/tools/cmd/benchcmp) as a bit more of a standardized way to show results here. Big fan of benchcmp.

❤️  That would be amazing, @mattrobenolt. I'm surprised there isn't a halfway decent service for this already :/

protocol_grpc.go
```diff
@@ -887,61 +894,92 @@ func grpcStatusFromError(err error) *statusv1.Status {
...
+	switch {
+	case '0' <= char && char <= '9':
+		return char - '0'
+	case 'a' <= char && char <= 'f':
+		return char - 'a' + 10
+	case 'A' <= char && char <= 'F':
+		return char - 'A' + 10
+	}
+	return 0
+}
```
akshayjshah, line 972: nit: blank line between functions

akshayjshah (commented):
Left some comments, all of which I'm going to fix by pushing some commit's to Ed's branch. The two main issues are that:

- We're leaving a little performance on the table by allocating extra strings when encoding timeouts,
- We've undone a small performance optimization for gRPC handlers, and
- I'd prefer not to remove the error return from unit lookup.

Ed, please take a look at my commits and make sure they're okay with you.

akshayjshah:
Pushed some extra commits to resolve all the review feedback I left. `make bench` now runs all the benchmarks in the root package, and you can use `BENCH=Timeout make bench` to run just a subset of tests.

With these changes, timeout encoding drops to one allocation.

emcfarlane:
All good with me!

Issue #584, opened 2023-09-12
Generic and custom handling for errors that occur before interceptors/endpoints.


anzboi:
**Is your feature request related to a problem? Please describe.**

Our security team flagged a low risk issue that is probably worth addressing anyway. Errors that occur before a request reaches interceptors/endpoint are not masked with a generic error response.

**Describe the solution you'd like**

In general it is good practice to catch any unhandled errors and provide some kind of generic response to clients. Connect currently has no generic error handler for errors that occur before request handling reaches the interceptors or endpoint. In particular, we would like the following

1. Errors that occur before a request reaches the interceptors/endpoint should not be returned, instead replaced with some generic error.
2. The ability to customise handling for errors like this.

**Describe alternatives you've considered**

Since this occurs before control is handed to anything we have control over, the only other option is to implement a custom `http.ResponseWriter` that somehow detects when the error response is created before handling reaches interceptors/endpoints. This is not a trivial task.

**Additional context**

**Example test case**. This produces a response showing some internal service details is being returned to a client (a client might guess the service is written in go). It is also worth noting there is no way to log this error on the server side AFAICT.

- We use the enum field to trigger an error in protojson unmarshal. While we technically have control over that and could mask/log errors, it is annoying and obscure to implement the above feature request using this method.
```proto
service TestService {
    rpc Method(Request) returns (Response) {}
}
message Request {
    MyEnum field1 = 1 [(validate.rules).enum = {defined_only: true}];
}
message Response {}
enum MyEnum {
    MyEnum_UNSPECIFIED = 0;
    MyEnum_VALUE = 1;
}
```

```go
func TestInvokeProtojsonError(t *testing.T) {
	mux := http.NewServeMux()
	mux.Handle(myapiconnect.NewTestServiceHandler(myapiconnect.UnimplementedTestServiceHandler{}))
	svr := &http.Server{Handler: h2c.NewHandler(mux, &http2.Server{})}
	lis, err := net.Listen("tcp", "127.0.0.1:0")
	if err != nil {
		t.Fatal(err)
	}
	go func() { _ = svr.Serve(lis) }()
	t.Cleanup(func() { _ = svr.Close() })

	// Set the enum field to a value outside int32 range
	body := bytes.NewReader([]byte(`{"field1": 1111111111111}`))
	url := "http://" + lis.Addr().String() + myapiconnect.TestServiceMethodProcedure
	contentType := "application/json"

	response, err := http.Post(url, contentType, body)
[...]

PR #599, opened 2023-10-03
Move the indication of destination message type into Codec

jhump:
This is so that a custom `Codec` can choose what details to reveal to clients, in cases where emitting the destination message type is considered to leak too much information.

See #584 for more context.

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -118,7 +118,11 @@ func (c *protoBinaryCodec) Unmarshal(data []byte, message any) error {
 	if !ok {
 		return errNotProto(message)
 	}
-	return proto.Unmarshal(data, protoMessage)
+	err := proto.Unmarshal(data, protoMessage)
+	if err != nil {
+		return fmt.Errorf("unmarshal into %T: %w", message, err)
+	}
+	return nil
 }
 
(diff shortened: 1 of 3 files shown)


akshayjshah (approved):
LGTM! Thanks for the fix, Josh.

akshayjshah:
Actually...should we do the same on the marshaling side?

jhump:
> Actually...should we do the same on the marshaling side?

We don't actually include the type anywhere in marshal errors today. Instead we just return an internal error with "marshal message" as the error context.

But while searching, I did totally miss a spot that needed to be updated 🤦. I had updated only one spot where we unmarshal messages -- in the GET part of the Connect protocol. The other part (that handles all other protocols, unmarshalling from the body) is now also fixed.

For consistency, the code now just adds "unmarshal message" as error context, to decorate whatever the codec returns. It is up to the codec to include any other details (such as the target or source type). Do you think I should _add_ the type to the default codecs' marshal methods? (I assumed not and that your suggestion was just to be sure we were consistent for both halves of the codec.)

akshayjshah:
> Do you think I should add the type to the default codecs' marshal methods?

Nope! This LGTM, let's merge it.

PR #605, opened 2023-10-11
Cast size as int64 to remove overflow checks

emcfarlane:
Three small fixes:
- [`io.ReadFull`](https://pkg.go.dev/io#ReadFull) on prefixes can't return an `io.EOF` if `prefixBytes > 0`
- Cast `size` to `int64` over `int` to avoid the overflow check
- `prefixBytesRead` must equal 5 on success

### DIFF
--- a/envelope.go
+++ b/envelope.go
@@ -229,12 +229,6 @@ func (r *envelopeReader) Read(env *envelope) *Error {
 	prefixBytesRead, err := io.ReadFull(r.reader, prefixes[:])
 
 	switch {
-	case (err == nil || errors.Is(err, io.EOF)) &&
-		prefixBytesRead == 5 &&
-		isSizeZeroPrefix(prefixes):
-		// Successfully read prefix and expect no additional data.
-		env.Flags = prefixes[0]
-		return nil
 	case err != nil && errors.Is(err, io.EOF) && prefixBytesRead == 0:
 		// The stream ended cleanly. That's expected, but we need to propagate them
 		// to the user so that they know that the stream has ended. We shouldn't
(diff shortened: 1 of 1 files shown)


jhump (approved):

envelope.go
```diff
@@ -226,21 +226,13 @@ func (r *envelopeReader) Unmarshal(message any) *Error {
...
-		// to the user so that they know that the stream has ended. We shouldn't
-		// add any alarming text about protocol errors, though.
-		return NewError(CodeUnknown, err)
-	case err != nil:
+	if _, err := io.ReadFull(r.reader, prefixes[:]); err != nil {
+		if errors.Is(err, io.EOF) {
+			// The stream ended cleanly. That's expected, but we need to propagate them
+			// to the user so that they know that the stream has ended. We shouldn't
+			// add any alarming text about protocol errors, though.
+			return NewError(CodeUnknown, err)
```
akshayjshah, line 234: Can we document in this comment that `io.ReadFull` only returns `io.EOF` if it can't read _any_ bytes? That clarifies why this is not a concerning scenario. (I know that's part of the `ReadFull` docs, but it'd be nice to explain it again here.)

envelope.go
```diff
@@ -226,21 +226,13 @@ func (r *envelopeReader) Unmarshal(message any) *Error {
...
-		return nil
-	case err != nil && errors.Is(err, io.EOF) && prefixBytesRead == 0:
-		// The stream ended cleanly. That's expected, but we need to propagate them
-		// to the user so that they know that the stream has ended. We shouldn't
-		// add any alarming text about protocol errors, though.
-		return NewError(CodeUnknown, err)
-	case err != nil:
+	if _, err := io.ReadFull(r.reader, prefixes[:]); err != nil {
+		if errors.Is(err, io.EOF) {
+			// The stream ended cleanly. That's expected, but we need to propagate them
```
akshayjshah, line 231:
```suggestion
			// The stream ended cleanly. That's expected, but we need to propagate an EOF
```

akshayjshah (approved):

akshayjshah (approved):

PR #606, opened 2023-10-11
Enforce discard limits

emcfarlane:
I'm not sure if this is the correct solution but raising as a PR to get feedback. Currently enveloped messages can cause a large discard call as if the size is above the message size, which is above the the discard limit, then we always discard the next N bytes which could be 4Gb (max uint32).

Connect clients will discard when they've read as much input as the max message size, plus the discard limit.

Should the discard size be greater than the message size? For enveloped payloads should we discard the message limit? And for connect payloads should we count the message size read into the discard limit? 

Another option could be to move the discard calls to occur after the error return,  try discard for TCP reuse as a cleanup routine.

### DIFF
--- a/envelope.go
+++ b/envelope.go
@@ -259,11 +259,8 @@ func (r *envelopeReader) Read(env *envelope) *Error {
 		return errorf(CodeInvalidArgument, "message size %d overflowed uint32", size)
 	}
 	if r.readMaxBytes > 0 && size > r.readMaxBytes {
-		_, err := io.CopyN(io.Discard, r.reader, int64(size))
-		if err != nil && !errors.Is(err, io.EOF) {
-			return errorf(CodeUnknown, "read enveloped message: %w", err)
-		}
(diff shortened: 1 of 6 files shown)


envelope.go
```diff
@@ -259,11 +259,8 @@ func (r *envelopeReader) Read(env *envelope) *Error {
 		return errorf(CodeInvalidArgument, "message size %d overflowed uint32", size)
 	}
 	if r.readMaxBytes > 0 && size > r.readMaxBytes {
-		_, err := io.CopyN(io.Discard, r.reader, int64(size))
-		if err != nil && !errors.Is(err, io.EOF) {
-			return errorf(CodeUnknown, "read enveloped message: %w", err)
-		}
-		return errorf(CodeResourceExhausted, "message size %d is larger than configured max %d", size, r.readMaxBytes)
+		src := io.LimitReader(r.reader, int64(size))
```
emcfarlane, line 262: For grpc-web tests I've limited this to the message payload. We discard the message but not the body trailers. Otherwise the count is incorrect.

Issue #609, opened 2023-10-13
Specialize the unary RPC flow internally


jhump:
Currently, all RPCs flow through the same internal abstraction: `duplexHTTPCall`. This abstraction can accommodate all manner of calls, including full-duplex bidirectional calls, but it is overkill for unary RPC. Some of the machinery needed to support bidirectional streaming (synchronization mechanisms and an extra goroutine and `io.Pipe` per call) is pure overhead and unnecessary for simple, unary RPCs.

We can eliminate this overhead by providing a dedicated flow for unary RPCs via a new method on the `protocolClient` interface. Something like so:
```go
// Invoke invokes a unary RPC. The given req is sent to the server and the reply
// is unmarshalled into resp. Returns the HTTP method used, response headers,
// trailers, and optional error.
Invoke(ctx context.Context, spec Spec, reqHdr http.Header, req, resp any) (method string hdr, tlr http.Header, error)
```

A very naive implementation of this new method might look like so (just to demonstrate what it conceptually does):
```go
func (m *myClient) Invoke(ctx context.Context, spec Spec, reqHdr http.Header, req, resp any) (http.Header, http.Header, error) {
	conn := client.protocolClient.NewConn(ctx, spec, reqHdr)
	if err := conn.Send(req); err != nil && !errors.Is(err, io.EOF) {
		_ = conn.CloseRequest()
		_ = conn.CloseResponse()
		return nil, err
	}
	if err := conn.CloseRequest(); err != nil {
		_ = conn.CloseResponse()
		return nil, err
	}
	
	if err := conn.Receive(resp); err != nil {
		return nil, err
	}
	// Make sure there's no extra message and also
	// make sure we've received any trailers.
	if err := conn.Receive(resp); err == nil {
		return nil, NewError(CodeUnknown, errors.New("unary stream has multiple messages"))
	} else if err != nil && !errors.Is(err, io.EOF) {
		return nil, NewError(CodeUnknown, err)
	}
	return http.MethodPost, conn.ResponseHeader(), conn.ResponseTrailer(), conn.CloseResponse() 
}
```

But the value of it is that it enables a much more bespoke flow that eschews the under-the-hood complexity of the `StreamingClientConn` (which is implemented by way of `duplexHTTPConn`).

Having this in place would also make it trivial to resolve #541 for unary RPCs in a way that doesn't further complicate `duplexHTTPConn`.

jhump:
[...]

PR #611, opened 2023-10-19
Special case unary HTTP calls 

emcfarlane:
Special paths the unary request flow to avoid some of the overhead with streaming request bodies. Unary requests now set `http.Request`'s `GetBody` field with the full buffer payload. Buffers are wrapped in a `payloadCloser` to ensure they can be safely re-used. 

See issues #609 and #541 .

## Status

Unary call optimization is applied to all support protocols. 
- [x] connect 
- [x] grpc
- [x] grpc-web
- [x] test `GetBody` is called on go away
- [x] merge #594 
- [ ] implement `io.Reader` and `io.WriterTo` for envelope and avoid buffer copy for client envelopes

Remaining work will be done in a cleanup PR to optimize the envelope payloads.

## Benchmarks
Updated to latest commit. From the benchmarks theres some reduction in allocs for Unary flows, small increase in bytes due to the missing optimization on envelopes.
```
                                │   base.txt   │               new.txt                │
                                │    sec/op    │    sec/op     vs base                │
Connect/connect/unary_big-8       2.013m ± ∞ ¹   1.948m ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/connect/unary_small-8     40.27µ ± ∞ ¹   39.64µ ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/connect/client_stream-8   56.23µ ± ∞ ¹   53.88µ ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/connect/server_stream-8   55.21µ ± ∞ ¹   51.11µ ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/connect/bidi_stream-8     56.27µ ± ∞ ¹   54.70µ ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpc/unary_big-8          1.994m ± ∞ ¹   1.968m ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpc/unary_small-8        44.49µ ± ∞ ¹   41.24µ ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpc/client_stream-8      44.50µ ± ∞ ¹   51.83µ ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpc/server_stream-8      44.10µ ± ∞ ¹   40.93µ ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpc/bidi_stream-8        46.22µ ± ∞ ¹   44.90µ ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpcweb/unary_big-8       2.087m ± ∞ ¹   2.008m ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpcweb/unary_small-8     57.98µ ± ∞ ¹   53.48µ ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpcweb/client_stream-8   57.83µ ± ∞ ¹   56.83µ ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpcweb/server_stream-8   56.03µ ± ∞ ¹   55.27µ ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpcweb/bidi_stream-8     56.62µ ± ∞ ¹   55.56µ ± ∞ ¹       ~ (p=1.000 n=1) ²
geomean                           106.4µ         103.5µ        -2.67%
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05

                                │   base.txt    │                new.txt                │
                                │     B/op      │     B/op       vs base                │
Connect/connect/unary_big-8       5.453Mi ± ∞ ¹   4.945Mi ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/connect/unary_small-8     21.60Ki ± ∞ ¹   21.14Ki ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/connect/client_stream-8   22.65Ki ± ∞ ¹   20.97Ki ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/connect/server_stream-8   22.44Ki ± ∞ ¹   23.98Ki ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/connect/bidi_stream-8     19.90Ki ± ∞ ¹   20.50Ki ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpc/unary_big-8          5.537Mi ± ∞ ¹   4.854Mi ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpc/unary_small-8        25.16Ki ± ∞ ¹   24.26Ki ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpc/client_stream-8      21.57Ki ± ∞ ¹   23.98Ki ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpc/server_stream-8      23.79Ki ± ∞ ¹   23.89Ki ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpc/bidi_stream-8        21.54Ki ± ∞ ¹   21.73Ki ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpcweb/unary_big-8       5.800Mi ± ∞ ¹   5.203Mi ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpcweb/unary_small-8     30.97Ki ± ∞ ¹   32.82Ki ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpcweb/client_stream-8   27.02Ki ± ∞ ¹   29.21Ki ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpcweb/server_stream-8   29.13Ki ± ∞ ¹   32.04Ki ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpcweb/bidi_stream-8     19.33Ki ± ∞ ¹   19.66Ki ± ∞ ¹       ~ (p=1.000 n=1) ²
geomean                           70.59Ki         70.54Ki        -0.07%
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05

                                │  base.txt   │               new.txt               │
                                │  allocs/op  │  allocs/op   vs base                │
Connect/connect/unary_big-8       170.0 ± ∞ ¹   162.0 ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/connect/unary_small-8     152.0 ± ∞ ¹   145.0 ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/connect/client_stream-8   191.0 ± ∞ ¹   188.0 ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/connect/server_stream-8   185.0 ± ∞ ¹   179.0 ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/connect/bidi_stream-8     173.0 ± ∞ ¹   172.0 ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpc/unary_big-8          231.0 ± ∞ ¹   220.0 ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpc/unary_small-8        210.0 ± ∞ ¹   204.0 ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpc/client_stream-8      213.0 ± ∞ ¹   211.0 ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpc/server_stream-8      208.0 ± ∞ ¹   201.0 ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpc/bidi_stream-8        196.0 ± ∞ ¹   195.0 ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpcweb/unary_big-8       218.0 ± ∞ ¹   208.0 ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpcweb/unary_small-8     198.0 ± ∞ ¹   192.0 ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpcweb/client_stream-8   201.0 ± ∞ ¹   199.0 ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpcweb/server_stream-8   195.0 ± ∞ ¹   189.0 ± ∞ ¹       ~ (p=1.000 n=1) ²
Connect/grpcweb/bidi_stream-8     174.0 ± ∞ ¹   173.0 ± ∞ ¹       ~ (p=1.000 n=1) ²
geomean                           193.3         188.2        -2.64%
```


.golangci.yml
```diff
@@ -68,9 +68,9 @@ issues:
     # If future reflect.Kinds are nil-able, we'll find out when a test fails.
     - linters: [exhaustive]
       path: internal/assert/assert.go
-    # We need our duplex HTTP call to have access to the context.
-    - linters: [containedctx]
-      path: duplex_http_call.go
+    # We need our HTTP call to have access to the context and ignore body close.
+    - linters: [containedctx, bodyclose]
+      path: http_call.go
```
jhump, line 73: Why exactly do we now ignore closing the body? Superficially, this doesn't seem right.

Also, I think it would be better to add these exclusions via `nolint` comments, in the actual code, to make it much easier to assess and understand the lint exclusion alongside, by examining the corresponding code.

http_call.go
```diff
@@ -24,6 +25,102 @@ import (
...
+	request := (&http.Request{
+		Method:     http.MethodPost,
+		URL:        url,
+		Header:     header,
+		Proto:      "HTTP/1.1",
+		ProtoMajor: 1,
+		ProtoMinor: 1,
+		Body:       nil, // Set in Do.
+		Host:       url.Host,
+	}).WithContext(ctx)
```
jhump, line 63: We should consolidate this into a helper function instead of having this snippet and associated comment in here 2x.

http_call.go
```diff
@@ -24,6 +25,102 @@ import (
...
+	return &unaryHTTPCall{
+		httpClient: httpClient,
+		request:    request,
+	}
+}
+
+// Do sends the request and waits for the response.
+func (u *unaryHTTPCall) Do(body *bytes.Buffer) *Error {
```
jhump, line 71: Why do we need to split the operation into two phases -- `newUnaryHTTPCall` and then `call.Do`? Having mutators on the call, to support GET method, seems a little silly when the caller could just compute everything first and then construct the call with the correct method, body, and query params.

This split is needed for streaming (mainly client and bidi streams) so that the caller can define headers because actually sending the request. But that isn't the case for unary calls.

http_call.go
```diff
@@ -24,6 +25,102 @@ import (
...
+		err = wrapIfLikelyH2CNotConfiguredError(u.request, err)
+		err = wrapIfLikelyWithGRPCNotUsedError(err)
+		err = wrapIfRSTError(err)
+		if cerr, ok := asError(err); ok {
+			return cerr
+		}
+		return NewError(CodeUnavailable, err)
+	}
+	u.response = response
+	return u.validateResponse(response)
```
jhump, line 104: If the protocol client fails to call `SetValidateResponse`, this will panic. Maybe a nil guard? Or maybe this should just return the response and let the caller validate it instead of this callback approach.

http_call.go
```diff
@@ -267,7 +364,7 @@ func (d *duplexHTTPCall) makeRequest() {
 	}
 	// Once we send a message to the server, they send a message back and
 	// establish the receive side of the stream.
-	response, err := d.httpClient.Do(d.request) //nolint:bodyclose
+	response, err := d.httpClient.Do(d.request)
```
jhump, line 367: I think it's definitely preferable to have this close to the code in question instead of moving them far away into the golang config. I think the golang config is a better place for "cross-cutting" rules -- like disable "forcetypeassert" checks in `*_test.go` files.

jhump (commented):
This didn't actually add a separate entry point to `protocolClient` as previously discussed. That's a bit awkward since it means the `NewConn` still has to special-case unary in ways that cause the shape of the abstractions to be more complicated than needed. For example, `connectUnaryClientConn` seems to be pure bloat. We could get rid of it, and then also improve the `unaryHTTPCall`'s interface if we separate unary from stream calls higher up the stack.

.golangci.yml
emcfarlane: I thought it was desired to use golangci over nolint.

protocol_grpc.go
```diff
@@ -319,6 +315,51 @@ func (g *grpcClient) NewConn(
 	return wrapClientConnWithCodedErrors(conn)
 }
 
+func (g *grpcClient) Invoke(
+	ctx context.Context, spec Spec, request AnyRequest, response AnyResponse,
+) error {
+	// TODO: should we have a grpcUnaryClient struct?
```
emcfarlane, line 321: @jhump how did you see this Invoke implementation working? I started to flesh out what is needed, but feels like it's just copy pasta of the streaming. Shall I factor out all the shareable parts of the streaming call?

http_call.go
emcfarlane: I agree, was following the current pattern. I don't see why it is required for streaming though. Header map is provided on conn creation so  can be mutated without creating the request object. What am I missing?

jhump: > Header map is provided on conn creation

This is not the case for client-stream and bidi endpoints. We create the conn and then return it in the form of a `ClientStreamForClient` or `BidiStreamForClient`. The user can then mutate headers via `stream.RequestHeader()` before sending via `stream.Send()`.

mattrobenolt:
The only thing I wanted to add to this right now is this should in theory be able to be applied just fine for server streaming as well, and I'd suspect some optimizations for client streaming?

The only special case needing the io.Pipe should be for bidi. So if we can swallow up "everything except bidi", that'd be my dream implementation. 🙏 

Only commenting now since I'd hate to be hyper specific on unary and not catch server stream, at minimum.

http_call.go
```diff
@@ -316,3 +373,27 @@ func cloneURL(oldURL *url.URL) *url.URL {
...
+	// NewRequestContext and doesn't effect the actual version
+	// being transmitted.
+	return (&http.Request{
+		Method:     http.MethodPost,
+		URL:        url,
+		Header:     header,
+		Proto:      "HTTP/1.1",
+		ProtoMajor: 1,
+		ProtoMinor: 1,
+		Body:       nil, // Set in Do.
```
mattrobenolt, line 395:
```suggestion
		Body:       body,
```

Is your intent here, since the pipeReader is passed for bidi, and nil for unary.

jhump:
> The only thing I wanted to add to this right now is this should in theory be able to be applied just fine for server streaming as well, and I'd suspect some optimizations for client streaming?

Hmmm, that feels like hyper-specialization. I think it's worth it for unary since they are the overwhelming majority of RPCs, and the flow is straight-forward enough that a specialized flow would maximize readability. But also separately optimizing other streaming types feel like it will either (1) lead to too many permutations (we suddenly have four specializations to maintain instead of two) or (2) require substantial re-architecting of `protocolClient` to make request-handling abstractions independent of response-handling.

> The only special case needing the io.Pipe should be for bidi. So if we can swallow up "everything except bidi", that'd be my dream implementation. 🙏

I don't see how this is true. How does client-streaming work without an `io.Pipe`? The client sends (i.e. writes) request data, but the `*http.Request.Body` is a reader. A pipe is what is used to transform the writer into a reader. The only way I can see client-streaming work without a pipe is to buffer the entire request stream, and that would be pointless as it defeats the benefits of using a client-stream instead of a unary RPC.

mattrobenolt:
I understand I'm not the one implementing this, so I don't fully know the context of everything here, but there are two sides to this. Client request is buffered or streamed, or the server response is buffered or streamed, or the permutation of both streaming.

A simple Unary is buffered request and buffered response, while a server stream is buffered request, streamed response.

Both pure unary and server stream are the same "write request, read response" with Body being a reader. The choice is to either buffer or stream it.

I might be misspeaking on client streaming since that one is a bit more unique and might be limited by stdlib APIs.

But the io.Pipe is only necessary to facilitate reading and writing asynchronously.

imo the "unary" implementation at minimum can easily cover the pure unary and server stream. If it makes sense for APIs available, to lump client streaming with bidi, I think that's fine too, since client streaming is used even less.

We quite heavily utilize server streaming, and there's no reason that needs to be done with the extra goroutine either. The implementaiton is just reading N messages rather than 1.

mattrobenolt:
> Maybe we should take this discussion to the issue (https://github.com/connectrpc/connect-go/issues/609), to better nail down what stakeholders are expecting?

Oh, I missed this! Yes, I agree.

http_call.go
emcfarlane: It is the case for client-stream and bidi. 
- newDuplexHTTP creates the conn and passes in the header: https://github.com/connectrpc/connect-go/blob/b6d44ecfd51ddab5fa37fbbc7c3cbcece7940179/duplex_http_call.go#L59
- `stream.RequestHeader()` points to the map passed into the request object: https://github.com/connectrpc/connect-go/blob/b6d44ecfd51ddab5fa37fbbc7c3cbcece7940179/protocol_connect.go#L471

emcfarlane: On `stream.Send(nil)` we need to create the request object. For `RequestHeader()` we can return the `http.Header` that we would use to create the request object with and not mutate a request, like for unary as you suggested. I still don't see why we have to create the request object up front.

http_call.go
emcfarlane: Thanks for catching!

http_call.go
emcfarlane: Current behaviour is to panic if not provided. I think it would be nice if the `newDuplexCall` took it as an arg, but can also check for the function. Resolving for now.

http_call.go
emcfarlane: Resolving as removed unary http call struct.

duplex_http_call.go
```diff
@@ -15,55 +15,49 @@
...
-	requestBodyReader *io.PipeReader
+	// io.Pipe is used to implement the request body for client streaming calls.
+	// If the request is unary, this is nil.
 	requestBodyWriter *io.PipeWriter
 
-	sendRequestOnce sync.Once
-	responseReady   chan struct{}
-	request         *http.Request
-	response        *http.Response
+	requestSent atomic.Bool
```
emcfarlane, line 42: Switch to `atomic.Bool` to be able to get the state of the `sync.Once` and error on unary if already called.

duplex_http_call.go
```diff
@@ -130,179 +116,196 @@ func (d *duplexHTTPCall) CloseWrite() error {
 	// forever. To make sure users don't have to worry about this, the generated
 	// code for unary, client streaming, and server streaming RPCs must call
 	// CloseWrite automatically rather than requiring the user to do it.
-	return d.requestBodyWriter.Close()
+	if c.requestBodyWriter != nil {
+		return c.requestBodyWriter.Close()
```
emcfarlane, line 120: Ensure we keep the close semantics in https://github.com/connectrpc/connect-go/pull/594

http_call_test.go
```diff
@@ -0,0 +1,220 @@
...
+		if x == 0 {
+			t.Fatal("expected getBody to be called at least once")
+		}
+		atomic.StoreUint32(&getBodyCount, 0)
+	}
+	close(workChan)
+	wg.Wait()
+}
+
+func TestHTTPCallRaces(t *testing.T) {
```
emcfarlane, line 146: This isn't a very nice test but did help recreate the race conditions I was seeing in CI. Any suggestions for improvements?

duplex_http_call.go
```diff
@@ -316,3 +313,81 @@ func cloneURL(oldURL *url.URL) *url.URL {
...
+	p.readN += readN
+	if readN == 0 && p.readN == len(src) {
+		p.completeWithLock()
+		err = io.EOF
+	}
+	return readN, err
+}
+
+// Close implements io.Closer. It signals that the payload has been fully read.
+func (p *payloadCloser) Close() error {
```
emcfarlane, line 356: Close should be called by the `http.Transport` on receiving an `io.EOF` and finished with the request body. But under testing  the HTTP2 transport would sometimes delay the call leading to a deadlock on `Wait()`. We therefore also signal completion by reading all the bytes in the buffer.

duplex_http_call.go
```diff
@@ -301,3 +326,81 @@ func cloneURL(oldURL *url.URL) *url.URL {
...
+	}
+	return true
+}
+
+// Wait blocks until the payload has been fully read.
+func (p *payloadCloser) Wait() {
+	p.wait.Wait()
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	p.buf = nil // Discard the buffer
```
jhump, line 397: I think this will defeat the `GetBody` improvement. It seems likely the HTTP framework could rain the entire body (particularly for small bodies) and/or call `Close` and still need to be able to rewind the buffer and send again.

duplex_http_call.go
```diff
@@ -131,160 +120,196 @@ func (d *duplexHTTPCall) CloseWrite() error {
...
+		c.request.ContentLength = int64(buffer.Len())
+		c.request.GetBody = func() (io.ReadCloser, error) {
+			if payload.Rewind() {
+				return payload, nil
+			}
+			return nil, errors.New("payload cannot be rewound")
+		}
+		// Wait for the payload to be fully drained before we return
+		// from Send to ensure the buffer can safely be reused.
+		defer payload.Wait()
```
jhump, line 237: This seems excessively dangerous to have unbounded blocking here. Is it really 100% certain that the body will _always_ be drained or closed in a timely manner, in the face of all possible errors?

And, even if so, is it worth paying potential latency hit to wait for this? I would think instead we'd release the buffer (i.e. return to pool) asynchronously, instead of holding of the train here.

jhump (commented):
I'm trying to grok the formulation in #622 as it relates to the changes here. I take it that you'd change the `httpCall.Send(*bytes.Buffer)` method to instead be `httpCall.Send(*envelope)`, avoid having to create the unified buffer (which means copying the prefix and the serialized message to yet-another-buffer). Is that how you see these two fitting together?

duplex_http_call.go
emcfarlane: `Close` has to be called as any reader would leak otherwise. This keeps the current behaviour between `sendUnary` and `sendStream`: both don't modify the buffer and both return and the buffer is safe to mutate afterwards.

duplex_http_call.go
emcfarlane: `GetBody` shouldn't be called after we've received a response. Draining or calling `Close` will allow `Send` to return and the buffer to be recycled but `GetBody` can call `Rewind()` to reset as many times as needed (5x I think max). The number of `Close` calls or drains isn't counted.

emcfarlane:
Superseded by #649

PR #617, opened 2023-10-26
Add flags to running golangci-lint

oliversun9:
Add `--modules-download-mode=readonly` and `--timeout=3m0s` to running `golangci-lint`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ install: ## Install all binaries
 .PHONY: lint
 lint: $(BIN)/golangci-lint $(BIN)/buf ## Lint Go and protobuf
 	go vet ./...
-	golangci-lint run
+	golangci-lint run --modules-download-mode=readonly --timeout=3m0s
 	buf lint
 	buf format -d --exit-code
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -52,7 +52,7 @@ lint: $(BIN)/golangci-lint $(BIN)/buf ## Lint Go and protobuf
 
 .PHONY: lintfix
 lintfix: $(BIN)/golangci-lint $(BIN)/buf ## Automatically fix some lint errors
-	golangci-lint run --fix
+	golangci-lint run --fix --modules-download-mode=readonly --timeout=3m0s
```
jhump, line 55: @oliversun9, is this just for consistency with other repos somewhere, or was this an issue in CI?

Also, I suspect you meant to add these flags to the invocation up above, on line 49, instead of here.

akshayjshah: Consistency with other repos. And yes, agree that these flags should probably go on both this invocation and the one above.

oliversun9: > I suspect you meant to add these flags to the invocation up above, on line 49, instead of here.

Great catch! Yes, I meant to add it to `lint`. Now it's added to both invocations.

akshayjshah (approved):

PR #622, opened 2023-11-02
Add Read and WriterTo methods to envelopes

emcfarlane:
Allow `src` and `dst` to be set for envelope read/writers. Envelopes also implement `io.Reader` and `io.WriterTo` methods to allow passing envelopes as a message payload.

Requirement for #611 to avoid copying the buffers. See #609

### DIFF
--- a/envelope.go
+++ b/envelope.go
@@ -42,14 +42,52 @@ var errSpecialEnvelope = errorf(
 type envelope struct {
 	Data  *bytes.Buffer
 	Flags uint8
+
+	offset int
 }
 
 func (e *envelope) IsSet(flag uint8) bool {
 	return e.Flags&flag == flag
 }
 
+// Read implements io.Reader.
+func (e *envelope) Read(data []byte) (readN int, err error) {
+	if e.offset < 5 {
+		prefix := makeEnvelopePrefix(e.Flags, e.Data.Len())
+		readN = copy(data, prefix[e.offset:])
+		e.offset += readN
+		if e.offset < 5 {
+			return readN, nil
+		}
+		data = data[readN:]
+	}
+	n := copy(data, e.Data.Bytes()[e.offset-5:])
+	e.offset += n
+	readN += n
+	if readN == 0 && e.offset == e.Data.Len()+5 {
+		err = io.EOF
+	}
+	return readN, err
+}
+
+// WriteTo implements io.WriterTo.
+func (e *envelope) WriteTo(dst io.Writer) (wroteN int64, err error) {
+	if e.offset < 5 {
+		prefix := makeEnvelopePrefix(e.Flags, e.Data.Len())
+		prefixN, err := dst.Write(prefix[e.offset:])
+		e.offset += prefixN
+		wroteN += int64(prefixN)
+		if e.offset < 5 {
+			return wroteN, err
+		}
+	}
+	n, err := dst.Write(e.Data.Bytes()[e.offset-5:])
+	e.offset += n
+	wroteN += int64(n)
+	return wroteN, err
+}
+
 type envelopeWriter struct {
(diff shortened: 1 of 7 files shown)


envelope.go
```diff
@@ -42,69 +42,105 @@ var errSpecialEnvelope = errorf(
...
-	if env.IsSet(flagEnvelopeCompressed) ||
-		w.compressionPool == nil ||
-		env.Data.Len() < w.compressMinBytes {
-		if w.sendMaxBytes > 0 && env.Data.Len() > w.sendMaxBytes {
-			return errorf(CodeResourceExhausted, "message size %d exceeds sendMaxBytes %d", env.Data.Len(), w.sendMaxBytes)
+func (w *envelopeWriter) Write(dst io.Writer, env *envelope) *Error {
+	if !env.IsSet(flagEnvelopeCompressed) &&
+		w.compressionPool != nil &&
+		env.Data.Len() > w.compressMinBytes {
+		if err := w.compress(env.Data); err != nil {
```
emcfarlane, line 122: Behaviour change on `Write`, now mutates the `*env` to compress, if needed. Nothing currently depended on the `*env` being unchanged.

envelope.go
```diff
@@ -42,69 +42,105 @@ var errSpecialEnvelope = errorf(
...
+		wroteN += int64(prefixN)
+		if e.offset < 5 {
+			return wroteN, err
+		}
+	}
+	n, err := dst.Write(e.Data.Bytes()[e.offset-5:])
+	e.offset += n
+	wroteN += int64(n)
+	return wroteN, err
+}
```
emcfarlane, line 88: For #611 will also need to add `func (e *envelope) Rewind() bool { b.offset = 0}`

jhump: I think it would be way cleaner if `envelope` were immutable once sent. So that means we'd instead want a `NewReader() io.Reader` method, so _that_ thing can be mutable/rewindable. That way, if any synchronization becomes necessary (as exists in the payload thingie in #611), it can live in that reader impl and not leak into the envelope struct itself.

jhump (commented):
I don't see why we're changing from currying the writer in the envelopeWriter to not. That seems unrelated and unnecessary.

I know you stated in the description that you don't see why it's "necessary", but I'm afraid I don't see why it's necessary to change. This seems like aesthetic choice (IMO, low value) that expands the blast radius of this PR by several files and dozens of call sites. If it's not needed for the change, let's please leave it out. In fact, it seems like you'll end up having to undo/redo it anyway to actually change `duplexHTTPCall` so it has a `send(*envelope)` method instead of implementing `io.Writer`.

In fact, I think it's actually worth making that change (changing `duplexHTTPCall` to actually use what's in this branch) _in this PR_, so we can see what that looks like even before the unary-request specialization change.

emcfarlane:
> currying the writer in the envelopeWriter to not. That seems unrelated and unnecessary.

I don't understand what you are asking for. How can I call a different method on `duplexHTTPCall` and still wrap it as an `io.Writer`?

> I know you stated in the description that you don't see why it's "necessary", but I'm afraid I don't see why it's necessary to change.

I didn't, but they aren't. Why do we have a different call sites for unary vs envelope? The only difference should be the prefix which we can cover by wrapping in an envelope payload.  I'll make a proposal for the work and post it in the issue.

jhump:
> How can I call a different method on duplexHTTPCall and still wrap it as an io.Writer?

You _can't_, hence me suggesting that the changes here will need to be undone/re-done to weave a different type through.

> Why do we have a different call sites for unary vs envelope?

Sorry, I guess I need to go look at a little more of the code to get adequate context for this.

emcfarlane:
Closing, will discuss the internal API changes needed in #611

Issue #626, opened 2023-11-06
Debug flaky `TestStreamForServer` test


emcfarlane:
**Describe the bug**

On [bump golang.org/x/net from 0.16.0 to 0.17.0](https://github.com/connectrpc/connect-go/actions/runs/6751140307/job/18354726039) tests can fail on `Send` reporting a closed request body error:
```
--- FAIL: TestStreamForServer (0.25s)
    --- FAIL: TestStreamForServer/get-spec (0.37s)
        connect_ext_test.go:1562: 
            assertion:	assert.Nil
            got:	unknown: EOF
```

**To Reproduce**

```bash
go test -run=TestStreamForServer -count=1000
```

**Additional context**

Tracing down the calls:
- [`Write`](https://github.com/connectrpc/connect-go/blob/2be1555cbf36879b6b1197cd09bc6e4cbe10290f/duplex_http_call.go#L107) returns an `io.EOF` error from a close on the request body (`io.ErrClosedPipe`)
- [`Close`](https://github.com/golang/net/blob/b225e7ca6dde1ef5a5ae5ce922861bda011cfabd/http2/transport.go#L449) is called by the transport
- [`abortStream`](https://github.com/golang/net/blob/b225e7ca6dde1ef5a5ae5ce922861bda011cfabd/http2/transport.go#L2934) from a process reset stream frame with the error:
```
stream error: stream ID 3; NO_ERROR; received from peer:
```

Unrelated to changes in x/net 0.17.0. Occurs in the latest v1.12 release.

jhump:
Hmm. Do you know _why_ the stream is being aborted? That error message ("NO_ERROR received from peer") suggests the server sent a GOAWAY frame with error code NO_ERROR, which is supposed to mean that the server is gracefully shutting down. Is there a possible race with an RPC operation completing the server being shutdown prematurely? (Like maybe there's a `defer` function that should instead be a `t.Cleanup` function somewhere?)

jhump:
> The PR goes into the details.

Okay, moving this convo to the PR.

emcfarlane:
Would be good to discuss the intended error behaviour here. I think that `Send(nil)` should always succeed when initializing the request headers. From the `get-spec` test above, the behaviour that `Send(nil)` could return `io.EOF` is surprising to me and I'd expect shouldn't happen before calling `CloseRequest`:
[...]

PR #627, opened 2023-11-06
Debug flaky `TestStreamForServer` test

emcfarlane:
`TestStreamForServer` can fail on `Send(nil)` returning `io.EOF`. This occurs because `Send(nil)` will try to `Write(nil)` to the request body pipe. In almost all cases this will be read (and therefore succeed) when buffering the request body. However, on the server returning quickly, the request can be closed causing the `Send` to fail with `io.EOF` error. 

To ensure a `Send` succeeds we need to match it with a correlating `Receive` in the server. Nil writes will never be guaranteed unless the server tries to `Receive` a message, which would read the request body and therefore enforce the successful drain. This is a problem for `Send(nil)` as the intended behaviour is to ensure the request headers are sent, which does complete but returns with an unexpected error.

To solve for `Send(nil)` we now check if this write caused the request to send and avoid triggering a body write if so. Any other cases and later if `Send(nil)` is called we `Write(nil)` which keeps the same behaviour.

Tests `client-stream-conn` and `client-stream-send-msg` write a message `Send(msg)`. For these I've modified the test cases to `Receive()` the message to ensure that the write completes and doesn't otherwise affect the test case.

Fixes https://github.com/connectrpc/connect-go/issues/626

### DIFF
--- a/connect_ext_test.go
+++ b/connect_ext_test.go
@@ -1511,6 +1511,7 @@ func TestBidiStreamServerSendsFirstMessage(t *testing.T) {
 func TestStreamForServer(t *testing.T) {
 	t.Parallel()
 	newPingClient := func(t *testing.T, pingServer pingv1connect.PingServiceHandler) pingv1connect.PingServiceClient {
+		t.Helper()
 		mux := http.NewServeMux()
 		mux.Handle(pingv1connect.NewPingServiceHandler(pingServer))
 		server := memhttptest.NewServer(t, mux)
--- a/connect_ext_test.go
+++ b/connect_ext_test.go
@@ -1510,7 +1510,7 @@ func TestBidiStreamServerSendsFirstMessage(t *testing.T) {
 
 func TestStreamForServer(t *testing.T) {
 	t.Parallel()
-	newPingClient := func(pingServer pingv1connect.PingServiceHandler) pingv1connect.PingServiceClient {
+	newPingClient := func(t *testing.T, pingServer pingv1connect.PingServiceHandler) pingv1connect.PingServiceClient {
 		mux := http.NewServeMux()
 		mux.Handle(pingv1connect.NewPingServiceHandler(pingServer))
 		server := memhttptest.NewServer(t, mux)
@@ -1522,7 +1522,7 @@ func TestStreamForServer(t *testing.T) {
 	}
 	t.Run("not-proto-message", func(t *testing.T) {
 		t.Parallel()
-		client := newPingClient(&pluggablePingServer{
+		client := newPingClient(t, &pluggablePingServer{
 			cumSum: func(ctx context.Context, stream *connect.BidiStream[pingv1.CumSumRequest, pingv1.CumSumResponse]) error {
 				return stream.Conn().Send("foobar")
 			},
@@ -1536,7 +1536,7 @@ func TestStreamForServer(t *testing.T) {
 	})
 	t.Run("nil-message", func(t *testing.T) {
 		t.Parallel()
-		client := newPingClient(&pluggablePingServer{
+		client := newPingClient(t, &pluggablePingServer{
 			cumSum: func(ctx context.Context, stream *connect.BidiStream[pingv1.CumSumRequest, pingv1.CumSumResponse]) error {
 				return stream.Send(nil)
 			},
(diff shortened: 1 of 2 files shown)


jhump:
> `TestStreamForServer` can close the stream before reading the request body.

But _why_. Is this intentional?

If not, should we better synchronize the closing of the stream so it happens after the body has been read? If it _is_ intentional, seems like we should instead just relax the test assertions to allow `io.EOF` from the call to `Send`.

> To write a headers only request `Send` is called with zero length payload.

This is not accurate. It is called with `nil`. Any other zero-length payload actually sends a zero-length request message.

A remark on the linked issue:
> `Send(nil)` is ignored if we get a close.

I don't understand the value of this. It seems to be solely for our test assertions, for which a superior fix is to relax the assertion. IMO, it seems perfectly reasonable that `Send(nil)` could return an error if it's expected that the connection might be closing/closed. Especially since this formulation _only_ means "write headers only" if no messages have been sent; otherwise it is a no-op. So it seems odd to me that `Send(nonNilMsg); Send(nil); Send(nonNilMessage)` could conceivably result in a non-nil error, then a nil error(??). then a non-nil error.

emcfarlane:
I'm not sure `Send(nil)` should be calling `Write(nil)` on the request body. The send ensures the request is created with client headers attached. I'd think the behaviour we would want from `Send(nil)` is to return nil if successfully creating the request and sending the headers. What should the behaviour be if multiple `Send(nil)` are called? 

> superior fix is to relax the assertion

Depends on the behaviour we expect from `Send(nil)`. I thought returning an `io.EOF` would be unexpected.

jhump:
> Depends on the behaviour we expect from `Send(nil)`. I thought returning an `io.EOF` would be unexpected.

`Send` is specified to return `io.EOF` if the stream has been closed, so the current behavior seems fine.

> What should the behaviour be if multiple `Send(nil)` are called?

I think it's actually fine as is. Is there actually a problem with the current behavior? It calling code were organized in a way that multiple, independent layers of logic all called `Send(nil)`, to ensure the request was started, are you suggesting that should that be an error? I don't think we need to special-case `nil` beyond the current implementation.

IMO, the bug in the flaky test is because it could realistically expect to receive an error but the assertion was too strict.

BTW, I'm still curious: **why** can `TestStreamForServer` close the stream before reading the request body. Is that an intentional aspect of the test?

emcfarlane:
> why can TestStreamForServer close the stream before reading the request body

https://github.com/connectrpc/connect-go/blob/20b5723aa3c09d35c0ce1b3b120796b5382388fd/connect_ext_test.go#L1525-L1532

The `Send(nil)` on line 1531 is never matched by a receive in the server handler. I think this almost always passes now because of buffering reads. But to ensure this nil write succeeds and can't be close by the handler exit I think you'd want:

```go
func(ctx context.Context, stream *connect.BidiStream[pingv1.CumSumRequest, pingv1.CumSumResponse]) error {
	go func() { _, err := stream.Receive() }() // forces Read, get's io.EOF on return
	return stream.Conn().Send("foobar")
}
```

This would force the request body to read the nil message, ensuring `Send(nil)` get a nil error. And because there's no `Send(msg)` the server `Receive()` will get an `io.EOF` when the handler returns.

duplex_http_call.go
```diff
@@ -229,10 +234,12 @@ func (d *duplexHTTPCall) BlockUntilResponseReady() error {
 // ensureRequestMade sends the request headers and starts the response stream.
 // It is not safe to call this concurrently. Write and CloseWrite call this but
 // ensure that they're not called concurrently.
-func (d *duplexHTTPCall) ensureRequestMade() {
+func (d *duplexHTTPCall) ensureRequestMade() (isFirst bool) {
 	d.sendRequestOnce.Do(func() {
+		isFirst = true
```
jhump, line 239: Eek, this is kinda weird, relying on only one closure over the variable actually mutating it.

I know I suggested _not_ changing the `Once` to an `atomic.Bool` in a past PR, but that was because in that PR it was being changed unnecessarily and felt like "noise" in the diff.. But here, it's actually the perfect change to make since it would make this construct more clear.

connect_ext_test.go
```diff
@@ -1510,7 +1510,8 @@ func TestBidiStreamServerSendsFirstMessage(t *testing.T) {
 
 func TestStreamForServer(t *testing.T) {
 	t.Parallel()
-	newPingClient := func(pingServer pingv1connect.PingServiceHandler) pingv1connect.PingServiceClient {
+	newPingClient := func(t *testing.T, pingServer pingv1connect.PingServiceHandler) pingv1connect.PingServiceClient {
+		t.Helper()
```
jhump, line 1514: I think you can undo all of the changes in this file, right? Looks like the addition of `t` here is unused, and this appears to be the only change in here.

emcfarlane: The servers close are matched to the clients which is nice. The other change in this file is the two `Receives` for the test cases that `Send` a non nil message.

duplex_http_call.go
emcfarlane: Switched to atomic.

connect_ext_test.go
emcfarlane: As in `server := memhttptest.NewServer(t, mux)` takes the correct subtests `*testing.T` rather than the parent.

jhump: Ah, okay. That was a bit subtle due to the shadowing of the name `t`.

jhump (approved):

Issue #571, opened 2023-08-23
Expose deprecated option in interceptor


makasim:
**Is your feature request related to a problem? Please describe.**
It would allow us to implement an interceptor that logs the usage of deprecated endpoints.

**Describe the solution you'd like**

RPC specs provide a `deprecated` option that could be used on a method. Currently, Connect only added a deprecated comment. Exposing the value to interceptors could help to build more robust solutions.  

```
service CreateService {
  rpc Create(CreateRequest) returns (CreateResponse) {
    option deprecated = true;
  }
}
```

The solution I see: The option could be added to `connect.Spec`. That would be enough for this particular case. 

```
WrapUnary(unaryFunc connect.UnaryFunc) connect.UnaryFunc {
	return func(ctx context.Context, request connect.AnyRequest) (connect.AnyResponse, error) {
	    if request.Spec().Deprecated {
	       log.Printf("WARNING: %s is deprecated", request.Spec().Procedure)
	    }
            return unaryFunc(ctx, request)
	}
}
```

A more generic approach is to expose all the options in `connect.Spec` somehow (map?). In this case, we could add a custom option with a api removal date and set some alerts for it. 

**Describe alternatives you've considered**
-

**Additional context**
-

akshayjshah:
Hey @makasim! Great suggestion - I've been meaning to expose the RPC's descriptor on `Spec`, which would give you access to `deprecated`, any custom options, and a bunch of other info.

In the meantime, you can get the descriptor yourself from the global registry: see [this comment on another issue](https://github.com/connectrpc/connect-go/issues/404#issuecomment-1328271332). `protoc-gen-connect-go` now generates constants for service names, so you shouldn't need a hard-coded string to look up the descriptor.

makasim:
Looks like a global registry would solve my issue. Thank you for the hint!

Issue #600, opened 2023-10-03
setting and reading custom options


bumberboy:
Hello there

Thanks for building this!

I'm looking to set some custom options on my rpcs:

```
import "google/protobuf/descriptor.proto";

extend google.protobuf.MethodOptions {
  optional string abc = 51234;
}

service XXXService {
  rpc Xxx(XxReq) returns (XxResp) {
    option (abc) = "hello, world";
  }

}

```

And be able to read them within an interceptor. The intention is to build an authorization interceptor that checks the custom option on the rpc on request to run the right ACL auth validation. Any idea how I can achieve this?

Thanks in advance!

emcfarlane:
As an example it might be good to have a look at: https://github.com/akshayjshah/connectauth

Can use context to propagate auth values.

bumberboy:
@emcfarlane Thanks for the suggestion! That looks perfect for authentication (and I've already implemented something similar to that.

I'm looking to implement authorization now, where I try to answer the question: can this user perform this action (READ, WRITE, DELETE) on this resource?

I'm intending to annotate each rpc that requires authorization with a `resource` and an `action`, and then within my authorization interceptor, make a call to an external service that determines if the user has the right permissions. I'm looking to use spicedb from authzed for the auth service.

jeffmccune:
I'm working on a similar use case Ed asked me to provide feedback on.  I'd like to include primitive roles (owner, editor, viewer) to start my project and check "has rbac" off the list, then potentially add fine grained permissions later if customers need it.

I've got an ID token with a groups claim, which is a json array of strings.  Groups in this specific case is really a role and may be a different claim depending on the id provider.  Keycloak for example has a role claim.

It would buy a lot of runway early in the project if I could tag each service method as allowing messages from one of owners, editors, or viewers.

One specific use case I've been thinking of:  Given a PatchUser(UserRequest) UserResponse method, allow only requests with user.email == idToken.email so that users cannot set their email to arbitrary values.

jhump:
This should be addressed by #629.

PR #629, opened 2023-11-06
Add Schema field to Spec for introspection

emcfarlane:
New field Schema of type `any` on [`Spec`](https://github.com/connectrpc/connect-go/pull/629/files#diff-82f6a0eb10c47bc6590f594adc997b5a12de370dfd735a05c44cf8ff6fbf7a12R318). For proto based schemas the type will be of `protoreflect.MethodDescriptor`. This allows for easy introspection to interceptors.

Updated generated code to add the new `WithSchema` option to both handlers and clients. Schema is fetched from the proto file using the [`file.GoDescriptorIdent`](https://pkg.go.dev/google.golang.org/protobuf/compiler/protogen#File). To ensure generated code is matched with a library that includes the option a new version tag is added for v1.13.

### Example

For a unary request one can access the input and output descriptors by casting the interface and using protoreflect:
```go
methodDesc, ok := req.Spec().Schema.(protoreflect.MethodDescriptor)
if !ok {
	return nil, fmt.Errorf("expected protoreflect.MethodDescriptor got %T", req.Spec().Schema)
}
inputDesc := methodDesc.Input()
outputDesc := methodDesc.Output()
```

Fixes https://github.com/connectrpc/connect-go/issues/571 and https://github.com/connectrpc/connect-go/issues/600

### DIFF
--- a/client_ext_test.go
+++ b/client_ext_test.go
@@ -196,71 +196,34 @@ func TestSpecSchema(t *testing.T) {
 		connect.WithInterceptors(&assertSchemaInterceptor{t}),
 	))
 	server := memhttptest.NewServer(t, mux)
-	testcases := []struct {
-		name string
-		opts []connect.ClientOption
-	}{{
-		name: connect.ProtocolConnect,
-	}, {
-		name: connect.ProtocolGRPC,
-		opts: []connect.ClientOption{
-			connect.WithGRPC(),
-		},
-	}, {
-		name: connect.ProtocolGRPCWeb,
-		opts: []connect.ClientOption{
-			connect.WithGRPC(),
-		},
-	}}
-	for _, testcase := range testcases {
-		testcase := testcase
-		t.Run(testcase.name, func(t *testing.T) {
-			ctx := context.Background()
-			client := pingv1connect.NewPingServiceClient(
-				server.Client(),
-				server.URL(),
-				connect.WithClientOptions(testcase.opts...),
-				connect.WithInterceptors(&assertSchemaInterceptor{t}),
-			)
-			t.Parallel()
-			t.Run("unary", func(t *testing.T) {
-				unaryReq := connect.NewRequest[pingv1.PingRequest](nil)
-				_, err := client.Ping(ctx, unaryReq)
-				assert.Nil(t, err)
-				text := strings.Repeat(".", 256)
-				r, err := client.Ping(ctx, connect.NewRequest(&pingv1.PingRequest{Text: text}))
-				assert.Nil(t, err)
-				assert.Equal(t, r.Msg.Text, text)
-			})
-			t.Run("client_stream", func(t *testing.T) {
-				clientStream := client.Sum(ctx)
-				t.Cleanup(func() {
-					_, closeErr := clientStream.CloseAndReceive()
-					assert.Nil(t, closeErr)
-				})
-				assert.NotZero(t, clientStream.Spec().Schema)
-				err := clientStream.Send(&pingv1.SumRequest{})
-				assert.Nil(t, err)
-			})
-			t.Run("server_stream", func(t *testing.T) {
-				serverStream, err := client.CountUp(ctx, connect.NewRequest(&pingv1.CountUpRequest{}))
-				t.Cleanup(func() {
-					assert.Nil(t, serverStream.Close())
-				})
-				assert.Nil(t, err)
-			})
-			t.Run("bidi_stream", func(t *testing.T) {
-				bidiStream := client.CumSum(ctx)
-				t.Cleanup(func() {
-					assert.Nil(t, bidiStream.CloseRequest())
-					assert.Nil(t, bidiStream.CloseResponse())
-				})
-				assert.NotZero(t, bidiStream.Spec().Schema)
-				err := bidiStream.Send(&pingv1.CumSumRequest{})
-				assert.Nil(t, err)
-			})
+	ctx := context.Background()
+	client := pingv1connect.NewPingServiceClient(
+		server.Client(),
+		server.URL(),
+		connect.WithInterceptors(&assertSchemaInterceptor{t}),
+	)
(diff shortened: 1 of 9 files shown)


cmd/protoc-gen-connect-go/main.go
```diff
@@ -283,17 +268,19 @@ func generateClientImplementation(g *protogen.GeneratedFile, service *protogen.S
 		)
 		g.P("httpClient,")
 		g.P(`baseURL + `, procedureConstName(method), `,`)
+		g.P(connectPackage.Ident("WithSchema"), "(",
+			g.QualifiedGoIdent(file.GoDescriptorIdent),
```
emcfarlane, line 272: Should I break out the service as a variable?

client_ext_test.go
```diff
@@ -186,6 +188,81 @@ func TestGetNotModified(t *testing.T) {
...
 
+func TestSpecSchema(t *testing.T) {
+	t.Parallel()
+	mux := http.NewServeMux()
+	mux.Handle(pingv1connect.NewPingServiceHandler(
+		pingServer{},
+		connect.WithInterceptors(&assertSchemaInterceptor{t}),
+	))
+	server := memhttptest.NewServer(t, mux)
+	testcases := []struct {
```
emcfarlane, line 199: I test over all protocols, but the method should be invariant to it. Shall I reduce to only one and a unary/stream test?

emcfarlane: I've removed the extra testcases.

cmd/protoc-gen-connect-go/main.go
emcfarlane: Have moved to a `serviceDescriptor` within the func. 

Adding a private global var for each method that resolves the method descriptor would move the panic to the init. Shall I try it out? Panicing on the `NewClient` or `NewHandler` call over global vars. Would there be any use cases for someone modifying the public file descriptor var and then calling `NewClient` and expecting the types to include the modified descriptors?

jhump: > Adding a private global var for each method that resolves the method descriptor would move the panic to the init. Shall I try it out?

Yeah, I think that would be better.

> Would there be any use cases for someone modifying the public file descriptor var and then calling NewClient and expecting the types to include the modified descriptors?

Eek, that sounds like a seriously misbehaving program, and I personally don't think we should bother trying to support that case.

emcfarlane: Switched to global var descriptors.

client_ext_test.go
```diff
@@ -233,3 +273,47 @@ func (a *assertPeerInterceptor) WrapStreamingHandler(next connect.StreamingHandl
...
+
+func (a *assertSchemaInterceptor) WrapUnary(next connect.UnaryFunc) connect.UnaryFunc {
+	return func(ctx context.Context, req connect.AnyRequest) (connect.AnyResponse, error) {
+		if !assert.NotNil(a.tb, req.Spec().Schema) {
+			return next(ctx, req)
+		}
+		methodDesc, ok := req.Spec().Schema.(protoreflect.MethodDescriptor)
+		assert.True(a.tb, ok)
+		procedure := fmt.Sprintf("/%s/%s", methodDesc.Parent().FullName(), methodDesc.Name())
+		assert.Equal(a.tb, procedure, req.Spec().Procedure)
```
jhump, line 289: If the assertion (line 287) fails, then the next line will cause a nil dereference panic when it tries to call `methodDesc.Parent()`.
```suggestion
		if assert.True(a.tb, ok) {
			procedure := fmt.Sprintf("/%s/%s", methodDesc.Parent().FullName(), methodDesc.Name())
			assert.Equal(a.tb, procedure, req.Spec().Procedure)
		}
```

cmd/protoc-gen-connect-go/main.go
```diff
@@ -225,6 +211,24 @@ func generateServiceNameConstants(g *protogen.GeneratedFile, services []*protoge
 	g.P()
 }
 
+func generateServiceNameVariables(g *protogen.GeneratedFile, file *protogen.File) {
+	wrapComments(g, "These variables are the protoreflect.Descriptor objects for the ", file.Desc.Name(),
+		" service's methods.")
```
jhump, line 216: This comment doesn't seem too helpful. It's a little confusing because `file.Desc.Name()` returns the last element of the package.

Did you mean `file.Desc.Path()`? That would actually be awkward, too. Also, grammatically this probably wants to say "services and methods" instead of "service's methods" because there could be multiple services and "services' methods" is probably a little confusing to read.

jhump (approved):

cmd/protoc-gen-connect-go/main.go
emcfarlane: Thanks for catching, simplified to:
```
// These variables are the protoreflect.Descriptor objects for the RPCs defined in this package.
```

cmd/protoc-gen-connect-go/main.go
```diff
@@ -212,8 +212,7 @@ func generateServiceNameConstants(g *protogen.GeneratedFile, services []*protoge
 }
 
 func generateServiceNameVariables(g *protogen.GeneratedFile, file *protogen.File) {
-	wrapComments(g, "These variables are the protoreflect.Descriptor objects for the ", file.Desc.Name(),
-		" service's methods.")
+	wrapComments(g, "These variables are the protoreflect.Descriptor objects for the RPCs defined in this package.")
```
jhump, line 215: nit: just the RPCs defined in this file. A package can have multiple protos with separate services.
```suggestion
	wrapComments(g, "These variables are the protoreflect.Descriptor objects for the RPCs defined in this file.")
```

jhump: Oh, never mind. I see you've just mirrored the comments for the service and procedure names.

PR #630, opened 2023-11-07
Support dynamic messages with Initializer options

### LINKED ISSUE #523: Support for dynamic or reflected types for client
_Disclaimer: It may be possible that connect does support this and my lack of understanding makes me unable to make it work. However I couldn't find anything in GitHub issues or in docs (please note, that there is a similar issue #312, but author decided to go with other approach and there is no answer about dynamic types)._

How one can perform connect request with dynamic types, like when a request is created using google.golang.org/protobuf/types/dynamicpb package and there is no concrete Go type available (the service, method and type information is inferred using reflection or read from a file containing the file descriptor set)?

As an example we can talk about tools like grpcurl - how to perform a task similar to one done by such a tool, but using connect protocol?

I imagine something like UntypedClient (in contrast to currently available Client that uses generic) or a special handling of dynamic messages inside of the current implementation.

It feels like it should be possible to call a service using connect starting with just a "google.golang.org/protobuf/reflect/protoreflect".MethodDescriptor and without reimplementing of large parts of connect-go package.

I have found something about (and experimented with) creating custom codec, but it feels like it should be possible to just use connect with dynamic message, without special hacks, especially that dynamicpb is part of protobuf in Go.

I have also attached in the following messages results of my experimentation - a workaround if someone needs to achieve this with connect.

**Additional context**
If I'm wrong and there is other issue or it is mentioned in the documentation, then I'm sorry that I have bothered you. If there is already a support for such things in connect, then please provide explanation in this issue, to allow others (like me ;) ) to find it.

emcfarlane:
To support dynamic clients and servers we need to initialize messages on `Receive`. This adds two new options `WithRequestInitializer` and `WithResponseInitializer` that provide `InitializerFunc`s that help construct request messages before use. Proto based schemas can use the new Schema field on Spec to introspect the method types. Other IDLs can set their own schema using the `WithSchema` option. 

### Dynamic Message Initializer

As an example to support [`dynamicpb.Message`](https://pkg.go.dev/google.golang.org/protobuf/types/dynamicpb#Message) we will inspect the schema and set the message descriptor. An intializer func can be run on the client or handller. This func is provided to the examples below as an option:
```go
// dynamicMessageInitialzier inits a dynamicpb.Message descriptor to the correct
// input or output type for the method.
func dynamicMessageInitializer(spec Spec, msg any) error {
	dynamic, ok := msg.(*dynamicpb.Message)
	if !ok {
		return nil
	}
	desc, ok := spec.Schema.(protoreflect.MethodDescriptor)
	if !ok {
		return fmt.Errorf("unexpected schema type %T for %T message", spec.Schema, dynamic)
	}
	// Client initializer is run on response, Server on request.
	if spec.IsClient {
		*dynamic = *dynamicpb.NewMessage(desc.Output())
	} else {
		*dynamic = *dynamicpb.NewMessage(desc.Input())
	}
	return nil
}
```

### Dynamic Clients

Construct a client using `dynamicpb.Message`. Ensure the URL includes the suffix of the handler method to invoke and to include `WithSchema` and `WithResponseInitializer` options:
```go
desc, _ = protoregistry.GlobalFiles.FindDescriptorByName("connect.ping.v1.PingService.Ping")
methodDesc := desc.(protoreflect.MethodDescriptor)
// Create a client with [dynamicpb.Message] as the request and response message types.
client := connect.NewClient[dynamicpb.Message, dynamicpb.Message](
	http.DefaultClient,
	serverURL + "/connect.ping.v1.PingService/Ping",
	connect.WithSchema(methodDesc),
	connect.WithResponseInitializer(dynamicMessageInitializer),
	connect.WithIdempotency(connect.IdempotencyNoSideEffects),
)
// Build the message using the input descriptor.
msg := dynamicpb.NewMessage(methodDesc.Input())
msg.Set(
	methodDesc.Input().Fields().ByName("number"),
	protoreflect.ValueOfInt64(42),
)
res, _ := client.CallUnary(context.Background(), connect.NewRequest(msg))
// Get values from the message using the output descriptor.
got := res.Msg.Get(
	methodDesc.Output().Fields().ByName("number")
).Int()
```

### Dynamic Handlers

Create the server func with a signature that includes the `dynamicpb.Message`. Then construct the handler specifying the procedure URL and ensure to include `WithSchema` and `WithRequestInitializer` options:
```go
// Create a handler func with [dynamicpb.Message] as the request and response message types.
dynamicPing := func(_ context.Context, req *connect.Request[dynamicpb.Message]) (*connect.Response[dynamicpb.Message], error) {
	got := req.Msg.Get(methodDesc.Input().Fields().ByName("number")).Int()
	msg := dynamicpb.NewMessage(methodDesc.Output())
	msg.Set(
		methodDesc.Output().Fields().ByName("number"),
		protoreflect.ValueOfInt64(got),
	)
	return connect.NewResponse(msg), nil
}
mux := http.NewServeMux()
mux.Handle("/connect.ping.v1.PingService/Ping",
	connect.NewUnaryHandler(
		"/connect.ping.v1.PingService/Ping",
		dynamicPing,
		connect.WithSchema(methodDesc),
		connect.WithRequestInitializer(dynamicMessageInitializer),
		connect.WithIdempotency(connect.IdempotencyNoSideEffects),
	),
)
```

Fixes #523

### DIFF
--- a/client.go
+++ b/client.go
@@ -189,6 +189,7 @@ type clientConfig struct {
 	URL                    *url.URL
 	Protocol               protocol
 	Procedure              string
+	Schema                 any
 	CompressMinBytes       int
 	Interceptor            Interceptor
 	CompressionPools       map[string]*compressionPool
@@ -251,6 +252,7 @@ func (c *clientConfig) newSpec(t StreamType) Spec {
 	return Spec{
 		StreamType:       t,
 		Procedure:        c.Procedure,
+		Schema:           c.Schema,
 		IsClient:         true,
 		IdempotencyLevel: c.IdempotencyLevel,
 	}
--- a/client.go
+++ b/client.go
@@ -227,7 +227,6 @@ func newClientConfig(rawURL string, options []ClientOption) (*clientConfig, *Err
 		Procedure:        protoPath,
 		CompressionPools: make(map[string]*compressionPool),
 		BufferPool:       newBufferPool(),
-		Initializer:      defaultInitializer,
 	}
 	withProtoBinaryCodec().applyToClient(&config)
 	withGzip().applyToClient(&config)
--- a/client.go
+++ b/client.go
@@ -92,7 +92,7 @@ func NewClient[Req, Res any](httpClient HTTPClient, url string, options ...Clien
 			_ = conn.CloseResponse()
 			return nil, err
 		}
-		response, err := receiveUnaryResponse[Res](conn)
+		response, err := receiveUnaryResponse[Res](conn, config)
 		if err != nil {
 			_ = conn.CloseResponse()
 			return nil, err
@@ -135,7 +135,10 @@ func (c *Client[Req, Res]) CallClientStream(ctx context.Context) *ClientStreamFo
 	if c.err != nil {
 		return &ClientStreamForClient[Req, Res]{err: c.err}
 	}
-	return &ClientStreamForClient[Req, Res]{conn: c.newConn(ctx, StreamTypeClient, nil)}
+	return &ClientStreamForClient[Req, Res]{
+		conn:   c.newConn(ctx, StreamTypeClient, nil),
+		config: c.config,
+	}
 }
 
 // CallServerStream calls a server streaming procedure.
@@ -160,7 +163,10 @@ func (c *Client[Req, Res]) CallServerStream(ctx context.Context, request *Reques
 	if err := conn.CloseRequest(); err != nil {
 		return nil, err
 	}
-	return &ServerStreamForClient[Res]{conn: conn}, nil
+	return &ServerStreamForClient[Res]{
+		conn:   conn,
+		config: c.config,
+	}, nil
 }
 
 // CallBidiStream calls a bidirectional streaming procedure.
@@ -168,7 +174,10 @@ func (c *Client[Req, Res]) CallBidiStream(ctx context.Context) *BidiStreamForCli
 	if c.err != nil {
 		return &BidiStreamForClient[Req, Res]{err: c.err}
 	}
-	return &BidiStreamForClient[Req, Res]{conn: c.newConn(ctx, StreamTypeBidi, nil)}
+	return &BidiStreamForClient[Req, Res]{
+		conn:   c.newConn(ctx, StreamTypeBidi, nil),
+		config: c.config,
+	}
 }
 
 func (c *Client[Req, Res]) newConn(ctx context.Context, streamType StreamType, onRequestSend func(r *http.Request)) StreamingClientConn {
@@ -190,6 +199,7 @@ type clientConfig struct {
 	Protocol               protocol
 	Procedure              string
 	Schema                 any
+	Initializer            InitializerFunc
 	CompressMinBytes       int
 	Interceptor            Interceptor
 	CompressionPools       map[string]*compressionPool
@@ -217,6 +227,7 @@ func newClientConfig(rawURL string, options []ClientOption) (*clientConfig, *Err
 		Procedure:        protoPath,
 		CompressionPools: make(map[string]*compressionPool),
 		BufferPool:       newBufferPool(),
+		Initializer:      defaultInitializer,
 	}
 	withProtoBinaryCodec().applyToClient(&config)
 	withGzip().applyToClient(&config)
--- a/connect.go
+++ b/connect.go
@@ -38,9 +38,10 @@ const Version = "1.13.0-dev"
 // These constants are used in compile-time handshakes with connect's generated
 // code.
 const (
-	IsAtLeastVersion0_0_1 = true
-	IsAtLeastVersion0_1_0 = true
-	IsAtLeastVersion1_7_0 = true
+	IsAtLeastVersion0_0_1  = true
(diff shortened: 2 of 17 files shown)


client.go
```diff
@@ -135,7 +135,10 @@ func (c *Client[Req, Res]) CallClientStream(ctx context.Context) *ClientStreamFo
 	if c.err != nil {
 		return &ClientStreamForClient[Req, Res]{err: c.err}
 	}
-	return &ClientStreamForClient[Req, Res]{conn: c.newConn(ctx, StreamTypeClient, nil)}
+	return &ClientStreamForClient[Req, Res]{
+		conn:   c.newConn(ctx, StreamTypeClient, nil),
+		config: c.config,
```
emcfarlane, line 140: I've passed around a config object. Should this be limited to `Initializer`?

protocol.go
```diff
@@ -394,3 +397,24 @@ func canonicalizeContentTypeSlow(contentType string) string {
 	}
 	return mime.FormatMediaType(base, params)
 }
+
+// defaultInitializer is the default initializer for dynamic messages. It
+// initializes the message to the type specified in the Spec.
+func defaultInitializer(spec Spec, msg any) error {
```
emcfarlane, line 403: This is the current default initializer that support `dynamicpb.Message`. Do we want this? Or only explicitly providing an initializer? Having this as the default will always interface cast the type.

connect.go
```diff
@@ -358,15 +358,18 @@ type handlerConnCloser interface {
...
+		return nil, err
+	}
 	if err := conn.Receive(&msg); err != nil {
 		return nil, err
 	}
 	// In a well-formed stream, the response message may be followed by a block
 	// of in-stream trailers or HTTP trailers. To ensure that we receive the
 	// trailers, try to read another message from the stream.
-	if err := conn.Receive(new(T)); err == nil {
+	if err := conn.Receive(nil); err == nil {
```
emcfarlane, line 372: Made it nil to avoid calling initializer twice, should never be used to unmarshal into.

option.go
```diff
@@ -194,6 +194,22 @@ func WithSchema(schema any) Option {
...
+type InitializerFunc func(spec Spec, msg any) error
+
+// WithInitializer provides a function that initializes a message. It may be
+// used to dynamically construct messages.
+//
+// By default, an initializer is provided to support [dynamicpb.Message]
+// messages. This initializer sets the descriptor from the Schema field of the
+// [Spec]. The Schema must be of type [protoreflect.MethodDescriptor] to use
+// dynamicpb.Message.
+func WithInitializer(initializer InitializerFunc) Option {
```
emcfarlane, line 209: Splitting into two might make it clearer, wanted to try the single version but will switch to:
```go
func WithResponseInitializer(initializer InitializerFunc) ClientOption
func WithRequestInitializer(initializer InitializerFunc) HandlerOption
```

emcfarlane: Split into two.

protocol.go
jhump: We decided to provide this as an option as an alternative to adding transparent internal support for `dynamicpb`. So I don't think we need to do both. The default can be nil, which simply means no initialization needed.

jhump (commented):
FYI, I've merged the schema PR, so this needs a rebase

connect.go
emcfarlane: Could also pass the message again?

jhump: If this succeeds (like if there _is_ another message on the stream), this seems like a source of likely nil-dereference panics. Please don't change this to nil.

I think it's safest as it was and just call the initializer 2x. Maybe just add a TODO about optimizing away the allocation and second initialization (like an alternate method on the streaming conn that could be used to test if there is more data on the stream, or push this check for a second message into the streaming conn, so it can done without allocating and providing a destination message).

client.go
akshayjshah: I'd say yes. Config structs are nice on the outer edge of the program, but they have a _lot_ of random stuff on them that shouldn't be hobo-ing its way through every struct.

option.go
```diff
@@ -194,6 +194,27 @@ func WithSchema(schema any) Option {
...
 }
 
+// InitializerFunc is a function that initializes a message. It may be used to
+// dynamically construct messages. It is called on client and handler receives
+// to construct the message to be unmarshaled into.
+//
+// The message will be a non nil pointer to the type created by the client or
+// handler. Use the Schema field of the [Spec] to determine the type of the
+// message.
+type InitializerFunc func(spec Spec, message any) error
```
akshayjshah, line 204: Does this exported type carry its weight? There aren't any methods on it - why not inline it into the two options below?

akshayjshah (commented):
Haven't looked at the implementation in detail, but the exported API makes sense to me.

emcfarlane:
Closing, reopening as a fork PR: https://github.com/connectrpc/connect-go/pull/640

option.go
emcfarlane: This is address in #640

client.go
emcfarlane: See #640

PR #631, opened 2023-11-09
Update CI deps

emcfarlane:
Updates golanglint-ci to v1.55.2 and buf to v1.27.2. Fix lint issues from protogetter. Disabled [inamedparam](https://pkg.go.dev/github.com/macabu/inamedparam#section-readme)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -58,6 +58,7 @@ linters:
     - varcheck          # abandoned
     - wrapcheck         # don't _always_ need to wrap errors
     - wsl               # generous whitespace violates house style
+    - inamedparam       # convention is not followed
 issues:
   exclude:
     # Don't ban use of fmt.Errorf to create new errors, but the remaining
--- a/Makefile
+++ b/Makefile
@@ -80,15 +80,15 @@ $(BIN)/protoc-gen-connect-go:
 
 $(BIN)/buf: Makefile
 	@mkdir -p $(@D)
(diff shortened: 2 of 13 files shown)


.golangci.yml
```diff
@@ -58,6 +58,7 @@ linters:
     - varcheck          # abandoned
     - wrapcheck         # don't _always_ need to wrap errors
     - wsl               # generous whitespace violates house style
+    - inamedparam       # convention is not followed
```
jhump, line 61: FWIW, when recently upgrading another repo, I ignored the new `protogetter` linter, too 🤷 
https://github.com/bufbuild/protocompile/commit/089712432bdc5d54241e66771e884c0e6fbb754f#diff-6179837f7df53a6f05c522b6b7bb566d484d5465d9894fb04910dd08bb40dcc9R80

Admittedly, there were several places where I needed a `Has<Field>` method which does not exist on generated messages. The only way to test for this is direct field access, to test against nil, and that also triggers the linter :(

jhump (approved):

.golangci.yml
emcfarlane: Shall I disable it here too? Let me know and I can resolve either way.

akshayjshah: I think disabling it here is the right call. Also, please keep the list of disabled linters alphabetized - it's much easier to keep in sync across repositories that way.

emcfarlane: Disabled.

PR #636, opened 2023-11-15
Type URLs can have any host and even include URI scheme

jhump:
The previous implementation _assumed_ that the `type_url` field in an `Any` would _always_ have a `type.googleapis.com/` prefix. Though that is typical in practice, it is not actually required per the spec for that field. In fact, `protoc` allows the use of `Any` messages (inside message literals for complex option values) that contain a prefix of `type.googleprod.com/` (therefore, so does the `buf` CLI).

The way all protobuf runtimes handle `Any` is to instead look at the last path element and ignore the prefix (which is also mentioned in the spec for the `type_url` field). So this makes the code more resilient to unexpected values for type URLs. Previously, the `ErrorDetail.Type()` method would return the full URL, instead of the type name, in the event that the URL had an unexpected prefix.

### DIFF
--- a/error.go
+++ b/error.go
@@ -74,11 +74,11 @@ func (d *ErrorDetail) Type() string {
 	// than plain type names, but there aren't any descriptor registries
 	// deployed. With the current state of the `Any` code, it's not possible to
 	// build a useful type registry either. To hide this from users, we should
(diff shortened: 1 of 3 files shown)


akshayjshah (approved):
Good catch!

PR #640, opened 2023-11-20
Add support for dynamic message types

### LINKED ISSUE #523: Support for dynamic or reflected types for client

emcfarlane:
Adds support for dynamic clients and servers with options to initialize messages on receives. See #523 for more details. Two new options `WithRequestInitializer` and `WithResponseInitializer` provide initializer functions that can construct messages before unmarshalling. Proto based schemas can use the new `Schema` field on `Spec` to introspect the method types. Other IDLs can set their own schema using the `WithSchema` option. 

### Dynamic Message Initializer

As an example to support [`dynamicpb.Message`](https://pkg.go.dev/google.golang.org/protobuf/types/dynamicpb#Message) we will inspect the schema and set the message descriptor. An intializer func can be run on the client or handller. This func is provided to the examples below as an option:
```go
// dynamicMessageInitialzier inits a dynamicpb.Message descriptor to the correct
// input or output type for the method.
func dynamicMessageInitializer(spec Spec, msg any) error {
	dynamic, ok := msg.(*dynamicpb.Message)
	if !ok {
		return nil
	}
	desc, ok := spec.Schema.(protoreflect.MethodDescriptor)
	if !ok {
		return fmt.Errorf("unexpected schema type %T for %T message", spec.Schema, dynamic)
	}
	// Client initializer is run on response, Server on request.
	if spec.IsClient {
		*dynamic = *dynamicpb.NewMessage(desc.Output())
	} else {
		*dynamic = *dynamicpb.NewMessage(desc.Input())
	}
	return nil
}
```

### Dynamic Clients

Construct a client using `dynamicpb.Message`. Ensure the URL includes the suffix of the handler method to invoke and to include `WithSchema` and `WithResponseInitializer` options:
```go
desc, _ = protoregistry.GlobalFiles.FindDescriptorByName("connect.ping.v1.PingService.Ping")
methodDesc := desc.(protoreflect.MethodDescriptor)
// Create a client with [dynamicpb.Message] as the request and response message types.
client := connect.NewClient[dynamicpb.Message, dynamicpb.Message](
	http.DefaultClient,
	serverURL + "/connect.ping.v1.PingService/Ping",
	connect.WithSchema(methodDesc),
	connect.WithResponseInitializer(dynamicMessageInitializer),
	connect.WithIdempotency(connect.IdempotencyNoSideEffects),
)
// Build the message using the input descriptor.
msg := dynamicpb.NewMessage(methodDesc.Input())
msg.Set(
	methodDesc.Input().Fields().ByName("number"),
	protoreflect.ValueOfInt64(42),
)
res, _ := client.CallUnary(context.Background(), connect.NewRequest(msg))
// Get values from the message using the output descriptor.
got := res.Msg.Get(
	methodDesc.Output().Fields().ByName("number")
).Int()
```

### Dynamic Handlers

Create the server func with a signature that includes the `dynamicpb.Message`. Then construct the handler specifying the procedure URL and ensure to include `WithSchema` and `WithRequestInitializer` options:
```go
// Create a handler func with [dynamicpb.Message] as the request and response message types.
dynamicPing := func(_ context.Context, req *connect.Request[dynamicpb.Message]) (*connect.Response[dynamicpb.Message], error) {
	got := req.Msg.Get(methodDesc.Input().Fields().ByName("number")).Int()
	msg := dynamicpb.NewMessage(methodDesc.Output())
	msg.Set(
		methodDesc.Output().Fields().ByName("number"),
		protoreflect.ValueOfInt64(got),
	)
	return connect.NewResponse(msg), nil
}
mux := http.NewServeMux()
mux.Handle("/connect.ping.v1.PingService/Ping",
	connect.NewUnaryHandler(
		"/connect.ping.v1.PingService/Ping",
		dynamicPing,
		connect.WithSchema(methodDesc),
		connect.WithRequestInitializer(dynamicMessageInitializer),
		connect.WithIdempotency(connect.IdempotencyNoSideEffects),
	),
)
```

Fixes #523

### DIFF
--- a/client_stream.go
+++ b/client_stream.go
@@ -25,8 +25,8 @@ import (
 // It's returned from [Client].CallClientStream, but doesn't currently have an
 // exported constructor function.
 type ClientStreamForClient[Req, Res any] struct {
-	conn   StreamingClientConn
-	config *clientConfig
+	conn        StreamingClientConn
+	initializer func(Spec, any) error
 	// Error from client construction. If non-nil, return for all calls.
 	err error
 }
@@ -79,7 +79,7 @@ func (c *ClientStreamForClient[Req, Res]) CloseAndReceive() (*Response[Res], err
 		_ = c.conn.CloseResponse()
 		return nil, err
 	}
-	response, err := receiveUnaryResponse[Res](c.conn, c.config)
+	response, err := receiveUnaryResponse[Res](c.conn, c.initializer)
 	if err != nil {
 		_ = c.conn.CloseResponse()
 		return nil, err
@@ -98,9 +98,9 @@ func (c *ClientStreamForClient[Req, Res]) Conn() (StreamingClientConn, error) {
 // It's returned from [Client].CallServerStream, but doesn't currently have an
 // exported constructor function.
 type ServerStreamForClient[Res any] struct {
-	conn   StreamingClientConn
-	config *clientConfig
-	msg    *Res
+	conn        StreamingClientConn
+	initializer func(Spec, any) error
+	msg         *Res
 	// Error from client construction. If non-nil, return for all calls.
 	constructErr error
 	// Error from conn.Receive().
@@ -117,8 +117,8 @@ func (s *ServerStreamForClient[Res]) Receive() bool {
 		return false
 	}
 	s.msg = new(Res)
-	if s.config.Initializer != nil {
-		if err := s.config.Initializer(s.conn.Spec(), s.msg); err != nil {
+	if s.initializer != nil {
+		if err := s.initializer(s.conn.Spec(), s.msg); err != nil {
 			s.receiveErr = err
 			return false
 		}
@@ -183,8 +183,8 @@ func (s *ServerStreamForClient[Res]) Conn() (StreamingClientConn, error) {
 // It's returned from [Client].CallBidiStream, but doesn't currently have an
 // exported constructor function.
 type BidiStreamForClient[Req, Res any] struct {
-	conn   StreamingClientConn
-	config *clientConfig
+	conn        StreamingClientConn
+	initializer func(Spec, any) error
 	// Error from client construction. If non-nil, return for all calls.
 	err error
 }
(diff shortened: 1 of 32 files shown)


option.go
```diff
@@ -350,6 +368,22 @@ func (o *schemaOption) applyToHandler(config *handlerConfig) {
...
+	Initializer func(spec Spec, message any) error
+}
+
+func (o *requestInitializerOption) applyToHandler(config *handlerConfig) {
+	config.Initializer = o.Initializer
+}
+
+type responseInitializerOption struct {
+	Initializer func(spec Spec, message any) error
+}
```
jhump, line 381: I see the value of two strongly-typed option functions. But the value of two differen implementation types is much less. I think you could collapse these two into a single `initializerOption` type.

client_stream.go
```diff
@@ -115,6 +117,12 @@ func (s *ServerStreamForClient[Res]) Receive() bool {
 		return false
 	}
 	s.msg = new(Res)
+	if s.initializer != nil {
+		if err := s.initializer(s.conn.Spec(), s.msg); err != nil {
+			s.receiveErr = err
+			return false
+		}
+	}
```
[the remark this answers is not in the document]
emcfarlane: Instead of having the function branch off the two configs I've added a `maybeInitializer` that wraps the func and does the nil check. Wdyt?

option.go
emcfarlane: Unified.

PR #642, opened 2023-11-28
Run conformance tests in CI

jhump:
This uses the newest updates in the connectrpc/conformance repo to run conformance tests as part of CI, with every PR and every commit to main.

Since the conformance repo's reference client and reference server were written using connect-go, we are simply leveraging that implementation in the conformance repo. And we use a `go.mod` file with a replace directive to effectively build the conformance tests and the reference client and server using the current HEAD of this repo.


internal/conformance/runconformance.sh
```diff
@@ -0,0 +1,21 @@
...
+mkdir -p $BINDIR
+GO="${GO:-go}"
+
+# These will get built using current HEAD of this connect-go repo
+# thanks to replace directive in go.mod. So by testing the reference
+# implementations (which are written with connect-go), we are effectively
+# testing changes in this repo.
+"$GO" build -o $BINDIR/connectconformance connectrpc.com/conformance/cmd/connectconformance
+"$GO" build -o $BINDIR/referenceclient connectrpc.com/conformance/cmd/referenceclient
+"$GO" build -o $BINDIR/referenceserver connectrpc.com/conformance/cmd/referenceserver
```
akshayjshah, line 15: Does `"$GO"` need to be quoted here? I don't think so, since we don't care about word splitting.

internal/conformance/runconformance.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+set -e
+cd "$(dirname $0)"
```
akshayjshah, line 3: I think we need double quotes around `$0`, don't we?

internal/conformance/runconformance.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+set -e
```
akshayjshah, line 2: Any objections to the full `set -euo pipefail`? Erroring on undefined variables seems nice, and some extra safety around pipes will be nice if we ever add one later on.

Shell scripts aren't my favorite, so I prefer maximum strictness. If you haven't already, please also run this through `shellcheck`.

akshayjshah (changes requested):

internal/conformance/config.yaml
```diff
@@ -0,0 +1,21 @@
...
+    - HTTP_VERSION_2
+    - HTTP_VERSION_3
+  protocols:
+    - PROTOCOL_CONNECT
+    - PROTOCOL_GRPC
+    - PROTOCOL_GRPC_WEB
+  codecs:
+    - CODEC_PROTO
+    - CODEC_JSON
+    - CODEC_TEXT
```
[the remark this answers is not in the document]
akshayjshah: Nothing to do here, just an idle comment on the not-great content-type we end up with. It's nice that we can test this in the conformance suite, though!

internal/conformance/runconformance.sh
jhump: 🤷, if we need quotes around `"$0"` (which I agree that I should add), then we probably want it here, too, as a just in case and for the same reason. Like if someone did something like `GO="/mnt/c/Program Files/cygwin/bin/go"`.

I prefer to quote anything that is user-provided since we don't for sure whether it will have whitespace, and whitespace is totally allowed in file paths, so seemed fair to do this way.

Or were you thinking that someone might want to use whitespace to do a `go` command with add'l args, like `GO="go-wrapper --some arg -v"`?

jhump: Went ahead and made the requested change, assuming you were thinking of a go command with args.

internal/conformance/runconformance.sh
jhump: done

internal/conformance/runconformance.sh
jhump: Sure, that's fine. I am fully aware of that "strict mode" setting, but am not in a habit of using it because I often use `grep` in scripts, which is a serious pain to use with `set -o pipefail` since it exits non-zero if there are no matches... (Writing `(grep ... || true) | xargs ..` is pretty gross.)

Done.

internal/conformance/go.mod
```diff
@@ -0,0 +1,45 @@
+module connectrpc.com/connect/internal/conformance
+
+go 1.19
+
+require connectrpc.com/conformance v1.0.0-rc1
```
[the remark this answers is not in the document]
jhump: I've added a stanza to `dependabot.yaml`. Hopefully it's right and hopefully it won't do anything annoying with the indirect dependencies here.

internal/conformance/config.yaml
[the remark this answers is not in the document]
jhump: Yep, it's true. I mainly added to the conformance tests because it is a codec that is supported by default by the grpc-web JS stuff (since they can't support JSON because they're using a protobuf runtime w/out JSON support, that is the codec to use for human-readable payloads).

Similarly, all the compression algorithms listed are ones that are actually referenced in the Connect wire protocol spec (br, zstd) or the gRPC HTTP/2 protocol spec (deflate, snappy). So if someone implements them, it seemed reasonable to be able to check them.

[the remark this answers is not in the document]
jhump: No, the text encoding is not the same as the base64-encoding of binary stuff -- grpc-web-text does not produce human readable messages in the network inspector.

I can't seem to find the support for the text codec anymore. Maybe it was removed. At FullStory, we investigated using grpc-web circa 2017/2018. That was before there was an official one, so we were looking at the TS stuff in [improbable-eng/grpc-web](https://github.com/improbable-eng/grpc-web). We experimented with using the text codec. (I remember having to configure our gRPC servers to support it, since it is not a supported codec out-of-the-box in grpc-go. I also remember being very confused and frustrated that there was no support in protobuf-js for JSON.)

So it may not be present anymore, in either the official implementation or the improbable-eng one. But I had added it at as a supported codec in the conformance tests from my memory of that. We can always remove it if we think no one would ever use it.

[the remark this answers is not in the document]
timostamm: I can see how the text format would have looked like a good option back then. Doesn't seem to hurt to test it. It is definitely not used anywhere I know with gRPC-Web.

akshayjshah (approved):

PR #643, opened 2023-11-29
Fix "deadline exceeded" issues discovered in conformance tests

jhump:
When adding timeout tests to the conformance suite in https://github.com/connectrpc/conformance/pull/715 and https://github.com/connectrpc/conformance/pull/716, @smaye81 encountered several issues where the responses from the connect-go client would sometimes have a code of `unavailable` or `unknown`.

The `unavailable` code was happening [here](https://github.com/connectrpc/connect-go/blob/main/duplex_http_call.go#L269-L271). The `wrapIfContextError` above it was not successful because, sadly, a timeout error during dialing can get returned as `os.ErrDeadlineExceeded` instead of `context.DeadlineExceeded` 😭. (I filed a Go bug for this and added a work-around to `wrapIfContextError`.) 

The `unknown` codes were when a deadline exceeded error was returned from a call to write a message. The code was extremely inconsistent about handling of context errors and connect errors that could be returned from a call to `Write`. I've updated it so that all code that writes or reads now checks for context and connect errors, though many places (which read and write from a buffer or a pipe) can't actually trigger it. I figured it was safer (and more consistent) to have the checks everywhere instead of trying to be selective where it could or couldn't happen (which requires non-trivial analysis and could also be brittle if this code is changed in the future).

I tried creating a repro test case that reconstructed what was happening in the failing conformance test case. But I wasn't able to find a formulation that could actually do so consistently. So what I did to repro/test this was put the conformance tests in a loop, running them until they failed. Before these fixes, they semi-regularly failed (at least once every five runs). With this fix, I let it run over 500 times and still zero failures.


compression.go
```diff
@@ -111,8 +115,12 @@ func (c *compressionPool) Compress(dst *bytes.Buffer, src *bytes.Buffer) *Error
 	if err != nil {
 		return errorf(CodeUnknown, "get compressor: %w", err)
 	}
-	if _, err := io.Copy(compressor, src); err != nil {
+	if _, err := src.WriteTo(compressor); err != nil {
```
jhump, line 118: This was just for consistency/symmetry with the decompress code above, which directly calls `dst.ReadFrom` instead of using `io.Copy`.

envelope.go
```diff
@@ -252,7 +261,7 @@ func (r *envelopeReader) Read(env *envelope) *Error {
 	if r.readMaxBytes > 0 && size > int64(r.readMaxBytes) {
 		_, err := io.CopyN(io.Discard, r.reader, size)
 		if err != nil && !errors.Is(err, io.EOF) {
-			return errorf(CodeUnknown, "read enveloped message: %w", err)
+			return errorf(CodeResourceExhausted, "message is larger than configured max %d - unable to determine message size: %w", r.readMaxBytes, err)
```
jhump, line 264: This was made consistent with the exact same logic that lives in the decompression code. Its handling (always returning "resource exhausted" since we know the message is too big) seemed better, and its error message more helpful, so I copy+pasted it here.

error.go
```diff
@@ -292,6 +293,12 @@ func wrapIfContextError(err error) error {
 	if errors.Is(err, context.DeadlineExceeded) {
 		return NewError(CodeDeadlineExceeded, err)
 	}
+	// Ick, some dial errors can be returned as os.ErrDeadlineExceeded
+	// instead of context.DeadlineExceeded :(
+	// https://github.com/golang/go/issues/64449
+	if errors.Is(err, os.ErrDeadlineExceeded) {
+		return NewError(CodeDeadlineExceeded, err)
+	}
```
akshayjshah, line 301: Confirming: whenever we hit this case, the actual context has been cancelled - correct?

akshayjshah (approved):
LGTM. Before merging, please do respond to my question about `os.ErrDeadlineExceeded`.

Also, if the best way to flush this error out is to run the conformance tests repeatedly...should we do that on one platform in CI? If it's too painful to do on every PR, we could at least do a cron job and run it nightly.

error.go
jhump: Okay, I think I've now squashed _all_ of the bugs.

The upside is that the test seemed to fail pretty consistently without all of these fixes, so I think it will be a reasonably good way to identify issues in CI. But there are a couple of downsides:
1. It never repro'ed the issues with the race detector enabled. So I have the test currently running only when the race detector is disabled. And so there is now an extra command run for `make test` that just runs this one test, without `-race`.
2. When there are no bugs, it slows down CI by a full 20 seconds, hoping to find a non-existent bug.

See the last three commits for details:
1.  d5965e8f21e308b2ce9b9fefc868a763f63634cd adds the test
2. 39a162c3ded6549ab0379ab2f40e371b8883b86d updates the test so the output is more useful in tracking down the source of the error; this also makes sure to use the gRPC protocol, too, in case there protocol-specific places that fail to classify the error correctly. (It just uses gRPC-Web right now, since that is the same protocol handler implementation and basically a superset of activity, since it includes add'l body writes for the trailers.)
3. d42edba030949fe950ffc486fb7eb5570f7d49e3 fixes the three add'l places found by the test that were incorrectly classifying errors. One of them was wrapping a `*connect.Error` with another `*connect.Error` and changing the code to "unknown". Oops!

client_ext_norace_test.go
```diff
@@ -0,0 +1,268 @@
...
+				httpClient.CloseIdleConnections()
+				// Make sure to give a little time for the OS to release socket resources
+				// to prevent resource exhaustion.
+				time.Sleep(time.Millisecond / 2)
+				return pingv1connect.PingServicePingProcedure, rpcErrors{recvErr: err}
+			},
+		)
+	})
+
+	// This run is much creates significantly more load than the above one, but re-uses
```
emcfarlane, line 85:
```suggestion
	// This run creates significantly more load than the above one, but re-uses
```

client_ext_norace_test.go
```diff
@@ -0,0 +1,268 @@
...
+	t.Cleanup(svr.Close)
+
+	// This run creates a new connection for each RPC to verify that timeouts during dialing
+	// won't cause issues. This is historically easier to reproduce, so it uses a smaller
+	// duration, no concurrency, and fewer iterations. This is important because if we used
+	// a new connection for each RPC in the bigger test scenario below, we'd encounter other
+	// issues related to overwhelming the loopback interface and exhausting ephemeral ports.
+	t.Run("dial", func(t *testing.T) {
+		t.Parallel()
+		transport, ok := svr.Client().Transport.(*http.Transport)
```
emcfarlane, line 64:
```suggestion
		transport := svr.Transport()
```

client_ext_norace_test.go
```diff
@@ -0,0 +1,268 @@
...
+		if !assert.True(t, ok) {
+			t.FailNow()
+		}
+		testClientDeadlineBruteForceLoop(t,
+			5*time.Second, 5, 1,
+			func(ctx context.Context) (string, rpcErrors) {
+				httpClient := &http.Client{
+					Transport: transport.Clone(),
+				}
```
emcfarlane, line 73: Why the `transport.Clone()` and not use a new client? I think this is effectively the same:
```go
httpClient := svr.Client()
```

client_ext_norace_test.go
```diff
@@ -0,0 +1,268 @@
...
+				client := pingv1connect.NewPingServiceClient(httpClient, svr.URL)
+				_, err := client.Ping(ctx, connect.NewRequest(&pingv1.PingRequest{Text: "foo"}))
+				httpClient.CloseIdleConnections()
```
emcfarlane, line 76: Why do you call `CloseIdleConnections`?

emcfarlane (commented):
Potentially we could move these new test cases that relies on thrashing to a test on the writers and readers by controlling when errors are injected? This wouldn't help in uncovering new issues but we could build a suite of known error cases and protect against them. This would ensure the testcases are able to reproduce the issues under `-race`.

client_ext_norace_test.go
jhump: Possibly. I'll look into it. It wasn't obvious when I was looking at it that this would create a new a transport (and new associated connection pool).

But I could definitely simplify a little. I originally had some concurrency in this test (the dial test) as well, and I wanted to make sure there was no chance of any of the tests re-using, since I'm trying to make sure every RPC has to re-dial. But I've since changed this to use no parallelism (one goroutine) because (1) it wasn't necessary as the dial error repro's fairly consistently and (2) it would run into issues with exhausting ephemeral ports. In fact, the latter could happen even with one goroutine (hence the addition of one millisecond sleep after trying to close idle connections, to make sure ephemeral ports are released).

client_ext_norace_test.go
jhump: The name of this case is `dial` because I am testing deadline-exceeded issues during dial. That means we want to use a different connection for every operation. The comment above explains it.

jhump:
> This wouldn't help in uncovering new issues but we could build a suite of known error cases and protect against them.

🤷. While the test would certainly be faster, it wouldn't be nearly as robust or confidence-inspiring. Also, I worry that adding the ability inject arbitrary errors into all of the different places could make the non-test code much more complicated.

client_ext_norace_test.go
jhump: `svr` has no method named `Transport()`.

client_ext_norace_test.go
jhump: `svr.Client()` re-uses the same client over and over, so definitely don't want to do that. It's fine for the `read-write` cases to use that client (and they do), but if we mix clients (and thus mix transports), we can't be sure that we are actually dialing for each RPC.

client_ext_norace_test.go
jhump: I updated the comment here (that was on the next line) to make this aspect more clear.

jhump:
@emcfarlane, @akshayjshah, since the test that reproduces the issue is slow, I've disabled it for now. It's actually enabled, but only when the race detector is off. And since the Makefile and CI always run the tests with `-race`, it's effectively a manual-only test.

Seem good enough to merge as is?

client_ext_norace_test.go
emcfarlane: Sorry, this was my bad drive by comments. I thought it was using `memhttptest`. Please ignore.

jhump:
After a chat w/ Akshay, I added it as a separate CI job, so it runs in parallel with other tests, and only for latest Go and linux. Looking at the timing above, it is faster than the main `ci` job, so isn't making CI jobs otherwise take longer (so time to merge is unaffected), they just use a little more resources.

error.go
akshayjshah: > One of them was wrapping a *connect.Error with another *connect.Error and changing the code to "unknown". Oops!

😬 Yikes!

.github/workflows/ci.yaml
```diff
@@ -39,3 +39,17 @@ jobs:
...
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+      - name: Install Go
+        uses: actions/setup-go@v4
+        with:
+          # only the latest
+          go-version: 1.21.x
+      - name: Run Slow Tests
+        run: make slowtest
```
akshayjshah, line 55: Talked to Josh live about this, but I'd love for this to be carved up a little differently. Locally, I'd really like `make` and `make test` to run these slower smoke tests. In CI, the highly-matrixed tests should run all the tests _except_ the smoke tests. The new job (`slowtest`) should run the slow tests, but it would be okay with me if it _also_ ran the faster tests.

I don't care that much whether we carve up the tests with an environment variable, `testing.Short()`, or with build tags.

jhump: `make test` now runs both short and slow tests. CI "Test" job only runs short tests, and there's a separate "Slow Tests" job for the others.

Also, while factoring out slow tests to a parallel job, I did the same for conformance tests, to shorten the critical path through the CI testing.

jhump: Because the main CI job must re-build the code (to instrument with race detector) and also does linting, the new shorttest job is actually faster than that other job (for now at least). So this new job does not increase the total duration. The longest job is still Windows, by a good margin.

PR #644, opened 2023-11-30
Improve GET request handling: clients should not include content headers, servers should disallow a body

jhump:
More stuff that @smaye81 found when developing conformance test cases, FTW!

The client should not include content headers ("content-type", "content-encoding", "content-length") since there is no body and everything is in the query string.

And the server should explicitly disallow a request body. It accepts content headers, but they are ignored and allowed as long as there is no body content.


jchadwick-buf (approved):
Nice! Also, I probably would be wise to implement a similar provision in Envoy, which is currently very permissive regarding protocol violations.

akshayjshah (approved):

PR #646, opened 2023-12-04
Refactor http calls to support message payloads

emcfarlane:
To enable client retries, the http calls now operates directly on message payloads instead of relying solely on `io.Writer`s. This refactor replaces write operations with send operations at message boundaries. It introduces a new interface, `messagePayload`, implemented by `*bytes.Reader`s and `*envelopes`, enabling them to adopt sender operations.

### Changes Made:
- A new interface, `messagePayload`, implemented by `*bytes.Reader` and `*envelope`, allowing for adoption of sender operations.
- A new interface, `messageSender`, implemented by `*duplexHTTPCall`, which acts on `messagePayloads` to send the message.
- An adapter `writeSender` for `http.ResponseWriter` to implement `messageSender` via converting send payloads to write operations. 
- Replaced write operations with send operations at message boundaries for handling message payloads.

### Objective:
The goal of this refactor is to facilitate client retries by directly manipulating message payloads rather than relying solely on `io.Writer`s. The `messagePayload` interface is a readable, seekable and sized payload which allows it to be replayed.

This PR is part of the fix for #609 . It implements the groundwork for the handling of message payloads.

### DIFF
--- a/envelope.go
+++ b/envelope.go
@@ -110,10 +110,10 @@ func (e *envelope) Seek(offset int64, whence int) (int64, error) {
 
 // Len returns the number of bytes of the unread portion of the envelope.
 func (e *envelope) Len() int {
-	if e.offset >= int64(e.Data.Len()+5) {
-		return 0
+	if length := int64(e.Data.Len()) + 5 - e.offset; length > 0 {
+		return int(length)
 	}
-	return int(int64(e.Data.Len()+5) - e.offset)
+	return 0
 }
 
 type envelopeWriter struct {
@@ -127,8 +127,8 @@ type envelopeWriter struct {
 
 func (w *envelopeWriter) Marshal(message any) *Error {
 	if message == nil {
-		// Send a zero-length message.
-		payload := bytes.NewReader(nil)
+		// Send no-op message to create the request and send headers.
+		payload := nopPayload{}
 		if _, err := w.sender.Send(payload); err != nil {
 			if connectErr, ok := asError(err); ok {
 				return connectErr
--- a/envelope.go
+++ b/envelope.go
@@ -40,16 +40,84 @@ var errSpecialEnvelope = errorf(
 // message length. gRPC and Connect interpret the bitwise flags differently, so
 // envelope leaves their interpretation up to the caller.
 type envelope struct {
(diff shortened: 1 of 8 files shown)


envelope.go
```diff
@@ -40,16 +40,84 @@ var errSpecialEnvelope = errorf(
...
+	e.offset = abs
+	return abs, nil
+}
+
+// Len returns the number of bytes of the unread portion of the envelope.
+func (e *envelope) Len() int {
+	if e.offset >= int64(e.Data.Len()+5) {
+		return 0
+	}
+	return int(int64(e.Data.Len()+5) - e.offset)
```
jhump, line 116: Maybe remove the repeated computation here?
```suggestion
	if length := int64(e.Data.Len()+5) - e.offset; length > 0 {
		return int(length)
	}
	return 0
```

envelope.go
```diff
@@ -59,7 +127,9 @@ type envelopeWriter struct {
 
 func (w *envelopeWriter) Marshal(message any) *Error {
 	if message == nil {
-		if _, err := w.writer.Write(nil); err != nil {
+		// Send a zero-length message.
```
jhump, line 130: This is misleading since a zero-length message on a stream is still a message, and this is really sending _nothing_ (since it elides any envelope).
```suggestion
		// Send a no-op (just flushing headers/making sure request sent)
```

envelope.go
```diff
@@ -59,7 +127,9 @@ type envelopeWriter struct {
...
+		payload := bytes.NewReader(nil)
```
jhump, line 131: Might be better to use a no-op empty struct type, so there's no allocation here and also to make it clear we aren't actually sending a message.

jhump (approved):
Nice! I left some comments, but they are mostly nits. LGTM

envelope.go
emcfarlane: Created the `nopPayload` type that implements `messagePayload`.

envelope.go
emcfarlane: Reworded as: `// Send no-op message to create the request and send headers.`

jhump (approved):

Issue #637, opened 2023-11-16
HandlerOption for ErrorWriter to fallback to JSON if no header is provided


weeco:
**Is your feature request related to a problem? Please describe.**
I have an API exposed via different protocols (connect, grpc and connect-gateway/http). This API is auth protected (Bearer token in header or via Cookie-based session) and I'm using middlewares for that. I'd like to return helpful errors (missing, invalid, expired bearer token for example) for all of these APIs. Thus, the ErrorWriter is almost the perfect tool.

One thing I do not really want to enforce is setting an apropriate `Content-Type` in the request - for example if someone just wants to test an API endpoint via the Browser and the content-type is not set the ErrorWriter would not work because it's missing the Content-Type header with the protocol information.

**Describe the solution you'd like**

I would like to have a little bit more control (e.g. an extra HandlerOption) when constructing a new ErrorWriter, so that I could configure the ErrorWriter to fallback to writing JSON if no header is set. 

**Describe alternatives you've considered**
 I could inject the Content-Type header when no Content-Type header is set. It is a bit silly to do so, because the sole purpose would be to trick the ErrorWriter to write JSON.

**Additional context**

```go
// In my middleware somewhere...
// This condition will evaluate to false if no Content-Type is set (e.g. request from the Browser), thus no error will be written by my ErrorWriter
if errorWriter.IsSupported(r) {
	err := apierrors.NewConnectError(
		connect.CodeUnauthenticated,
		errors.New("no bearer token found in authorization header"),
		apierrors.NewErrorInfo(commonv1alpha1.Reason_REASON_NO_AUTHENTICATION_TOKEN.String()),
	)
	_ = errorWriter.Write(w, r, err)
	return
}
```

`errorWriter.Write()` will only work if a known Content-Type header is set. Maybe there's a good reason why this should always be enforced, but I wouldn't know what reason that is.

mattrobenolt:
In my experience implementing this, since I do very similar (auth with an http middleware), I chose to fall back to plain text as the default:

Something like:

```go
if errorWriter.IsSupported(r) {
	errorWriter.Write(w, r, connect.NewError(connect.CodeUnauthenticated, err))
} else {
	w.WriteHeader(http.StatusUnauthorized)
	w.Write([]byte(err.Error()))
}
```

Obviously this behavior is subjective for what's right/wrong, but to me, it felt if the client isn't requesting JSON, there's no reason to try to return JSON. We can return some human readable text as `text/plain`.

[...]

PR #654, opened 2023-12-12
Fix ErrorWriter connect GET protocol classifier

emcfarlane:
`ErrorWriter` will now correctly classifying connect GET requests if the connect protocol version header is present or the connect protocol version is set in the URL query parameters. To handle connect GET requests without the protocol version the ErrorWriter will write errors as connect unary errors if the protocol is unknown. This ensures the error is always written to the client. The connect unary error payload is always in JSON and therefore makes a nice human-readable format for the caller. Clients are still recommended to check `IsSupported` first to ensure the protocol will match the desired and to provide fallback to other error writers if desired.

Fix for #637 . Defaults to connect unary errors. To support other fallbacks use the `(*ErrorWriter).IsSupported()` check before calling `(*ErrorWriter).Write(err)`.

### DIFF
--- a/connect_ext_test.go
+++ b/connect_ext_test.go
@@ -394,6 +394,14 @@ func TestServer(t *testing.T) {
 					connect.WithSendGzip(),
 				)
 			})
+			t.Run("json_get", func(t *testing.T) {
+				run(
+					t,
+					connect.WithProtoJSON(),
+					connect.WithHTTPGet(),
+					connect.WithHTTPGetMaxURLSize(1024, true),
+				)
+			})
 		})
 		t.Run("grpc", func(t *testing.T) {
 			t.Run("proto", func(t *testing.T) {
--- a/error_writer.go
+++ b/error_writer.go
@@ -137,6 +137,8 @@ func (w *ErrorWriter) Write(response http.ResponseWriter, request *http.Request,
 	case grpcWebProtocol:
 		setHeaderCanonical(response.Header(), headerContentType, ctype)
 		return w.writeGRPCWeb(response, err)
+	case unknownProtocol, connectUnaryProtocol:
+		fallthrough
 	default:
 		// Unary errors are always JSON. Unknown protocols are treated as unary
 		// because they are likely to be Connect clients and will still be able to
--- a/error_writer.go
+++ b/error_writer.go
@@ -122,18 +122,12 @@ func (w *ErrorWriter) IsSupported(request *http.Request) bool {
 // Write an error, using the format appropriate for the RPC protocol in use.
 // Callers should first use IsSupported to verify that the request is using one
 // of the ErrorWriter's supported RPC protocols. If the protocol is unknown,
-// Write will send the error in the Connect unary format.
+// Write will send the error as unprefixed, Connect-formatted JSON.
 //
 // Write does not read or close the request body.
 func (w *ErrorWriter) Write(response http.ResponseWriter, request *http.Request, err error) error {
 	ctype := canonicalizeContentType(getHeaderCanonical(request.Header, headerContentType))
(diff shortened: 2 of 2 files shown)


connect_ext_test.go
```diff
@@ -394,6 +394,14 @@ func TestServer(t *testing.T) {
 					connect.WithSendGzip(),
 				)
 			})
+			t.Run("json_get", func(t *testing.T) {
```
emcfarlane, line 397: As a test case I've extended `TestServer` to include an option with GET set. We exercise error writer in the middleware tests and this covers the case with ping as a GET and a large ping as a fallback to POST.

error_writer.go
```diff
@@ -46,74 +45,107 @@ func NewErrorWriter(opts ...HandlerOption) *ErrorWriter {
...
+// protocolType is one of the supported RPC protocols.
+type protocolType uint8
+
+const (
+	unknownProtocol protocolType = iota
+	connectUnaryProtocol
+	connectStreamProtocol
+	grpcProtocol
+	grpcWebProtocol
+)
```
akshayjshah, line 85: nit: these are a separate type - please follow the style guide and put them at the top of the page.

jhump:
It appears this would also address #637. I think the only potentially controversial part in here is the change to always fall back to the Connect protocol's JSON error format output even when the incoming protocol is cannot be determined. I'm okay with it, but am curious if @akshayjshah or @mattrobenolt have objections.

error_writer.go
```diff
@@ -46,74 +56,96 @@ func NewErrorWriter(opts ...HandlerOption) *ErrorWriter {
...
+	if _, ok := w.grpcContentTypes[ctype]; ok {
+		return grpcProtocol
+	}
+	if _, ok := w.grpcWebContentTypes[ctype]; ok {
+		return grpcWebProtocol
+	}
+	// Check for Connect-Protocol-Version header or connect protocol query
+	// parameter to support connect GET requests.
+	if request.Method == http.MethodGet {
+		connectVersion := request.Header.Get(connectProtocolVersion)
```
mattrobenolt, line 104: nit, but probably want `getHeaderCanonical` here.

mattrobenolt:
I think this is fine. This behavior wouldn't affect us. I use ErrorWriter.IsSupported() to spit back a plaintext error, and we don't use GET requests for anything.

error_writer.go
emcfarlane: Thanks!

jhump (approved):

error_writer.go
```diff
@@ -46,74 +56,96 @@ func NewErrorWriter(opts ...HandlerOption) *ErrorWriter {
...
+// of the ErrorWriter's supported RPC protocols. If the protocol is unknown,
+// Write will send the error in the Connect unary format.
 //
 // Write does not read or close the request body.
 func (w *ErrorWriter) Write(response http.ResponseWriter, request *http.Request, err error) error {
 	ctype := canonicalizeContentType(getHeaderCanonical(request.Header, headerContentType))
-	if _, ok := w.unaryConnectContentTypes[ctype]; ok {
-		// Unary errors are always JSON.
+	switch protocolType := w.classifyRequest(request); protocolType {
+	case connectUnaryProtocol, unknownProtocol:
```
akshayjshah, line 131: Can we make this the `default` branch below, to clarify that we're now never going to hit the unknown protocol case?

akshayjshah (commented):
This change is great! Supporting GETs was an unfortunate oversight, and defaulting to JSON makes sense. A few nits about the code structure and comments, but I'd love to get this landed and released ASAP :)

akshayjshah (approved):

bufdev (approved):

Issue #645, opened 2023-12-01
Incorrect client disconnection error codes


emcfarlane:
**Describe the bug**

On client disconnections for both HTTP1 and HTTP2 transports we should detect the client has disconnected and use an appropriate error code. Currently disconnects can see a `connect.CodeUknown` causing erroneous error reporting. To detect a client disconnect we could check the context for cancellation (see docs from deprecated [http.CloseNotifier](https://pkg.go.dev/net/http#CloseNotifier)). 

**To Reproduce**

Create a connection and close the underlying net.Conn to replicate a dropped client connection:

```go
var conn client.Conn // Captured client connection.
transport := server.Transport()
dialContext := transport.DialTLSContext
transport.DialTLSContext = func(ctx context.Context, network, addr string, cfg *tls.Config) (net.Conn, error) {
  conn, _ = dialContext(ctx, network, addr, cfg) // Capture the client connection.
  return conn, nil
}
stream := ... // Create stream
stream.Send(nil) // Send headers
conn.Close() // Abruptly close
// Check handler error codes
```
Handler errors will be (only checked the envelope flow):
- http1/read: `invalid_argument: protocol error: incomplete envelope: unexpected EOF`
- http1/write: `unknown: write envelope: io: read/write on closed pipe`
- http2/read: `invalid_argument: protocol error: incomplete envelope: client disconnected`
- http2/write: `unknown: write envelope: client disconnected`

In all cases checking the context error will return a `context.Cancelled` error. The http2 `client disconnect` string is from [this](https://github.com/golang/go/blob/master/src/net/http/h2_bundle.go#L3843) error.

**Environment (please complete the following information):**
- `connect-go` version or commit: `v0.1.12`
- `go version`: `go version go1.18.3 darwin/amd64`

**Additional context**

Thanks to the reports from:
- @pqn  https://bufbuild.slack.com/archives/CRZ680FUH/p1701370334963849
- @njiang747 https://bufbuild.slack.com/archives/CRZ680FUH/p1700005011055009

emcfarlane:
Related issue: https://github.com/golang/go/issues/52183 . Which seems to show`client disconnects` can be reported without a context cancellation.

jhump:
FWIW, closing the actual connection is not the same as a cancellation with http/2. With http/2, the client aborts the stream, but the connection remains open. With http/1.1, since multiplexing a single connection is not supported, the only way a client can cancel is to disconnect.

[...]

PR #659, opened 2023-12-21
Wrap errors with context cancellation codes

emcfarlane:
This PR wraps errors with the appropriate connect code of Cancelled or DeadlineExceeded if the context error is not nil.

Improves error handling for some well known error cases that do not surface context.Cancelled errors. For example HTTP2 "client disconnect" string errors are now raised with a Cancelled code not an Unknown. This lets handlers check the error code for better handling and reporting of errors.

Fix for #645


emcfarlane:
Looks like hitting some issues with testing and racing of ctx close:
```sh
--- FAIL: TestClientDisconnect (0.04s)
    --- FAIL: TestClientDisconnect/handler_writes (0.03s)
        connect_ext_test.go:2399:
            assertion:  assert.Equal
            got:        unknown
            want:       canceled

        connect_ext_test.go:2400:
            assertion:  assert.ErrorIs
            got:        <nil>
            want:       context canceled
```
Need to fix.

error.go
```diff
@@ -408,10 +428,18 @@ func wrapIfRSTError(err error) error {
 	}
 }
 
-func asMaxBytesError(err error, tmpl string, args ...any) *Error {
+// wrapIfMaxBytesError wraps errors returned reading from a http.MaxBytesHandler
+// whose limit has been exceeded.
+func wrapIfMaxBytesError(err error, tmpl string, args ...any) error {
```
emcfarlane, line 433: Converted `wrapIfMaxBytesError` to the same style as other error handling for consistency.

envelope.go
```diff
@@ -117,6 +118,7 @@ func (e *envelope) Len() int {
 }
 
 type envelopeWriter struct {
+	ctx              context.Context
```
emcfarlane, line 121: Wrapping the ctx in the envelope reader is a little odd, but would otherwise be needed to be stored on the stream interceptors and then passed down which makes this problem worse. Would be nice to find a better solution though!

chrispine:
So what's the status of this PR now? Did you still want to merge it, or is it no longer needed?

emcfarlane:
@chrispine I think this behaviour is still wanted to fix the issue linked. PR needs review.

envelope.go
```diff
@@ -209,6 +211,7 @@ func (w *envelopeWriter) marshal(message any) *Error {
 func (w *envelopeWriter) write(env *envelope) *Error {
 	if _, err := w.sender.Send(env); err != nil {
 		err = wrapIfContextError(err)
+		err = wrapWithContextError(w.ctx, err)
```
jhump, line 214: Perhaps this would be better to push the error classification into `sender.Send`? The sender (the `duplexHTTPCall`) has a reference to the request, which should also include the context.

Also, under what conditions would we ever call only one or the other of those context-wrap functions? Feels like they want to be a single function -- maybe even a method on `duplexHTTPCall` (so it can pull out the request context for the comparisons in the latter functions).

error.go
```diff
@@ -302,6 +302,26 @@ func wrapIfContextError(err error) error {
...
+	}
+	ctxErr := ctx.Err()
+	switch {
+	case errors.Is(ctxErr, context.Canceled):
+		return NewError(CodeCanceled, err)
+	case errors.Is(ctxErr, context.DeadlineExceeded):
+		return NewError(CodeDeadlineExceeded, err)
+	default:
+		return err
+	}
```
jhump, line 322: This doesn't seem quite right. I would expect cases that look more like what is in `wrapIfContextError` (which is why I suggested combining them above).

This could make for confusing errors -- `err` could be unrelated to the cancellation, but we'll still stamp it as canceled or deadline exceeded code. For example, it could be a `*http.MaxBytesError`, which will result in a _very_ confusing message for one of these codes.

envelope.go
emcfarlane: We call wrapWithContext more liberally to convert context errors but not to wrap errors with context codes. We will always call `wrapIfContext` with `wrapWithContext` but still need `wrapIfContext` separate so left them split out. A method on `duplexHTTPCall` would need a corresponding method on the handler side. The marshallers are used on both client/server so can handle it uniformly there.

error.go
emcfarlane: Changed the ordering of the calls to avoid any possibility of wrapping a max bytes error with context codes.

envelope.go
jhump: > We will always call `wrapIfContext` with `wrapWithContext` but still need `wrapIfContext` separate so left them split out.

Why do they need to be split? Under what cases do we care about the error _being_ a context error but if the context is actually done (or vice versa)? Also, the names are rather confusing. Just reading the above sentence is really not clear as to which is which.

error.go
jhump: I'm not sure this really addresses the concern; it addresses only the one example I gave. In the face of other errors that might occur, unrelated to the context timing out or being cancelled -- it will unconditionally wrap an unrelated error with a context-related code. I guess it's okay, since the error will have to be I/O- or protocol-related (low-level protocol, like HTTP/2 framing or flow control errors) at this point (I think?), so maybe it's most likely related to the context error even if non-obviously. But it feels brittle.

jhump (commented):
Let's at least make the names more clear. I think `wrapIfContextError` is reasonably clear. Maybe the other could be called `wrapIfContextDone`? Although I'm still not convinced these should be separate - it seems like wherever we're doing one, we'd likely want to do both.

error.go
emcfarlane: It was to avoid handling specific errors, like the http2 client disconnect, and instead use the context to capture a more suitable error code. Could be unrelated but most likely occurs due to an IO issues related to the context like you said.

envelope.go
emcfarlane: We check for context errors on non IO related errors like the wrapper to the stream handlers. I've merged the call to the `wrapIfContextError` from `wrapIfContextDone`.

emcfarlane: https://github.com/emcfarlane/connect-go/blob/c3324072f50f3882a98e4ee846ee7511d3911efa/error.go#L279

Which we don't currently have an easy way to access the context from.

jhump (approved):

PR #679, opened 2024-02-07
Client should verify response content-type

jhump:
@smaye81 found some more bugs with new conformance test cases.

The client is not verifying that the response has the correct content-type. So this change adds that verification.

Some tests, that had simple handlers that weren't setting a valid content-type, had to be updated. Sorry for the large diff stats, but it's not as bad as it looks. This PR adds three new table-driven tests that account for 75% of the lines of code.

### DIFF
--- a/protocol_connect.go
+++ b/protocol_connect.go
@@ -511,6 +511,21 @@ func (cc *connectUnaryClientConn) validateResponse(response *http.Response) *Err
 		}
 		cc.responseTrailer[strings.TrimPrefix(k, connectUnaryTrailerPrefix)] = v
 	}
+	err := connectValidateUnaryResponseContentType(
+		cc.marshaler.codec.Name(),
+		cc.duplexCall.Method(),
+		response.StatusCode,
+		response.Status,
+		getHeaderCanonical(response.Header, headerContentType),
+	)
+	if err != nil {
+		if IsNotModifiedError(err) {
+			// Allow access to response headers for this kind of error.
+			// RFC 9110 doesn't allow trailers on 304s, so we only need to include headers.
+			err.meta = cc.responseHeader.Clone()
+		}
+		return err
+	}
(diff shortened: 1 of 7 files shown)


protocol_connect.go
```diff
@@ -511,6 +511,21 @@ func (cc *connectUnaryClientConn) validateResponse(response *http.Response) *Err
 		}
 		cc.responseTrailer[strings.TrimPrefix(k, connectUnaryTrailerPrefix)] = v
 	}
+	err := connectValidateUnaryResponseContentType(
```
emcfarlane, line 514: Could this be moved into the validate response function?

jhump: It is. Do you mean inlined? I had made it a separate function to make it easier to test.

emcfarlane: Sorry misread, thought the method on the duplexcall was due to it not being part of this function, but thats for the request params.

emcfarlane (approved):

protocol_connect.go
```diff
@@ -511,6 +511,21 @@ func (cc *connectUnaryClientConn) validateResponse(response *http.Response) *Err
...
+		cc.marshaler.codec.Name(),
+		cc.duplexCall.Method(),
+		response.StatusCode,
+		response.Status,
+		getHeaderCanonical(response.Header, headerContentType),
+	)
+	if err != nil {
```
akshayjshah, line 521: Can we scope `err` to the `if` block? It's visually a little ugly, but the functions in this portion of the code are long enough that I'd love to minimize scope where possible.

akshayjshah (approved):

protocol_connect.go
jhump: I had not done so just because I find it a bit hard to read when the `if` keyword is far from the actual condition. But if that's accepted style, and even preferred in this case, I'll change it.

PR #681, opened 2024-02-08
Reduce scope of `err` var in response content-type checks

jhump:
This also formats the grpc version of the check so it is consistent with the other two.

Addresses comment in prior PR: https://github.com/connectrpc/connect-go/pull/679#discussion_r1483403176


akshayjshah (approved):
Thank you!

PR #683, opened 2024-02-08
Remove redundant work when validating unary Connect response

mattrobenolt:
We can swap TrimPrefix for just a string slice here since we're already guarding behind HasPrefix. TrimPrefix internally runs HasPrefix again, so we can just skip the redundant work.


akshayjshah (approved):
Love it - thanks, Matt!

PR #685, opened 2024-02-09
Verify that a trailers-only response is well-formed

jhump:
Previously, connect-go would accept a trailers-only gRPC response (where the trailers are in the HTTP headers, signaled by the presence of a "grpc-status" key in the headers) and assume it was valid w/out further verification.

However, it should reject trailers-only responses that _also_ include response body data and/or HTTP trailers, after the HTTP headers as those are malformed responses.

This _would_ have been caught by a conformance test, as exactly this sort of test is in the queue. But I haven't written those tests yet and just happened to notice when I was looking into something else in the reference client (working on somewhat-related checks of the wire-level details for gRPC-Web responses).

### DIFF
--- a/envelope.go
+++ b/envelope.go
@@ -305,9 +305,6 @@ func (r *envelopeReader) Read(env *envelope) *Error {
 			return NewError(CodeUnknown, err)
 		}
 		err = wrapIfContextError(err)
-		if connectErr, ok := asError(err); ok {
-			return connectErr
-		}
 		// Something else has gone wrong - the stream didn't end cleanly.
 		if connectErr, ok := asError(err); ok {
 			return connectErr
--- a/protocol_grpc.go
+++ b/protocol_grpc.go
@@ -727,7 +727,7 @@ func grpcVerifyTrailersOnly(response *http.Response) *Error {
 		// response) AND trailers after the body.
 		return errorf(
 			CodeInternal,
-			"corrupt response: HTTP trailers should not be present in trailers-only response",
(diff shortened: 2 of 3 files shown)


envelope.go
```diff
@@ -305,9 +305,6 @@ func (r *envelopeReader) Read(env *envelope) *Error {
 			return NewError(CodeUnknown, err)
 		}
 		err = wrapIfContextError(err)
-		if connectErr, ok := asError(err); ok {
-			return connectErr
-		}
```
jhump, line 310: Not related, but I happened to notice that this was being twice. (On line 312, these exact three lines are repeated. Oops, I think this is from a change of mine.)

protocol_grpc.go
```diff
@@ -696,6 +701,39 @@ func grpcValidateResponse(
...
+	// Make sure there's nothing in the body.
+	if numBytes, err := discard(response.Body); err != nil {
+		err = wrapIfContextError(err)
+		if connErr, ok := asError(err); ok {
+			return connErr
+		}
+		return errorf(CodeInternal, "corrupt response: I/O error after trailers-only response: %w", err)
+	} else if numBytes > 0 {
+		return errorf(CodeInternal, "corrupt response: %d extra bytes after trailers-only response", numBytes)
+	}
```
jhump, line 714: This was largely copied from the handling of end-stream messages (in `envelopeReader.Unmarshal`), where we also want to verify that there is no subsequent response body data.

akshayjshah (approved):
Apart from some suggestions on the error text, this LGTM.

PR #688, opened 2024-02-13
Drop @type from Connect error detail debug string

jhump:
This makes the output of a connect-go server look more like the examples in the docs. For example, [this section](https://connectrpc.com/docs/protocol/#error-end-stream) shows this example:
```json
{
  "code": "unavailable",
  "message": "overloaded: back off and retry",
  "details": [
    {
      "type": "google.rpc.RetryInfo",
      "value": "CgIIPA",
      "debug": {"retryDelay": "30s"},
    }
  ]
}
```
However, actual output from a connect-go server would instead have a "debug" entry that looked like this:
```json
{"@type": "type.googleapis.com/google.rpc.RetryInfo", "retryDelay": "30s"}
```
The type was redundantly stated as it was marshaling the `google.protobuf.Any` wrapper to JSON instead of the contained message.

This removes the discrepancy. Error output from a connect-go server will be a little more concise and more closely resemble the docs example with this change.

### DIFF
--- a/error.go
+++ b/error.go
@@ -49,8 +49,9 @@ var (
 // The [google.golang.org/genproto/googleapis/rpc/errdetails] package contains a
 // variety of Protobuf messages commonly used as error details.
 type ErrorDetail struct {
-	pb       *anypb.Any
-	wireJSON string // preserve human-readable JSON
+	pbAny    *anypb.Any
+	pbInner  proto.Message // if nil, must be extracted from pbAny
+	wireJSON string        // preserve human-readable JSON
 }
 
 // NewErrorDetail constructs a new error detail. If msg is an *[anypb.Any] then
@@ -59,13 +60,13 @@ type ErrorDetail struct {
 func NewErrorDetail(msg proto.Message) (*ErrorDetail, error) {
 	// If it's already an Any, don't wrap it inside another.
 	if pb, ok := msg.(*anypb.Any); ok {
-		return &ErrorDetail{pb: pb}, nil
+		return &ErrorDetail{pbAny: pb}, nil
 	}
 	pb, err := anypb.New(msg)
 	if err != nil {
 		return nil, err
 	}
-	return &ErrorDetail{pb: pb}, nil
+	return &ErrorDetail{pbAny: pb, pbInner: msg}, nil
 }
 
 // Type is the fully-qualified name of the detail's Protobuf message (for
@@ -79,13 +80,13 @@ func (d *ErrorDetail) Type() string {
 	//
 	// If we ever want to support remote registries, we can add an explicit
 	// `TypeURL` method.
-	return typeNameFromURL(d.pb.GetTypeUrl())
+	return typeNameFromURL(d.pbAny.GetTypeUrl())
 }
 
 // Bytes returns a copy of the Protobuf-serialized detail.
 func (d *ErrorDetail) Bytes() []byte {
-	out := make([]byte, len(d.pb.GetValue()))
-	copy(out, d.pb.GetValue())
+	out := make([]byte, len(d.pbAny.GetValue()))
+	copy(out, d.pbAny.GetValue())
 	return out
 }
 
@@ -93,7 +94,10 @@ func (d *ErrorDetail) Bytes() []byte {
 // Detail into a strongly-typed message. Typically, clients use Go type
 // assertions to cast from the proto.Message interface to concrete types.
 func (d *ErrorDetail) Value() (proto.Message, error) {
-	return d.pb.UnmarshalNew()
+	if d.pbInner != nil {
+		return d.pbInner, nil
+	}
+	return d.pbAny.UnmarshalNew()
 }
 
 // An Error captures four key pieces of information: a [Code], an underlying Go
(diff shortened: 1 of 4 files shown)


protocol_connect.go
```diff
@@ -1151,10 +1151,15 @@ func (d *connectWireDetail) MarshalJSON() ([]byte, error) {
...
-		wire.Debug = json.RawMessage(debug)
+	if wire.Value != "" { // don't bother sending `{}`
+		msg, err := d.pb.UnmarshalNew()
+		if err == nil {
+			var codec protoJSONCodec
+			debug, err := codec.Marshal(msg)
+			if err == nil {
+				wire.Debug = debug
+			}
+		}
```
akshayjshah, line 1162: In the common case where the user created the error detail using `NewErrorDetail`, this feels a little wasteful - we had the original protobuf message in hand. Could we change the `ErrorDetail` struct to also keep a reference to the original protobuf message so we don't always need to go through an unmarshal here?

protocol_connect.go
```diff
@@ -1151,10 +1151,15 @@ func (d *connectWireDetail) MarshalJSON() ([]byte, error) {
...
 	}
 	return json.Marshal(wire)
```
akshayjshah, line 1164: I'm really looking forward to jsonv2's `jsontext` package so we don't need to go through all this nonsense with anonymous structs and reflection :/

protocol_connect.go
jhump: Done.

error.go
```diff
@@ -79,21 +80,24 @@ func (d *ErrorDetail) Type() string {
...
 
 // Value uses the Protobuf runtime's package-global registry to unmarshal the
 // Detail into a strongly-typed message. Typically, clients use Go type
 // assertions to cast from the proto.Message interface to concrete types.
 func (d *ErrorDetail) Value() (proto.Message, error) {
-	return d.pb.UnmarshalNew()
+	if d.pbInner != nil {
+		return d.pbInner, nil
+	}
+	return d.pbAny.UnmarshalNew()
```
akshayjshah, line 100: Previously, the returned messages were safe to mutate b/c they were pointers to freshly-allocated structs. It feels like we may want to preserve that property and not expose `pbInner`?

protocol_connect.go
```diff
@@ -1146,15 +1146,20 @@ func (d *connectWireDetail) MarshalJSON() ([]byte, error) {
...
+		Type:  typeNameFromURL(d.pbAny.GetTypeUrl()),
+		Value: base64.RawStdEncoding.EncodeToString(d.pbAny.GetValue()),
 	}
 	// Try to produce debug info, but expect failure when we don't have
 	// descriptors.
-	var codec protoJSONCodec
-	debug, err := codec.Marshal(d.pb)
-	if err == nil && len(debug) > 2 { // don't bother sending `{}`
-		wire.Debug = json.RawMessage(debug)
+	if wire.Value != "" { // don't bother sending `{}`
```
akshayjshah, line 1154: You'd know better than me, but I think that the zero value of some of the WKTs (like timestamps) produces useful JSON output. I think we'd be okay without this check: sending the zero value of an error detail feels like an uncommon enough case that we don't need to optimize for it.

error.go
jhump: Yeah, that's a good point.

protocol_connect.go
jhump: 🤷, it's debatable how useful that is. But I had changed it because a value that is two or fewer characters isn't always `{}` with WKTs, too. While it may be a contrived (and not realistic) case, a `google.protobuf.IntValue` could serialize as the JSON value `1`.

I guess I can change the comparison to directly compare the resulting JSON output to `"{}"`, but had done this hoping it was a convenient way to skip the serialization step. (If you see that the `value` property for a WKT is `""`, it is intuitive enough that it's a zero value whose `debug` value is not particularly valuable?)

jhump: Also, I just realized that a non-zero value for `google.protobuf.Value` could still produce `{}`, so it might be confusing that we just drop the `debug` key in this case. (Admittedly, I can't think of a legit reason to use a error detail of type `google.protobuf.Value`...).

But perhaps the best check is to skip the key if _both_ `wire.Value == "" && serializedJSON == "{}"`.

jhump: I implemented what I put in that last comment. WDYT?

error.go
jhump: Done. I've changed it to use `proto.Clone` because it should be a little more efficient than `*anypb.Any.UnmarshalNew`. But for serializing, we can still directly refer to `pbInner` since that won't expose it to potential mutations.

protocol_connect.go
akshayjshah: I think my initial urge to skip sending `{}` was misguided. On balance, it seems better to just send the debug info, whatever it may be; I don't want to rely on end users understanding the semantics of zero values and protojson.

akshayjshah (approved):
Stamping this under the assumption that the only remaining change is to send the debug details, whatever they may be (even if that's `{}`).

akshayjshah (approved):
Conformance failures expected?

jhump:
> Conformance failures expected?

@akshayjshah, sadly, yes. This will be fixed in the next release candidate of conformance tests. I noticed this happening in connect-kotlin, too, and merged a change to relax the timing of this test, to make it less likely to flake: https://github.com/connectrpc/conformance/pull/761/files#r1462481695

Issue #689, opened 2024-02-16
ErrorWriter.IsSupported is overly strict


jhump:
Currently `ErrorWriter.IsSupported` uses the currently supported codecs (provided in the form of handler options) to compute all of the possible content-types that the server would support. However, this is not necessary. There are only four protocol types, and the actual registered codecs aren't necessary in order to classify them:
1. If the content-type is `application/grpc` or starts with `application/grpc+` then it's a gRPC request. 
2. If the content-type is `application/grpc-web` or starts with `application/grpc-web+` then it's a gRPC-Web request. 
3. If the content-type starts with `application/connect+` then it's a Connect streaming request. 
4. Lastly, if the content-type starts with `application/` and matches none of the above OR it has a `Connect-Protocol-Version: 1` header OR it uses a GET HTTP method and includes a `connect=v1` query param then it's a Connect unary request.
5. Only when none of the above apply should the request be un-classifiable and `IsSupported` return false.

In all of the above cases, the actual supported codecs are irrelevant because, once classified as above, an error can be correctly written (error formats are codec-agnostic).

Addressing this would also enable error handling that more closely resembles gRPC. In gRPC servers, if the content-type is not a gRPC content-type (i.e. it is not `application/grpc` and does not start with `application/grpc+`) then the server sends back a 415 HTTP status code, which clients will then classify as an unknown error. If it's a gRPC content-type, but not a supported codec, the server sends back a gRPC error (HTTP status code 200, error code and message in HTTP trailers) with a code of "internal". So with the above logic, when the content-type is not supported, an `ErrorWriter` could be used to write an RPC internal error and a 415 HTTP status code only used if the `ErrorWriter` reports that the request is unsupported.

PR #690, opened 2024-02-16
Reconcile trailers-only and misc error behavior with grpc-go

jhump:
This largely undoes a recent change to do more validation of trailers-only responses (#685), which disallows a body or trailers in what appeared to be a trailers-only response. In that change, a trailers-only response was identified by the presence of a "grpc-status" key in the headers.

In this PR, a trailers-only response is instead defined by the lack of body and trailers (not the presence of a "grpc-status" header). This PR also tweaks some other error scenarios:

* If trailers (or an end-stream message) is completely missing from a response, it's considered an `internal` error. But if trailers are present, but the "grpc-status" key is missing, it's considered an issue determining the status, which is an `unknown` error.
* Similarly, if a response content-type doesn't appear to be the right protocol (like it may have come from a non-RPC server), the error code is now `unknown`. But if it looks like the right protocol but uses the wrong sub-format/codec, it's an `internal` error.
   * Note that in grpc-go, this behavior is also seen in the client, but this PR doesn't attempt to address that in the connect-go client. Instead, that change can be made when #689 is addressed.

This PR also now makes connect-go more strict about the "compressed" flag in a streaming protocol when there was no compression algorithm negotiated. Previously, this library was lenient and did not consider it an error if the message in question was empty (zero bytes). But to correctly adhere to gRPC semantics, it MUST report this case as an `internal `error (see bullet 6 in the [test cases section of the gRPC compression docs](https://github.com/grpc/grpc/blob/master/doc/compression.md#test-cases)).

### DIFF
--- a/connect_ext_test.go
+++ b/connect_ext_test.go
@@ -741,7 +741,7 @@ func TestGRPCMissingTrailersError(t *testing.T) {
 		var connectErr *connect.Error
 		ok := errors.As(err, &connectErr)
 		assert.True(t, ok)
-		assert.Equal(t, connectErr.Code(), connect.CodeInternal)
+		assert.Equal(t, connectErr.Code(), connect.CodeUnknown)
 		assert.True(
 			t,
 			strings.HasSuffix(connectErr.Message(), "protocol error: no Grpc-Status trailer: unexpected EOF"),
@@ -1838,7 +1838,9 @@ func TestUnflushableResponseWriter(t *testing.T) {
 	t.Parallel()
 	assertIsFlusherErr := func(t *testing.T, err error) {
 		t.Helper()
-		assert.NotNil(t, err)
+		if !assert.NotNil(t, err) {
+			return
+		}
 		assert.Equal(t, connect.CodeOf(err), connect.CodeInternal, assert.Sprintf("got %v", err))
 		assert.True(
 			t,
@@ -1875,8 +1877,9 @@ func TestUnflushableResponseWriter(t *testing.T) {
 				assertIsFlusherErr(t, err)
 				return
 			}
-			assert.False(t, stream.Receive())
-			assertIsFlusherErr(t, stream.Err())
+			if assert.False(t, stream.Receive()) {
+				assertIsFlusherErr(t, stream.Err())
+			}
 		})
 	}
 }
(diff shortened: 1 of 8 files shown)


header_test.go
```diff
@@ -52,7 +52,6 @@ func TestHeaderMerge(t *testing.T) {
 	expect := http.Header{
 		"Foo": []string{"one", "two"},
 		"Bar": []string{"one"},
-		"Baz": nil,
```
emcfarlane, line 55: What caused this change?

emcfarlane (approved):

header_test.go
jhump: See the change to `mergeHeaders` above.

And the reason `mergeHeaders` was changed was so that `len(responseTrailers)` tells me the actual number of keys, without having to actually iterate the map and ignore nil/empty keys. This is for deciding if there were any trailers in the request (since a trailers-only response is now defined as "no body, no trailers").

envelope.go
```diff
@@ -222,6 +222,7 @@ func (w *envelopeWriter) write(env *envelope) *Error {
 type envelopeReader struct {
 	ctx             context.Context //nolint:containedctx
 	reader          io.Reader
+	bytesRead       int64
```
akshayjshah, line 225: Mind adding a comment here explaining what this state is for? We're only using it over in `protocol_grpc.go`, which is pretty far away.

akshayjshah, line 225:
```suggestion
	bytesRead       int64  // detect trailers-only gRPC responses
```

connect_ext_test.go
```diff
@@ -2442,97 +2483,6 @@ func TestClientDisconnect(t *testing.T) {
 	})
 }
 
-func TestTrailersOnlyErrors(t *testing.T) {
```
jhump, line 2445: These tests were for the old classification and no longer apply.

For example, they'd check that if we classify a response as trailers-only (due to presence of grpc-status key), then the client complains (in the form of an RPC error with "internal" code) if there is a non-empty body or non-empty HTTP trailers.

But since we now define trailers-only to be a request without body or HTTP trailers, they are moot. In these situations (where a response has a grpc-status key in headers and a non-empty body) we now just ignore the grpc-status key in headers (which is also what grpc-go does).

protocol_grpc.go
```diff
@@ -768,18 +723,29 @@ func grpcHTTPToCode(httpCode int) Code {
 // binary Protobuf format, even if the messages in the request/response stream
 // use a different codec. Consequently, this function needs a Protobuf codec to
 // unmarshal error information in the headers.
+//
+// A non-nil error is only returned when a grpc-status key IS present, but it
+// indicates a code of zero (no error). If no grpc-status key is present, this
+// returns a non-nil *Error that wraps errTrailersWithoutGRPCStatus.
```
akshayjshah, line 729: I'm having a hard time understanding this comment. Should it look like this instead?

```suggestion
// A nil error is only returned when a grpc-status key IS present, but it
// indicates a code of zero (no error). If no grpc-status key is present, this
// returns a non-nil *Error that wraps errTrailersWithoutGRPCStatus.
```

akshayjshah (approved):
Apart from my suggestions on comments, this looks great. I'm _very_ glad that we've invested in a more thorough conformance testing suite!

protocol_grpc.go
jhump: Doh! You're absolutely right. My typing fingers seem to have reversed the intent of my brain 😂

PR #691, opened 2024-02-16
Upgrade to go1.22 and min version go1.20

emcfarlane:
This PR bumps the go version to 1.22 and the min version to go1.20. This was prompted by this CI failure: https://github.com/connectrpc/connect-go/actions/runs/7932731168/job/21659863630 which I believe is related to this issue: https://github.com/golang/go/issues/55846.

Also included is a removal of a workaround for 1.19 header handling introduced here: https://github.com/connectrpc/connect-go/pull/406

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module connectrpc.com/connect
 
-go 1.19
+go 1.20
 
 retract (
 	v1.10.0 // module cache poisoned, use v1.10.1
@@ -9,8 +9,8 @@ retract (
 
 require (
 	github.com/google/go-cmp v0.5.9
-	golang.org/x/net v0.17.0
+	golang.org/x/net v0.21.0
 	google.golang.org/protobuf v1.32.0
 )
 
-require golang.org/x/text v0.13.0 // indirect
+require golang.org/x/text v0.14.0 // indirect
--- a/header.go
+++ b/header.go
@@ -78,10 +78,3 @@ func setHeaderCanonical(h http.Header, key, value string) {
 func delHeaderCanonical(h http.Header, key string) {
 	delete(h, key)
 }
-
-// addHeaderCanonical is a shortcut for Header.Add() which
-// bypasses the CanonicalMIMEHeaderKey operation when we
-// know the key is already in canonical form.
-func addHeaderCanonical(h http.Header, key, value string) {
-	h[key] = append(h[key], value)
-}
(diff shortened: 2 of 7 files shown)


protocol_grpc.go
```diff
@@ -539,27 +539,7 @@ func (hc *grpcHandlerConn) Close(err error) (retErr error) {
...
-		// This block works around a bug in x/net/http2. Until Go 1.20, trailers
-		// written using http.TrailerPrefix were only sent if either (1) there's
-		// data in the body, or (2) the innermost http.ResponseWriter is flushed.
-		// To ensure that we always send a valid gRPC response, even if the user
-		// has wrapped the response writer in net/http middleware that doesn't
-		// implement http.Flusher, we must pre-declare our HTTP trailers. We can
-		// remove this when Go 1.21 ships and we drop support for Go 1.19.
-		for key := range mergedTrailers {
-			addHeaderCanonical(hc.responseWriter.Header(), headerTrailer, key)
-		}
```
jhump, line 552: This is the only part that is safe to remove. You still need the rest, to actually set the trailer values, but will need to use `http.TrailerPrefix` in the key to make it clear they are trailers (since they weren't already advertised in a `Trailer` header).

emcfarlane: I think it's safe to remove all. Trailers are written with the prefix below.

go.mod
```diff
@@ -9,8 +9,8 @@ retract (
 
 require (
 	github.com/google/go-cmp v0.5.9
-	golang.org/x/net v0.17.0
+	golang.org/x/net v0.21.0
```
emcfarlane, line 12: Related to release: https://github.com/golang/net/commit/c63010009c802314a29324ce49987897f9838e29

protocol_grpc.go
jhump: Ah, my bad. I somehow overlooked the `return nil` on line 561

header.go
```diff
@@ -78,10 +78,3 @@ func setHeaderCanonical(h http.Header, key, value string) {
 func delHeaderCanonical(h http.Header, key string) {
 	delete(h, key)
 }
-
-// addHeaderCanonical is a shortcut for Header.Add() which
-// bypasses the CanonicalMIMEHeaderKey operation when we
-// know the key is already in canonical form.
-func addHeaderCanonical(h http.Header, key, value string) {
```
akshayjshah, line 85: Please grep for `addHeaderCanonical` to make sure we remove all references, even in comments. I see one reference in `protocol_grpc.go` that still needs to be cleaned up.

akshayjshah (changes requested):
We also need to update the Windows CI configuration :)

akshayjshah (approved):
LGTM!

PR #698, opened 2024-02-21
Use action vars for go versions

emcfarlane:
Switches go version to use workflow variables. Makes it easier to update the go version in workflows.


emcfarlane:
Should work based on https://github.com/orgs/community/discussions/46785 but somethings borked. Closing for now. Will revisit later.

PR #700, opened 2024-02-23
Fix ErrorWriter IsSupportedCheck with required connect protocol option

### LINKED ISSUE #699: ErrorWriter does not respect connect.WithRequireConnectProtocolHeader option
**Describe the bug**

I'm using the gRPC gateway and want to use the errorWriter only for non-HTTP (connect/gRPC) methods. Unfortunately the `IsSupported()` method yields true in too many cases (i.e. REST client sends `application/json` content type). Thus I cannot differentiate between HTTP and connect/grpc using the error writer.

I hoped the opt `connect.WithRequireConnectProtocolHeader()` would help me with that, but it seems to be not considered.

```go
		connect.NewErrorWriter(connect.WithRequireConnectProtocolHeader())
		if errorWriter.IsSupported(r) {
			_ = errorWriter.Write(w, r, connectErr)
		} else {
			apierrors.HandleHTTPError(r.Context(), w, r, connectErr)
		}
```

**To Reproduce**

See above snippet.

**Environment (please complete the following information):**
- `connect-go` version or commit: `connectrpc.com/connect v1.14.0`
- `go version`: (for example, `go version go1.18.3 darwin/amd64`) v1.22.0

**Additional context**

Slack thread: https://bufbuild.slack.com/archives/CRZ680FUH/p1708557962229309
Related issue: https://github.com/connectrpc/connect-go/issues/689

emcfarlane:
This PR fixes ErrorWriter to correctly return unsupported protocol if the option `WithRequireConnectProtocolHeader` is set and the header or query value isn't include in the request. It will now correctly return unsupported to ensure fallback options can process the error.

Fixes #699

### DIFF
--- a/error_writer.go
+++ b/error_writer.go
@@ -45,6 +45,7 @@ type ErrorWriter struct {
 	grpcWebContentTypes          map[string]struct{}
 	unaryConnectContentTypes     map[string]struct{}
 	streamingConnectContentTypes map[string]struct{}
+	requireConnectProtocolHeader bool
 }
 
 // NewErrorWriter constructs an ErrorWriter. To properly recognize supported
@@ -60,6 +61,7 @@ func NewErrorWriter(opts ...HandlerOption) *ErrorWriter {
 		grpcWebContentTypes:          make(map[string]struct{}),
 		unaryConnectContentTypes:     make(map[string]struct{}),
 		streamingConnectContentTypes: make(map[string]struct{}),
+		requireConnectProtocolHeader: config.RequireConnectProtocolHeader,
 	}
 	for name := range config.Codecs {
 		unary := connectContentTypeFromCodecName(StreamTypeUnary, name)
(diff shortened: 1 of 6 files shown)


error_writer.go
```diff
@@ -87,9 +89,15 @@ func NewErrorWriter(opts ...HandlerOption) *ErrorWriter {
...
 	if _, ok := w.unaryConnectContentTypes[ctype]; ok {
+		if err := connectCheckProtocolVersion(request, w.requireConnectProtocolHeader); err != nil {
+			return unknownProtocol
+		}
 		return connectUnaryProtocol
 	}
 	if _, ok := w.streamingConnectContentTypes[ctype]; ok {
+		if err := connectCheckProtocolVersion(request, w.requireConnectProtocolHeader); err != nil {
+			return unknownProtocol
+		}
```
jhump, line 100: This doesn't feel right. Even if the server config requires a connect protocol version header, the content-type alone makes it clear that this _is_ the Connect streaming protocol. Only unary RPCs have ambiguity, due to use of "application/proto" and "application/json" content-types (which are not necessarily Connect calls).

@akshayjshah, what do you think?

emcfarlane: Ambiguity could arise for a v2 connect spec which we'd want older clients to protect against by failing on the version check. `unknown` is probably not the right error for this case.

jhump: It still wouldn't be ambiguous. If/when we introduce a v2, the version header will _have_ to become required in the spec, which means a missing header _implies_ v1 of the protocol. So we still know the format for the error, regardless of the presence of the header.

The only ambiguity is between Connect unary and non-Connect responses that use the same content types (most notably "application/json").

akshayjshah: @jhump, I think you've nailed it: we should use _different_ header-checking logic here. If the version header is present and indicates v1, use the current Connect streaming encoding. If the version header is missing, also use the current Connect streaming encoding. If the version header is present and _not_ set to v1, we've gotten a request using a version from the future: treat it as an unknown protocol. Regardless, we don't need to worry about whether the RPC server requires the header or not.

As stated in #699 and the , the header is really only useful to distinguish unary Connect requests from non-RPC requests which might use the same Content-Type.

emcfarlane: Set `required = false` for streaming ensuring we still check the version header is correct.

jhump (approved):

PR #701, opened 2024-02-23
Fix ErrorWriter to be codec agnostic

### LINKED ISSUE #689: ErrorWriter.IsSupported is overly strict

emcfarlane:
This PR changes the ErrorWriter to be more lenient with classifying protocols. Errors codecs are agnostic to the codec used. Therefore we avoid checking the codec in classifying the request. IsSupported will return true for an unknown codec which allows the server to encode a better error message to the client. If not supported a 415 error response could be used to match gRPC server like handling. If not supported and trying to write an error the ErrorWriter will default to connects unary encoding (consistent with current behaviour).

Fixes #689

### DIFF
--- a/connect_ext_test.go
+++ b/connect_ext_test.go
@@ -2423,7 +2423,9 @@ func TestClientDisconnect(t *testing.T) {
 			assert.NotNil(t, err)
 			<-gotResponse
 			assert.NotNil(t, handlerReceiveErr)
-			assert.Equal(t, connect.CodeOf(handlerReceiveErr), connect.CodeCanceled)
+			if !assert.Equal(t, connect.CodeOf(handlerReceiveErr), connect.CodeCanceled) {
+				t.Logf("handlerReceiveErr: %v", handlerReceiveErr)
+			}
 			assert.ErrorIs(t, handlerContextErr, context.Canceled)
 		})
 		t.Run("handler_writes", func(t *testing.T) {
@@ -2434,7 +2436,7 @@ func TestClientDisconnect(t *testing.T) {
 				gotResponse       = make(chan struct{})
 			)
 			pingServer := &pluggablePingServer{
-				countUp: func(ctx context.Context, req *connect.Request[pingv1.CountUpRequest], stream *connect.ServerStream[pingv1.CountUpResponse]) error {
+				countUp: func(ctx context.Context, _ *connect.Request[pingv1.CountUpRequest], stream *connect.ServerStream[pingv1.CountUpResponse]) error {
 					close(gotRequest)
 					var err error
 					for err == nil {
--- a/connect_ext_test.go
+++ b/connect_ext_test.go
@@ -2423,9 +2423,7 @@ func TestClientDisconnect(t *testing.T) {
 			assert.NotNil(t, err)
 			<-gotResponse
 			assert.NotNil(t, handlerReceiveErr)
-			if !assert.Equal(t, connect.CodeOf(handlerReceiveErr), connect.CodeCanceled) {
-				t.Logf("handlerReceiveErr: %v", handlerReceiveErr)
-			}
+			assert.Equal(t, connect.CodeOf(handlerReceiveErr), connect.CodeCanceled, assert.Sprintf("got %v", handlerReceiveErr))
 			assert.ErrorIs(t, handlerContextErr, context.Canceled)
 		})
 		t.Run("handler_writes", func(t *testing.T) {
--- a/error_writer.go
+++ b/error_writer.go
@@ -58,8 +58,6 @@ func NewErrorWriter(opts ...HandlerOption) *ErrorWriter {
 	return &ErrorWriter{
 		bufferPool:                   config.BufferPool,
 		protobuf:                     codecs.Protobuf(),
-		handleGRPC:                   config.HandleGRPC,
-		handleGRPCWeb:                config.HandleGRPCWeb,
 		requireConnectProtocolHeader: config.RequireConnectProtocolHeader,
 	}
 }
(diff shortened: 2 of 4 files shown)


connect_ext_test.go
```diff
@@ -2423,7 +2423,9 @@ func TestClientDisconnect(t *testing.T) {
 			assert.NotNil(t, err)
 			<-gotResponse
 			assert.NotNil(t, handlerReceiveErr)
-			assert.Equal(t, connect.CodeOf(handlerReceiveErr), connect.CodeCanceled)
+			if !assert.Equal(t, connect.CodeOf(handlerReceiveErr), connect.CodeCanceled) {
+				t.Logf("handlerReceiveErr: %v", handlerReceiveErr)
```
emcfarlane, line 2427: Unrelated, but noticed in testing. Raising the log to help debug better if seen in CI.

connect_ext_test.go
```diff
@@ -2423,7 +2423,9 @@ func TestClientDisconnect(t *testing.T) {
...
+			}
```
akshayjshah, line 2428: `assert.Sprintf` is for this:

```suggestion
			assert.Equal(
				t,
				connect.CodeOf(handlerReceiveErr),
				connect.CodeCanceled,
				assert.Sprintf("handerReceiveErr: %v", handlerReceiveErr),
			)
```

At some point, we may want to migrate assertions to `testify` or just use `cmp` directly. I like `assert`-style utility libraries, but it's not worth the mental overhead of having a special approach just for this one repository.

error_writer.go
```diff
@@ -54,69 +52,54 @@ type ErrorWriter struct {
 // Options supplied via [WithConditionalHandlerOptions] are ignored.
 func NewErrorWriter(opts ...HandlerOption) *ErrorWriter {
 	config := newHandlerConfig("", StreamTypeUnary, opts)
-	writer := &ErrorWriter{
+	codecs := newReadOnlyCodecs(config.Codecs)
```
akshayjshah, line 55: The GoDoc for `NewErrorWriter` also needs to be updated to explain why we need `...HandlerOption`: now it's not for codecs, it's for the Connect version header.

error_writer.go
```diff
@@ -54,69 +52,54 @@ type ErrorWriter struct {
...
 		}
+		return connectUnaryProtocol
+	case ctype == "":
+		if method != http.MethodGet {
+			break
+		}
+		if err := connectCheckProtocolVersion(request, w.requireConnectProtocolHeader); err != nil {
+			break
+		}
+		return connectUnaryProtocol
```
akshayjshah, line 102: I'm finding the big switch with nested `break`s difficult to read. It'd be clearer to me if we had an `isPost := method == http.MethodPost` before the switch and inlined the HTTP verb checks into the case statements where possible.

error_writer.go
```diff
@@ -54,69 +52,54 @@ type ErrorWriter struct {
...
+	case strings.HasPrefix(ctype, connectStreamingContentTypePrefix):
+		if method != http.MethodPost {
+			break
 		}
-		return connectUnaryProtocol
-	}
-	if _, ok := w.streamingConnectContentTypes[ctype]; ok {
 		if err := connectCheckProtocolVersion(request, w.requireConnectProtocolHeader); err != nil {
-			return unknownProtocol
+			break
```
akshayjshah, line 84: Just want to flag that there's an open discussion about this in #700. Let's make sure we sync this up with whatever's decided in that thread before merging to `main`.

error_writer.go
```diff
@@ -41,86 +41,58 @@ const (
...
-			return connectUnaryProtocol
+	case isPost && strings.HasPrefix(ctype, connectUnaryContentTypePrefix):
+		if err := connectCheckProtocolVersion(request, w.requireConnectProtocolHeader); err != nil {
+			return unknownProtocol
 		}
-		connectVersion = request.URL.Query().Get(connectUnaryConnectQueryParameter)
-		if connectVersion == connectUnaryConnectQueryValue {
-			return connectUnaryProtocol
+		return connectUnaryProtocol
+	case isGet && ctype == "":
```
jhump, line 88: In order to be compatible with versions of connect-go prior to v1.13 ([which includes a fix for the content-type with GET requests](https://github.com/connectrpc/connect-go/pull/644/files#diff-cdd39981f92fb3d0b9c81ce742c64f5574d2aed55233f5fdd21a4a7ee80677f3R1049)), perhaps we should be more lenient here and either disregard the content-type or allow it to have a `connectUnaryContentTypePrefix` prefix.

emcfarlane: Makes sense, dropped the check.

error_writer.go
```diff
@@ -80,12 +80,12 @@ func (w *ErrorWriter) classifyRequest(request *http.Request) protocolType {
 			return unknownProtocol
 		}
 		return connectStreamProtocol
-	case isPost && strings.HasPrefix(ctype, connectUnaryContentTypePrefix):
+	case isPost && strings.HasPrefix(ctype, connectUnaryContentTypePrefix) && !strings.HasPrefix(ctype, grpcContentTypeDefault):
```
emcfarlane, line 83: Should we retain the codec supported is a connect codec here? Even though it is error agnostic it is odd that a request can be classified as a connect request only on the `application/` prefix.

jhump: Sorry, I don't understand the question.

> odd that a request can be classified as a connect request only on the application/ prefix

This wasn't the case before. To what exactly are you alluding? Do you mean the fact that we'll conflate _any_ `application/` prefix to mean a Connect unary content type? Or are you referring to the check you added to omit the gRPC content-types?

emcfarlane: Should we remove the config settings for handling grpc and grpc web?

jhump (approved):

PR #702, opened 2024-02-28
In connect unary protocol, fallback to code based on HTTP status if unable to deserialize it

jhump:
The conformance tests, as of v1.0.0-rc3, currently require behavior that matches the current behavior of connect-go: if the error JSON body in a Connect unary response can be unmarshaled but is missing the code, it defaults to using "unknown" as the code.

However, in the Connect unary format (and not any of the others), we also have a non-200 HTTP status that could be used to determine a potentially better default than "unknown". So this PR changes it so that connect-go uses the HTTP status in this case, instead of always using "unknown". (Also see https://github.com/connectrpc/conformance/issues/810.)

It _also_ updates the JSON unmarshaling of the error body so it accepts a `"code"` property with an invalid value. In this case, the invalid value is discarded, and due to the change described above, the code will instead be computed from the HTTP status. But this lenience allows us to capture the message and details, in the event that the JSON contains these other properties but with (for example) a misspelled code string.

### DIFF
--- a/protocol_connect.go
+++ b/protocol_connect.go
@@ -1255,8 +1255,8 @@ func (e *connectWireError) UnmarshalJSON(data []byte) error {
 	// the other fields from the input JSON.
 	var wireError struct {
 		Code    string               `json:"code"`
-		Message string               `json:"message,omitempty"`
-		Details []*connectWireDetail `json:"details,omitempty"`
+		Message string               `json:"message"`
+		Details []*connectWireDetail `json:"details"`
 	}
 	err := json.Unmarshal(data, &wireError)
 	if err != nil {
--- a/protocol_connect.go
+++ b/protocol_connect.go
@@ -559,6 +559,10 @@ func (cc *connectUnaryClientConn) validateResponse(response *http.Response) *Err
 				errors.New(response.Status),
 			)
 		}
+		if wireErr.Code == 0 {
+			// code not set? default to one implied by HTTP status
+			wireErr.Code = connectHTTPToCode(response.StatusCode)
+		}
 		serverErr := wireErr.asError()
 		if serverErr == nil {
 			return nil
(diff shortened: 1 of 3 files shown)


protocol_connect.go
```diff
@@ -559,6 +559,10 @@ func (cc *connectUnaryClientConn) validateResponse(response *http.Response) *Err
 				errors.New(response.Status),
 			)
 		}
+		if wireErr.Code == 0 {
+			// code not set? default to one implied by HTTP status
+			wireErr.Code = connectHTTPToCode(response.StatusCode)
+		}
```
akshayjshah, line 565: Can you also open a PR to add this nuance to the wire protocol docs? This feels less like a behavioral specification and more like an aspect of the wire protocol itself.

protocol_connect.go
```diff
@@ -1245,6 +1249,26 @@ func (e *connectWireError) asError() *Error {
...
 }
 
+func (e *connectWireError) UnmarshalJSON(data []byte) error {
+	// We want to be lenient if the JSON has an unrecognized or invalid code.
+	// So if that occurs, we leave the code unset but can still de-serialize
+	// the other fields from the input JSON.
+	var wireError struct {
+		Code    string               `json:"code"`
+		Message string               `json:"message,omitempty"`
+		Details []*connectWireDetail `json:"details,omitempty"`
```
akshayjshah, line 1259: Do we need the `omitempty` tags here? I think they're meaningless for unmarshaling.

akshayjshah (approved):
Some small asks, but the approach lgtm.

protocol_connect.go
jhump: Nope, my bad. I just copied these out of the `connectWireError` struct without contemplating relevant tweaks.

protocol_connect.go
jhump: Roger that.

jhump: I was in the process of adding this and then read this part of the existing spec, describing the status code of the _Unary-Response_ production:

> When reading data from the wire, client implementations must use the [HTTP-to-Connect mapping](https://connectrpc.com/docs/protocol/#http-to-error-code) to infer a Connect error code if **Bare-Message** is missing or malformed.

IMO, this falls into this category. So it's arguable that always defaulting to "unknown" (as was previously done) was actually a bug.

jhump: So do you think the existing language needs an update or clarification?

protocol_connect.go
jhump: Done

akshayjshah:
Huh, I suppose that’s clear enough then!

PR #706, opened 2024-03-12
Revise code mappings per latest changes to spec

jhump:
This reconciles the implementation in this repo with the recent adjustments to HTTP<->Connect code mappings (https://github.com/connectrpc/connectrpc.com/pull/130).

The gRPC and Connect mappings from HTTP code to RPC code are now aligned, so I've unified them into a single function.

### DIFF
--- a/handler_ext_test.go
+++ b/handler_ext_test.go
@@ -231,7 +231,7 @@ func TestHandler_ServeHTTP(t *testing.T) {
 		resp, err := client.Do(req)
 		assert.Nil(t, err)
 		defer resp.Body.Close()
-		assert.Equal(t, resp.StatusCode, http.StatusNotFound)
+		assert.Equal(t, resp.StatusCode, http.StatusNotImplemented)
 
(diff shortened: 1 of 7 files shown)


handler_ext_test.go
```diff
@@ -231,7 +231,7 @@ func TestHandler_ServeHTTP(t *testing.T) {
 		resp, err := client.Do(req)
 		assert.Nil(t, err)
 		defer resp.Body.Close()
-		assert.Equal(t, resp.StatusCode, http.StatusNotFound)
+		assert.Equal(t, resp.StatusCode, http.StatusNotImplemented)
```
jhump, line 234: This is one case where this mapping makes a lot more sense: when the server receives an unrecognized content-encoding, it returns an "unimplemented" status error. And 501 makes way more sense than 404. (Ideally it would be 415, but let's take what we can get 😆)

akshayjshah (approved):
Looks great! Somehow I expected this to be a more involved change.

PR #707, opened 2024-03-12
Update to latest protobuf-go

jhump:
This will push our users past [CVE-2024-24786](https://github.com/advisories/GHSA-8r3f-844c-mc37) and also mean that users don't have to think about whether the next release is affected by the CVE when they do `go install connectrpc.com/cmd/protoc-gen-connect-go@v1.16.0`.

However it does mean users may be confronted with the [compatibility issue](https://github.com/golang/protobuf/issues/1596) (which, luckily, is easily fixed by users also updating their dependency for `github.com/golang/protobuf`).

gRPC recently [took this step](https://github.com/grpc/grpc-go/pull/7028), too. And since we're expecting another release soon (this week?), now seems like the right time.


akshayjshah (approved):
Makes sense.

emcfarlane (approved):

PR #711, opened 2024-03-14
Restrict metadata headers in error propagation

emcfarlane:
This PR addresses issues when propagating errors from a client back to a handler. On the client side connect errors will contain all response headers: transport (`Content-Type`, `Content-Length`, etc), protocol and application headers. These could break the transport when trying to re-encode the error or leak sensitive information between services. For any wire errors (errors decoded from a client response) we now disable meta propagation. For other errors we now also restrict the headers propagated.

### DIFF
--- a/header.go
+++ b/header.go
@@ -57,6 +57,46 @@ func mergeHeaders(into, from http.Header) {
 	}
 }
 
+// mergeMetdataHeaders merges the metadata headers from the "from" header into
+// the "into" header. It skips over non metadata headers that should not be
+// propagated from the server to the client.
+func mergeMetadataHeaders(into, from http.Header) {
+	for key, vals := range from {
+		if len(vals) == 0 {
+			// For response trailers, net/http will pre-populate entries
+			// with nil values based on the "Trailer" header. But if there
+			// are no actual values for those keys, we skip them.
+			continue
+		}
+		switch http.CanonicalHeaderKey(key) {
+		case headerContentType,
+			headerContentLength,
+			headerContentEncoding,
+			headerHost,
+			headerUserAgent,
+			headerTrailer,
+			headerDate:
+			// HTTP headers.
+		case connectUnaryHeaderAcceptCompression,
+			connectUnaryTrailerPrefix,
+			connectStreamingHeaderCompression,
+			connectStreamingHeaderAcceptCompression,
+			connectHeaderTimeout,
+			connectHeaderProtocolVersion:
+			// Connect headers.
+		case grpcHeaderCompression,
+			grpcHeaderAcceptCompression,
+			grpcHeaderTimeout,
+			grpcHeaderStatus,
+			grpcHeaderMessage,
+			grpcHeaderDetails:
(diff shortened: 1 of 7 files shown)


header.go
```diff
@@ -57,6 +57,46 @@ func mergeHeaders(into, from http.Header) {
...
+// propagated from the server to the client.
+func mergeMetadataHeaders(into, from http.Header) {
+	for key, vals := range from {
+		if len(vals) == 0 {
+			// For response trailers, net/http will pre-populate entries
+			// with nil values based on the "Trailer" header. But if there
+			// are no actual values for those keys, we skip them.
+			continue
+		}
+		switch http.CanonicalHeaderKey(key) {
```
mattrobenolt, line 71: Is the key already canonicalized at this point? It isn't being canonicalized in the other `mergeHeaders` which would also cause potentially different issues if they weren't all canonical.

Basically, canonicalizing, or even checking when it's a no-op is quite expensive if we know they are in the right format.

emcfarlane: You may choose to encode errors without canonicalization as we export the header field on [`Meta()`](https://pkg.go.dev/connectrpc.com/connect#Error.Meta). So:
```
connectErr.Meta()["content-length"] = []string{"1234"}
```
Would cause issues. This was done as a precaution but it could be removed.

protocol_connect.go
```diff
@@ -534,10 +534,12 @@ func (cc *connectUnaryClientConn) validateResponse(response *http.Response) *Err
 			cc.compressionPools.CommaSeparatedNames(),
 		)
 	}
+	cc.unmarshaler.compressionPool = cc.compressionPools.Get(compression)
 	if response.StatusCode != http.StatusOK {
 		unmarshaler := connectUnaryUnmarshaler{
+			ctx:             cc.unmarshaler.ctx,
```
emcfarlane, line 540: This caused a panic, fixed.

header.go
mattrobenolt: Ah, ok, just making sure this was potentially from user input. I think it's fine given it's only on errors anyways.

protocol_connect.go
```diff
@@ -765,8 +766,8 @@ func (hc *connectUnaryHandlerConn) writeResponseHeader(err error) {
 		header[headerVary] = append(header[headerVary], connectUnaryHeaderAcceptCompression)
 	}
 	if err != nil {
-		if connectErr, ok := asError(err); ok {
-			mergeHeaders(header, connectErr.meta)
+		if connectErr, ok := asError(err); ok && !connectErr.wireErr {
+			mergeMetadataHeaders(header, connectErr.meta)
```
jhump, line 770: Same change is also needed for streaming calls -- line 852, in `(*connectStreamingMarshaler).MarshalEndStream`.

akshayjshah:
Haven't looked at the logic in depth, but we should explain this behavior in the docs (`IsWireError`, `ConnectError.Meta`, or both).

jhump (approved):

PR #726, opened 2024-04-17
Fix multiple header writes for connect unary on error

emcfarlane:
On error a connect unary handler will call `Close(err)` to send the final error. For non-nil errors this will always attempt to encode the header status by calling `WriteHeader`. If the body has already been written this triggers the following log via `http.Server`:
```
http: superfluous response.WriteHeader call from connectrpc.com/connect.(*connectUnaryHandlerConn).Close (protocol_connect.go:743)
```
The common case for this superfluous log is when a message send is interrupted, due to a context cancel or other write error, and the error is then attempting to re-encode the headers and body.

This PR now moves the `wroteBody` check to a `wroteHeader` check on the unmarshaller. When any payload is sent, including a nil field for header only, we now stop attempting to encode any following errors, as it would be superfluous.

### DIFF
--- a/protocol_connect.go
+++ b/protocol_connect.go
@@ -686,7 +686,6 @@ type connectUnaryHandlerConn struct {
 	marshaler       connectUnaryMarshaler
 	unmarshaler     connectUnaryUnmarshaler
 	responseTrailer http.Header
-	wroteBody       bool
 }
 
 func (hc *connectUnaryHandlerConn) Spec() Spec {
@@ -713,7 +712,6 @@ func (hc *connectUnaryHandlerConn) Send(msg any) error {
 	if err := hc.marshaler.Marshal(msg); err != nil {
 		return err
 	}
-	hc.wroteBody = true
 	return nil // must be a literal nil: nil *Error is a non-nil error
 }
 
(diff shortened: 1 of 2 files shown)


jhump (approved):
LGTM

protocol_connect.go
```diff
@@ -757,7 +755,7 @@ func (hc *connectUnaryHandlerConn) getHTTPMethod() string {
 	return hc.request.Method
 }
 
-func (hc *connectUnaryHandlerConn) writeResponseHeader(err error) {
+func (hc *connectUnaryHandlerConn) mergeResponseHeader(err error) {
```
emcfarlane, line 758: Renamed to `mergeResponseHeaders` as this doesn't trigger a write.

protocol_connect.go
```diff
@@ -726,16 +724,16 @@ func (hc *connectUnaryHandlerConn) ResponseTrailer() http.Header {
...
+		hc.mergeResponseHeader(err)
 		// If the handler received a GET request and the resource hasn't changed,
 		// return a 304.
 		if len(hc.peer.Query) > 0 && IsNotModifiedError(err) {
 			hc.responseWriter.WriteHeader(http.StatusNotModified)
 			return hc.request.Body.Close()
 		}
 	}
-	if err == nil {
+	if err == nil || hc.marshaler.wroteHeader {
```
emcfarlane, line 736: We could error here to report to the `Close` that the error wasn't encoded.

akshayjshah (approved):
TY for the quick fix!

PR #731, opened 2024-04-22
Add constructors for streams

### LINKED ISSUE #719: Cannot test correct handling of `serverStream.Send` error
**Describe the bug**

`ServerStream.Send()` can return an error, which should be correctly handled in implementation of Handler.

```go
// ServerStream is the handler's view of a server streaming RPC.
//
// It's constructed as part of [Handler] invocation, but doesn't currently have
// an exported constructor.
type ServerStream[Res any] struct {
	conn StreamingHandlerConn
}
type StreamingHandlerConn interface {
	Spec() Spec
	Peer() Peer

	Receive(any) error
	RequestHeader() http.Header

	Send(any) error
	ResponseHeader() http.Header
	ResponseTrailer() http.Header
}
```

Unfortunately, the lack of a public constructor for `ServerStream` doesn't allow to mock `StreamingHandlerConn`, and test for the correct behavior of the handler. For example, if the handler needs to do some special cleanup or action when `Send` fails, today it's impossible to unit test this

**Additional context**
I'll be happy to contribute back the fix, though it's unclear to me whether we want to make the Handler depends on an interface rather than the `ServerStream` struct

pd93:
Fixes #719

As discussed in #719, it is currently difficult to unit test streams because there is no way to construct one with a custom/mock implementation of `StreamingHandlerConn` (or `StreamingClientConn` for client-side streams). This PR exposes new constructors for the following types:

- `ServerStream`
- `ServerStreamForClient`
- `ClientStream`
- `ClientStreamForClient`
- `BidiStream`
- `BidiStreamForClient`

This allows us to do something along the lines of this (example using `mockgen` and `gomock`):

```go
//go:generate $GOBIN/mockgen -destination=stream_mock.go -package=mypackage connectrpc.com/connect StreamingHandlerConn

ctrl := gomock.NewController(t)
mockStreamingHandlerConn := NewMockStreamingHandlerConn(ctrl)
mockStreamingHandlerConn.EXPECT().Send(
    // My expected stream message...
).Return(nil)

// New constructor allows us to create a ServerStream using our mock
stream := connect.NewServerStream[controllerv3.Requirements](mockStreamingHandlerConn)

// call service with mock stream
err := myservice.MyRPC(context.Background(), req, stream)
```

As requested [here](https://github.com/connectrpc/connect-go/issues/719#issuecomment-2070257102), I have also updated the internal code to use these new constructors. One thing I wasn't sure about is whether or not we'd want to expose any other parameters. There are a few cases internally where a stream is constructed with additional fields (such as a `maybeInitializer`) and not being able to pass one into the constructor leads to code like this:

```go
stream := NewClientStreamForClient[Req, Res](c.newConn(ctx, StreamTypeClient, nil))
stream.initializer = c.config.Initializer
return stream
```

However, my use case doesn't have a need to pass in any other parameters and it keeps the API nice and clean without. I think the style above is fine for internal code. Happy to discuss this if there is a need for it though.

### DIFF
--- a/client.go
+++ b/client.go
@@ -133,12 +133,13 @@ func (c *Client[Req, Res]) CallUnary(ctx context.Context, request *Request[Req])
 // CallClientStream calls a client streaming procedure.
 func (c *Client[Req, Res]) CallClientStream(ctx context.Context) *ClientStreamForClient[Req, Res] {
 	if c.err != nil {
-		return &ClientStreamForClient[Req, Res]{err: c.err}
+		stream := NewClientStreamForClient[Req, Res](nil)
+		stream.err = c.err
+		return stream
 	}
-	return &ClientStreamForClient[Req, Res]{
-		conn:        c.newConn(ctx, StreamTypeClient, nil),
-		initializer: c.config.Initializer,
-	}
+	stream := NewClientStreamForClient[Req, Res](c.newConn(ctx, StreamTypeClient, nil))
+	stream.initializer = c.config.Initializer
+	return stream
 }
 
 // CallServerStream calls a server streaming procedure.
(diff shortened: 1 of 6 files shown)


emcfarlane:
@pd93 thanks for looking into this issue. Closing for now as we need to discuss the API implementation further before looking at an implementation.

PR #741, opened 2024-05-21
Migrate buf configuration files to v2

bufdev:
Done via `buf config migrate`.


buf.yaml
```diff
@@ -0,0 +1,19 @@
+version: v2
+modules:
+  - path: internal/proto
+lint:
+  use:
+    - DEFAULT
+  except:
+    - FIELD_NOT_REQUIRED
+    - PACKAGE_NO_IMPORT_CYCLE
```
akshayjshah, line 9: We didn't need these before. Why do we need them now?

bufdev: We don't - they're excluded as part of `buf config migrate` as they aren't part of the `v1` `DEFAULT` lint rules, but I've removed them and verified we're passing them. We could make `buf config migrate` smarter to test a build without them, but that was too much.

nicksnyder: We could have the migrate command include a comment for each of these that says something to the effect of "This is a new lint rule this is on by default, but is disabled to ensure a backward compatible migration. Recommend removing these exceptions and fixing your build if it breaks"

akshayjshah: Ah, I'd forgotten about that discussion. Mea culpa.

Including a comment in the migration output would be nice.

bufdev: Comments are...difficult in yaml I think, given our setup, but we could look into it

buf.yaml
```diff
@@ -0,0 +1,16 @@
...
+  ignore:
+    - internal/proto/connectext/grpc/health/v1/health.proto
+    - internal/proto/connectext/grpc/reflection/v1alpha/reflection.proto
+    - internal/proto/connectext/grpc/status/v1/status.proto
+  disallow_comment_ignores: true
+breaking:
+  use:
+    - WIRE_JSON
+  except:
+    - FIELD_SAME_DEFAULT
```
akshayjshah, line 16: I think is similar to the lint excludes - it's new in v2, but we should probably enable it. I don't think anyone's really going to care about the test schemas, but we shouldn't change the defaults in health, reflection, or status.

bufdev: Whoops, missed that - updated.

akshayjshah (approved):
Think we should probably remove the breaking change exclusion too, but the change LGTM - it's nice to not need `buf.work.yaml` anymore! Feel free to merge without more review.

PR #748, opened 2024-05-28
Restrict handler metadata headers

emcfarlane:
This PR fixes the setting of protocol headers to avoid multiple value headers when providing metadata to a handler. The metadata headers are further restricted to avoid setting protocol headers like "Content-Type". This restriction allows the user to pass the response of a proxy call to a handler without having to filter the response headers themselves. This enforces the protocol headers are set by the handler and that they are unaffected from any user provided metadata.

Previously, returning the response of a client request to a handler would merge the headers together leading to protocol errors from invalid headers such as "Content-Type" having multiple values.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -132,3 +132,6 @@ issues:
     # We want to show examples with http.Get
     - linters: [noctx]
       path: internal/memhttp/memhttp_test.go
+    # We need to initialize a map of all protocol headers
+    - linters: [gochecknoglobals]
+      path: header.go
--- a/handler.go
+++ b/handler.go
@@ -71,8 +71,8 @@ func NewUnaryHandler[Req, Res any](
 		if err != nil {
 			return err
 		}
-		mergeMetadataHeaders(conn.ResponseHeader(), response.Header())
-		mergeMetadataHeaders(conn.ResponseTrailer(), response.Trailer())
+		mergeNonProtocolHeaders(conn.ResponseHeader(), response.Header())
+		mergeNonProtocolHeaders(conn.ResponseTrailer(), response.Trailer())
 		return conn.Send(response.Any())
 	}
 
--- a/handler.go
+++ b/handler.go
@@ -71,8 +71,8 @@ func NewUnaryHandler[Req, Res any](
 		if err != nil {
 			return err
 		}
-		mergeHeaders(conn.ResponseHeader(), response.Header())
-		mergeHeaders(conn.ResponseTrailer(), response.Trailer())
+		mergeMetadataHeaders(conn.ResponseHeader(), response.Header())
+		mergeMetadataHeaders(conn.ResponseTrailer(), response.Trailer())
 		return conn.Send(response.Any())
 	}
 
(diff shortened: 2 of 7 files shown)


header.go
```diff
@@ -124,3 +102,32 @@ func setHeaderCanonical(h http.Header, key, value string) {
 func delHeaderCanonical(h http.Header, key string) {
 	delete(h, key)
 }
+
+func isProtocolHeader(key string) bool {
```
jhump, line 106: FYI, this list is long enough that it may be worth benchmarking this approach vs a package map var. So instead of needing ~20 string comparisons before deciding a key is _not_ a protocol header, it could be done with a single map query.

handler.go
```diff
@@ -71,8 +71,8 @@ func NewUnaryHandler[Req, Res any](
 		if err != nil {
 			return err
 		}
-		mergeHeaders(conn.ResponseHeader(), response.Header())
-		mergeHeaders(conn.ResponseTrailer(), response.Trailer())
+		mergeMetadataHeaders(conn.ResponseHeader(), response.Header())
+		mergeMetadataHeaders(conn.ResponseTrailer(), response.Trailer())
```
jhump, line 75: This feels a bit brittle. What is the different between `mergeHeaders` and `mergeMetadataHeaders`? Maybe they should be renamed so the difference and their purpose is more clear? The doc comment added to `mergeHeaders` isn't very helpful.

header.go
emcfarlane: Made it a map.

handler.go
emcfarlane: Used `mergeNonProtocolHeaders` as an alt.

.golangci.yml
```diff
@@ -132,3 +132,6 @@ issues:
     # We want to show examples with http.Get
     - linters: [noctx]
       path: internal/memhttp/memhttp_test.go
+    # We need to initialize a map of all protocol headers
+    - linters: [gochecknoglobals]
+      path: header.go
```
jhump, line 137: For a single violation, seems _way_ better to put this at the actual point of use via `//nolint:gochecknoglobals`.

header.go
```diff
@@ -57,41 +84,17 @@ func mergeHeaders(into, from http.Header) {
 	}
 }
 
-// mergeMetdataHeaders merges the metadata headers from the "from" header into
-// the "into" header. It skips over non metadata headers that should not be
-// propagated from the server to the client.
-func mergeMetadataHeaders(into, from http.Header) {
+// mergeNonProtocolHeaders merges headers excluding protocol headers defined in
+// protocolHeaders.
+func mergeNonProtocolHeaders(into, from http.Header) {
```
jhump, line 89: How about if we also rename the other one `mergeAllHeaders` and add the doc comment back to it, but clarifying that it adds _all_ header values, including protocol headers?

jhump (approved):
Couple of minor suggestions. Otherwise LGTM

header.go
emcfarlane: I think `mergeHeaders` implies all.

.golangci.yml
emcfarlane: Will create a PR to convert over the others too, current convention is to add them here.

PR #756, opened 2024-06-25
Add protocol helpers to infer procedure and type

emcfarlane:
Two new methods are added to allow for inferring the procedure and protocol type of a request. These are provided to be used with http middleware. For example, authentication middleware may wish to block on certain protocols or to conditionally allow routes. These new methods pair well with the [authn-go](https://github.com/connectrpc/authn-go) middleware.


protocol.go
```diff
@@ -48,6 +48,45 @@ const (
 
 var errNoTimeout = errors.New("no timeout")
 
+// InferProtocolFromRequest returns the inferred protocol name for parsing an
+// HTTP request. It inspects the request's method and headers to determine the
+// protocol. If the request doesn't match any known protocol, an empty string
+// is returned.
+func InferProtocolFromRequest(request *http.Request) string {
```
bufdev, line 55: - The rest of the API uses For and not From
- would this be better named "ProtocolForRequest"?
- would it be better to return string, bool?

emcfarlane: The `For` is used to distinguish streams that are to be consumed by the client rather than the handler. So I think the `From`  is better here even tho its not currently in the API.

jhump: In my mind they are not necessarily synonyms where you'd always use one over another for consistency.
In my mind, the distinction is that `For` in a function name is used when the function constructs a new value that corresponds to its input or even wraps its input; `From` is used when the function extracts a value out of the inputs. In that case, `From` seems appropriate to me.

As far as returning `string` vs. `(string, bool)`: another option is to add a `ProtocolUnknown` constant, so that callers can use a switch and don't need to examine two values. If that approach sounds good, the constant value might need to be the empty string so that an uninitialized `Peer` object reports `ProtocolUnknown`. WDYT?

protocol.go
```diff
@@ -48,6 +48,45 @@ const (
...
+		return "", false
+	default:
+		return "", false
+	}
+}
+
+// ProcedureFromURL returns the inferred procedure name from a URL. It's
+// returned in the form "/service/method" if a valid suffix is found. If the
+// path doesn't contain a service and method, the entire path is returned.
+func ProcedureFromURL(url *url.URL) (string, bool) {
```
jhump, line 73: Why isn't this method used anywhere? The code that extracts this to report in the `Spec` should use this, no?

protocol.go
```diff
@@ -48,6 +48,45 @@ const (
...
+	ultimate := strings.LastIndex(path, "/")
+	if ultimate < 0 {
+		return url.Path, false
+	}
+	penultimate := strings.LastIndex(path[:ultimate], "/")
+	if penultimate < 0 {
+		return url.Path, false
+	}
+	procedure := path[penultimate:]
+	if len(procedure) < 4 { // two slashes + service + method
```
jhump, line 84: This isn't really adequate as it allows both `/foo/` and `//foo`. We should probably instead be checking that`ultimate < len(path)-1` and `penultimate < ultimate-1`.

protocol.go
emcfarlane: The handler returns the procedure name to be passed to a router and the generated code can then hardcode the method name.

protocol.go
```diff
@@ -81,10 +82,11 @@ func ProcedureFromURL(url *url.URL) (string, bool) {
 		return url.Path, false
 	}
 	procedure := path[penultimate:]
-	if len(procedure) < 4 { // two slashes + service + method
+	// Ensure that the service and method are non-empty.
+	if ultimate == len(path)-1 || penultimate == ultimate-1 {
 		return url.Path, false
 	}
-	return procedure, true
+	return procedure, false
```
jhump, line 89: This change looks like an accident. This was the one case to return true, right?

protocol_test.go
```diff
@@ -68,19 +68,23 @@ func BenchmarkCanonicalizeContentType(b *testing.B) {
 func TestProcedureFromURL(t *testing.T) {
 	t.Parallel()
 	tests := []struct {
-		name string
-		url  string
-		want string
+		name  string
+		url   string
+		want  string
+		valid bool
```
jhump, line 74: Need an assertion for this (which would have caught what I assume is a bug pointed out by other comment)

protocol.go
```diff
@@ -48,6 +48,47 @@ const (
 
 var errNoTimeout = errors.New("no timeout")
 
+// ProtocolFromRequest returns the inferred protocol name for parsing an
+// HTTP request. It inspects the request's method and headers to determine the
+// protocol. If the request doesn't match any known protocol, an empty string
+// and false is returned.
+func ProtocolFromRequest(request *http.Request) (string, bool) {
```
jhump, line 55: I've been thinking about if we'd want to ever return something more structured/richer -- like a `Protocol` type that can inform the caller if it's connect steram vs. connect unary or even the version (in the event we ever add a v2 of the protocol).

I suppose we can leave this as is. In `Spec`, we'd have to add another accessor -- like maybe `ProtocolDetails`, and maybe we could then deprecate this and add `ProtocolDetailsFromRequest` at that time. WDYT?

protocol_test.go
emcfarlane: Sorry, not sure how this slipped up. I think I was adding this testcase and missed the commit. Have fixed.

protocol_test.go
```diff
@@ -63,3 +66,113 @@ func BenchmarkCanonicalizeContentType(b *testing.B) {
...
+		contentType: "application/grpc-web",
+		method:      http.MethodGet,
+		want:        ProtocolConnect,
+		valid:       true,
+	}, {
+		name:        "grpcGet",
+		contentType: "application/grpc+json",
+		method:      http.MethodGet,
+		want:        ProtocolConnect,
+		valid:       true,
```
emcfarlane, line 119: This is an odd result and maybe should be revisited to defaulting to `unknown`. We allow any `application/` prefix to fall-through to be classified as a connect request.

jhump: I don't think that's right. I've got an issue in gRPC with some spec issues/clarifications (https://github.com/grpc/grpc/issues/36767), and the discussion there has brought up possible experimental support for other verbs, like GET, in gRPC.

If a content-type is set, we should likely use it to ascertain the protocol and not look at the HTTP method. If there is no content-type, then we'd see if it's GET and if so also check for the presence of required `message` and `encoding` query params. Otherwise, it's unknown. How does that sound?

protocol.go
jhump: > My preference would be to move this back to authn-go until we comeup with a better implementation.

It would be nice to have some canonical implementation that we can also use from vanguard. Admittedly, in the past, we (with @akshayjshah) discussed a possible `connectrpc/protocol-go` repo where we could put lower-level details of the protocol like this, and `connectrpc/connect-go` (et al) could then import that. I'm a little concerned with putting it in authn-go if we expect to move it out later. But I guess it's okay as long as we have this figured out before we get to a v1.0 of authn-go 🤷.

protocol_test.go
emcfarlane: Sounds good! Now check for query parameters for GET requests. Removed `Content-Type` from all GET requests as that should not be set for any valid request. Still enforce http method for protocol checks as I do think the spec currently requires it, so these are now otherwise unknown.

emcfarlane:
Closing, talked offline. Can move these helpers to authn-go for now.

PR #775, opened 2024-09-13
Fix error message about content-types

jhump:
The error message had a copy+pasta bug where it was showing the wrong prefix. This resulted in possibly very confusing error message where it complained that the content-type was wrong, but then said it was expecting exactly what it actually got.

### DIFF
--- a/duplex_http_call.go
+++ b/duplex_http_call.go
@@ -168,6 +168,8 @@ func (d *duplexHTTPCall) CloseWrite() error {
 	// response to read from.
 	if d.requestSent.CompareAndSwap(false, true) {
 		go d.makeRequest()
+		// We never setup a request body, so it's effectively already closed.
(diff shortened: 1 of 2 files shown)


duplex_http_call.go
```diff
@@ -168,6 +168,8 @@ func (d *duplexHTTPCall) CloseWrite() error {
 	// response to read from.
 	if d.requestSent.CompareAndSwap(false, true) {
 		go d.makeRequest()
+		// We never setup a request body, so it's effectively already closed.
+		// So nothing else to do.
```
jhump, line 172: Unrelated, but I was reading this code and confused at first why it does nothing to close the body. And it's because we haven't set one yet: if `d.requestSent` is false then the request body is still `http.NoBody`.

akshayjshah (approved):

PR #776, opened 2024-09-13
Connect unary should return unavailable instead of unimplemented for `io.EOF` errors from the transport

jhump:
This adds a test that could reproduce the issue where unary RPCs could return an `unimplemented` code when an `unavailable` code was more appropriate. Basically, it was mistakenly interpreting network EOF errors as "the server did not send us any response messages which is cardinality violation" instead of "the server connection broke so the server response is unavailable".

There are still cases where an `io.EOF` could likely be misinterpreted as a cardinality issue, if it the connection is severed after the headers have been received but before the first byte of the first envelope has been received. But the main case where this issue has been reported was for unary RPCs in the Connect protocol, which does not suffer from that issue (since it does not use envelopes).

So this isn't 100% perfect, but it gets us closer and likely fixes most of the issues in practice.


.golangci.yml
```diff
@@ -19,8 +19,6 @@ linters-settings:
   importas:
     no-unaliased: true
     alias:
-      - pkg: connectrpc.com/connect
-        alias: connect
```
jhump, line 23: Since v1.11.0 (the move to the connectrpc.com domain), it is no longer necessary/helpful to always have an import alias here. So I've removed this requirement and removed the (superfluous) alias in the file I touched below.

duplex_http_call.go
```diff
@@ -306,6 +306,11 @@ func (d *duplexHTTPCall) makeRequest() {
 	// pipe. Write's check for io.ErrClosedPipe and will convert this to io.EOF.
 	response, err := d.httpClient.Do(d.request) //nolint:bodyclose
 	if err != nil {
+		if errors.Is(err, io.EOF) {
+			// We use io.EOF as a sentinel in many places and don't want this
+			// transport error to be confused for those other situations.
+			err = io.ErrUnexpectedEOF
+		}
```
emcfarlane, line 313: Could be moved as a `wrapIf...` for consistency?

client_ext_test.go
```diff
@@ -227,6 +227,58 @@ func TestGetNoContentHeaders(t *testing.T) {
 	assert.Equal(t, http.MethodGet, unaryReq.HTTPMethod())
 }
 
+func TestConnectionDropped(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
+	for _, protocol := range []string{connect.ProtocolConnect, connect.ProtocolGRPC, connect.ProtocolGRPCWeb} {
```
emcfarlane, line 233: The protocol doesn't affect the test case, maybe that too implementation aware but this looks like it can be flattened to the one testcase.

emcfarlane (approved):

duplex_http_call.go
jhump: But we're not wrapping -- we are replacing.

client_ext_test.go
jhump: I don't understand this remark. What one test case would I flatten it to?

It does, in fact, effect the test case: we use this to build up client options, to select the protocol, and then pass those options below (line 249).

duplex_http_call.go
emcfarlane: The error could be wrapped and this avoids downstream errors mutating it. All these errors get wrapped by `NewError(CodeUnavailable, err)`

client_ext_test.go
emcfarlane: Setting the protocols doesn't seem to affect what the test does, but this is an implementation detail. So this testcase might be better suited as an internal test on duplexHTTPCall that `makeRequest` returns the correct error response.

jhump: > doesn't seem to affect what the test does

I still don't understand what you mean. They cause the client to change the protocol it uses, to make sure that the error is handled correctly through all protocol implementations, for both unary and streaming RPCs.

I definitely don't want it to just be a test of `makeRequest`, because somewhere further up the stack could be changed in a way that breaks this expectation. We want to assert that the "unavailable" error makes it all the way up to the RPC calling code. Previously, the "unavailable" error was getting _changed_ to "unimplemented" (and that happens _outside_ of `makeRequest`).

duplex_http_call.go
jhump: > All these errors get wrapped by `NewError(CodeUnavailable, err)`

I'm not really sure what you're saying. At this point, the `err` is directly from the `HTTPClient.Do` method -- no other Connect wrapping has occurred.

emcfarlane (approved):

emcfarlane:
@jhump comments were only nits, i've resolved.

client_ext_test.go
emcfarlane: Okay, this makes sense. Better as is.

duplex_http_call.go
emcfarlane: I'm referring to the error logic that happens below. Each wrap function checks the error for a connect error, then if not a connect error attempts to match it. Setting it as a connect error upfront would avoid any other error wrapper from wrapping it as a different error.

emcfarlane: For example `wrapIfContextError` will check with `asError(err)` and then return early before attempting to wrap as a different error. https://github.com/connectrpc/connect-go/blob/99d6b9ceaf98d18c295f66af16acde0e639caca1/error.go#L302-L304

jhump:
> @jhump comments were only nits, i've resolved.

Thanks. But I went ahead and pushed an alternative (#777) just to see if you like it better. Let me know.

PR #795, opened 2024-11-02
Add docs for StreamingHandlerConn concurrency constraints

emcfarlane:
Adds docs for the StreamingHandlerConn interface stating the constraints for concurrent use. The grouped methods are similar to the clients StreamingClientConn interface.

Edit: moving to draft to clarify behaviour.


emcfarlane:
Handler implementations broadly follow these limited concurrency guides but it is a breaking change to add these constraints. Closing, the server implementations don't support any concurrent use cases.

PR #799, opened 2024-11-28
Reproduce issue with client error handling + fix

jamesrom:
~~One line fix is pretty self-explanatory.~~ Update: it's still one line, but not-so self-explanatory 😅

If a client interceptor hides an error, [this type assertion](https://github.com/connectrpc/connect-go/blob/74a6754f29185b85fefa2915bf8fb680a36ca8f0/client.go#L116) will fail. This is because type information is [lost when unaryFunc returns nil](https://github.com/connectrpc/connect-go/blob/74a6754f29185b85fefa2915bf8fb680a36ca8f0/client.go#L98).

I have provided a test case that reproduces the issue.

### DIFF
--- a/client.go
+++ b/client.go
@@ -95,7 +95,7 @@ func NewClient[Req, Res any](httpClient HTTPClient, url string, options ...Clien
 		response, err := receiveUnaryResponse[Res](conn, config.Initializer)
 		if err != nil {
 			_ = conn.CloseResponse()
-			return response, err
+			return nil, err
 		}
 		return response, conn.CloseResponse()
 	})
@@ -110,7 +110,7 @@ func NewClient[Req, Res any](httpClient HTTPClient, url string, options ...Clien
 		request.peer = client.protocolClient.Peer()
 		protocolClient.WriteRequestHeader(StreamTypeUnary, request.Header())
 		response, err := unaryFunc(ctx, request)
-		if err != nil {
+		if err != nil || response == nil {
 			return nil, err
 		}
 		typed, ok := response.(*Response[Res])
--- a/client.go
+++ b/client.go
@@ -95,7 +95,7 @@ func NewClient[Req, Res any](httpClient HTTPClient, url string, options ...Clien
 		response, err := receiveUnaryResponse[Res](conn, config.Initializer)
 		if err != nil {
 			_ = conn.CloseResponse()
-			return nil, err
+			return response, err
 		}
 		return response, conn.CloseResponse()
 	})
--- a/client_error_handling_test.go
+++ b/client_error_handling_test.go
@@ -1,0 +1,76 @@
+// Copyright 2021-2024 The Connect Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package connect
+
+import (
+	"context"
+	"errors"
+	"net/http"
+	"testing"
+
+	"connectrpc.com/connect/internal/assert"
+	pingv1 "connectrpc.com/connect/internal/gen/connect/ping/v1"
+	"connectrpc.com/connect/internal/memhttp/memhttptest"
+)
+
+var ErrOKToIgnore = errors.New("some error that is ok to ignore")
+
+func TestClientErrorHandling(t *testing.T) {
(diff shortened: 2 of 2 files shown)


client.go
```diff
@@ -95,7 +95,7 @@ func NewClient[Req, Res any](httpClient HTTPClient, url string, options ...Clien
 		response, err := receiveUnaryResponse[Res](conn, config.Initializer)
 		if err != nil {
 			_ = conn.CloseResponse()
-			return nil, err
+			return response, err
```
mattrobenolt, line 98: I guess I could argue that every returning a nil, it should be a properly typed nil?

So maybe something like:

```go
var zero *Res
return zero, err
```

or

```go
return (*Res)(nil), err
```

jamesrom: If you'd like to guarantee that the response is empty it would be:
```go
return NewResponse[Res](nil), err
```
or more simply:
```go
return &Response[Res]{}, err
```

Which is fine by me, not here to argue.

jamesrom: Alternatively, [this err != nil check](https://github.com/connectrpc/connect-go/blob/74a6754f29185b85fefa2915bf8fb680a36ca8f0/client.go#L113) could be updated:

```go
if err != nil || response == nil {
	return nil, err
}
```
Let me know which way you'd like to go -- this might be the preferred approach because it keeps back-compat: there's no change to what interceptors receive for an error response.

client.go
```diff
@@ -110,7 +110,7 @@ func NewClient[Req, Res any](httpClient HTTPClient, url string, options ...Clien
 		request.peer = client.protocolClient.Peer()
 		protocolClient.WriteRequestHeader(StreamTypeUnary, request.Header())
 		response, err := unaryFunc(ctx, request)
-		if err != nil {
+		if err != nil || response == nil {
```
jhump, line 113: This doesn't seem right. A `nil, nil` response isn't really valid. What is the framework expected to do with this when returning a value to the caller?

client_error_handling_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+func (e errorHidingInterceptor) WrapStreamingHandler(StreamingHandlerFunc) StreamingHandlerFunc {
+	panic("unimplemented")
+}
+
+func (e errorHidingInterceptor) WrapUnary(next UnaryFunc) UnaryFunc {
+	return func(ctx context.Context, req AnyRequest) (_ AnyResponse, retErr error) {
+		res, err := next(ctx, req)
+		if CodeOf(err) == CodeCanceled { // some criteria for ignored errors
+			return res, nil
```
jhump, line 72: This causes the interceptor to return `nil, nil`. I don't think that should be legal. If an interceptor is "hiding" an error, it must synthesize an actual, non-nil response value.

This could be done dynamically by inspecting `req.Spec.Schema()`. If that is unset then a dynamic response is not possible.

We could also potentially have connect-go's internal implementation of `AnyRequest` implement an optional method that could help. For example, maybe something like so:
```go
if respTyped, ok := req.(interface {
    // Could instantiate the correct response type and then run
    // any configured initializer.
    NewResponse() any
}); ok {
    return respTyped.NewResponse(), nil
}
```
But, now that I've suggested it, I don't love that. It is much less intuitive of course. But it also just feels like a terrible pattern to support -- always returning a zero value response seems wrong and is almost certain to violate any expected RPC contract about what the method is _supposed_ to return. This just doesn't seem like an appropriate function for an interceptor.

It would be much better to inject this error handling much deeper -- like into the actual response handlers, so they can return a valid response value if the application wants to ignore a particular error. Another possibility would be to use a switch on the schema type or the actual procedure name so that it always returns a valid response (which may not necessarily be empty).

jamesrom: The client interceptor returning an actual, non-nil response value is reasonable. However, it also doesn't seem right that interceptors lose all information about the expected response in an error condition.

I don't think smuggling the type via AnyRequest would work either -- there's no reason (other than convention, I suppose) that the same request type isn't used by multiple service methods returning different response types.

Short of smuggling the type information through the Error struct, another option is that `resp.Any()` could do a nil check:

```go
// Any returns the concrete response message as an empty interface, so that
// *Response implements the [AnyResponse] interface. If the response message is
// nil, Any returns a new zero value of the message type.
func (r *Response[T]) Any() any {
	if r.Msg == nil {
		return new(T)
	}
	return r.Msg
}
```

But this means the framework would need to ensure that resp is never nil before being passed to the next interceptor.

client.go
jamesrom: See https://github.com/connectrpc/connect-go/pull/799#discussion_r1868507259

emcfarlane:
@jamesrom clarified the behaviour in https://github.com/connectrpc/connect-go/pull/912. Please let me know if you still have issues. I don't think the check `err != nil || response == nil` is valid: a response is only nil when `err != nil` and the response _must_ not be nil when the `err == nil`.

emcfarlane:
Thanks for the PR. Closing in favor of documentation solution in https://github.com/connectrpc/connect-go/pull/912.

Issue #789, opened 2024-10-15
ServerStreamForClient.Close hangs forever


amenzhinsky:
**Describe the bug**

A client service needs to forcibly stop a server stream, but the Close method hangs indefinitely, preventing the connection-handling function from ever returning.

```
goroutine 82 [sync.Cond.Wait, 42 minutes]:
sync.runtime_notifyListWait(0xc046ba84c8, 0x65)
	runtime/sema.go:569 +0x159
sync.(*Cond).Wait(0x1a?)
	sync/cond.go:70 +0x85
golang.org/x/net/http2.(*pipe).Read(0xc046ba84b0, {0xc047066000, 0x2000, 0x2000})
	golang.org/x/net@v0.29.0/http2/pipe.go:76 +0xdf
golang.org/x/net/http2.transportResponseBody.Read({0x1?}, {0xc047066000?, 0x1?, 0x0?})
	golang.org/x/net@v0.29.0/http2/transport.go:2637 +0x65
io.(*LimitedReader).Read(0xc0435c5d40, {0xc047066000?, 0x10?, 0x1?})
	io/io.go:479 +0x43
io.discard.ReadFrom({}, {0x12b2d80, 0xc0435c5d40})
	io/io.go:666 +0x6d
io.copyBuffer({0x12b4620, 0x1b89ae0}, {0x12b2d80, 0xc0435c5d40}, {0x0, 0x0, 0x0})
	io/io.go:415 +0x151
io.Copy(...)
	io/io.go:388
connectrpc.com/connect.discard({0x7d06ab099df0, 0xc046ba8480})
	connectrpc.com/connect@v1.16.2/protocol.go:296 +0xc5
connectrpc.com/connect.(*duplexHTTPCall).CloseRead(0xc0435c0d20)
	connectrpc.com/connect@v1.16.2/duplex_http_call.go:241 +0x6d
connectrpc.com/connect.(*connectStreamingClientConn).CloseResponse(0xc046f9a770?)
	connectrpc.com/connect@v1.16.2/protocol_connect.go:647 +0x17
connectrpc.com/connect.(*errorTranslatingClientConn).CloseResponse(0xc0435c4c18)
	connectrpc.com/connect@v1.16.2/protocol.go:218 +0x23
connectrpc.com/otelconnect.(*streamingClientInterceptor).CloseResponse(0xc0439fff50)
	connectrpc.com/otelconnect@v0.7.1/payloadinterceptor.go:38 +0x23
[...]

PR #801, opened 2024-12-04
test: cancelling a stream

oliverpool:
Related to #789 and #791.

According to @jhump in https://github.com/connectrpc/connect-go/issues/789#issuecomment-2413899375
> If you want to abort the RPC before the server's response has been received, one must cancel the context.Context that was used to create the RPC. Cancelling an RPC this way will trigger a cancellation error in the server.

However when http2 is enabled, cancelling the context does not suffice to unblock `stream.Close()`.

The `stream.Close` calltree indicates that it is blocked in `discard(reader io.Reader)` (add a `-timeout 500ms` to `go test` to get a dump).

When http2 is not enabled, it seems to behave correctly.

#791 would fix this.

### DIFF
--- a/connect_ext_close_test.go
+++ b/connect_ext_close_test.go
@@ -13,6 +13,7 @@ import (
 	"connectrpc.com/connect/internal/gen/connect/ping/v1/pingv1connect"
 )
 
+// See https://github.com/connectrpc/connect-go/pull/801 for further context
 func TestClientStream_CancelContext(t *testing.T) {
 	t.Run("HTTP2 disabled", func(t *testing.T) {
 		t.Parallel()
@@ -81,10 +82,7 @@ func testClientStream_CancelContext(t *testing.T, enableHTTP2 bool) {
 		t.Error("stream was not done receiving within 1s")
 	}
 
-	// The connection appears to not be properly closed with http2_enabled:
-	// The following line takes 10 seconds and outputs:
-	// httptest.Server blocked in Close after 5 seconds, waiting for connections:
-	//  *tls.Conn 0x0000 127.0.0.1:**** in state active
+	// ensure the connection gets closed quickly
 	startClosing := time.Now()
 	s.Close()
 	assert.True(t, time.Since(startClosing) < time.Second, assert.Sprintf("server.Close took too long: %s", time.Since(startClosing)))
--- a/connect_ext_close_test.go
+++ b/connect_ext_close_test.go
@@ -15,9 +15,11 @@ import (
 
 func TestClientStream_CancelContext(t *testing.T) {
 	t.Run("HTTP2 disabled", func(t *testing.T) {
+		t.Parallel()
 		testClientStream_CancelContext(t, false)
 	})
 	t.Run("HTTP2 enabled", func(t *testing.T) {
+		t.Parallel()
 		testClientStream_CancelContext(t, true)
 	})
 }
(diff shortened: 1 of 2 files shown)


jhump:
Hmmm, this seems like a bug in the HTTP/2 support in `net/http`. Canceling the context associated with the `http.Request` should abort the operation via a HTTP/2 `RST_STREAM` frame. I would expect the response body reader (that is being drained) to abort when that happens.

I did a quick search and was unable to find any existing issues in [github.com/golang/go](https://github.com/golang/go/issues) that describe this, so I may need to play around a little more and file such a bug (if I can convince myself this is a bug in the `net/http` runtime and not in connect-go).

oliverpool:
The rebase confirms that #791 solves the initial issue.

The only remaining issue is the following:
- create an http2 enabled server
- send a stream
- close the stream (it now gets timely closed :heavy_check_mark: )
- close the server (the server hangs for 10s :bug: )

If http2 is disabled, the bug does not appear.

Might be related to https://thrawn01.org/posts/a-golang-http-server-refused-to-shutdown

(from a quick look, the httptest.Server does not configure shutdown. I need to take a deeper look, but don't have time right now)

jhump:
@oliverpool, I just pulled down this branch and see the opposite: HTTP/2 works just fine but it's when HTTP/2 is _disabled_ that it hangs.

I suspect the reason is that HTTP 1.1 does not have a way for the server to tell the client to "go away", other than to add a "Connection: Close" response header to a response. So I think the way HTTP 1.1 servers shutdown is to await a request on an idle connection and then add that header to the response (or add it to responses for all requests that are in-progress at the time the shutdown call is made). And if there are no incoming requests, then they eventually close the connection after an idle timeout.

I was able to get it to pass with a few changes:
1. Set the server config's `WriteTimeout` to one second. Without this, it takes longer for the server handler to get an I/O timeout error from calls to `stream.Send`.
2. Increase the close threshold from one second to three seconds. Even though the write timeout is only one second, it still seems to take some time for the server to acknowledge the error. There must be some level of buffering, though I'm not sure where (connect-go doesn't buffer: `stream.Send` from the server will directly write the data via the `net/http.ResponseWriter`). 🤷

With the above, the test passes with both HTTP 1.1 and HTTP/2.

oliverpool:
> HTTP/2 works just fine but it's when HTTP/2 is disabled that it hangs.

You are right! I oversaw this somehow.

> I suspect the reason is that HTTP 1.1 does not have a way for the server to tell the client to "go away"

But here the problem is that the client wants to tell the server to stop. The following reproducer indicates that this issue present in the stdlib:
```go
func TestHang(t *testing.T) {
	s := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		rc := http.NewResponseController(w)
		n := byte(0)
		for {
			if n > 10 {
				return
			}
			t.Log("write", n)
			_, err := w.Write([]byte{n})
			if err != nil {
				t.Log("bye", err)
				return
			}
			rc.Flush() // needed otherwise the request hangs forever
			
			time.Sleep(time.Second)
			n++
		}
	}))

	ctx, cancel := context.WithCancel(context.Background())
	req, err := http.NewRequestWithContext(ctx, "GET", s.URL, nil)
	assert.Nil(t, err)

	go func() {
		rsp, err := s.Client().Do(req)
		assert.Nil(t, err)
		t.Log(io.Copy(os.Stdout, rsp.Body))
	}()

	time.Sleep(10 * time.Millisecond)
	cancel()
	t.Log("cancelled")

	startClosing := time.Now()
	s.Close()
	assert.True(t, time.Since(startClosing) < time.Second, assert.Sprintf("server.Close took too long: %s", time.Since(startClosing)))
}
```

oliverpool:
I found a solution: the server must take into account the context (`net/http.ResponseWriter` succeeds even if the context is cancelled): c3fdd5a89976165c8bdf942e09a1ad0e828a8b85

jhump:
I think between the approach outlined in c3fdd5a and the fix in #791, this is resolved.

PR #803, opened 2024-12-18
Add same_package option to protoc-gen-connect-go

bufdev:
This adds an option `same_package` to `protoc-gen-connect-go`. If specified, the connect-go types will be generated to the same package as the base types.

Per https://github.com/connectrpc/connect-go/discussions/310#discussioncomment-4089062, I'm not 100% confident in this design - I'd be open to other naming patterns for this option, including making it "enum style".


cmd/protoc-gen-connect-go/main.go
```diff
@@ -89,39 +90,47 @@ func main() {
...
 			string(file.GoImportPath),
 			string(file.GoPackageName),
-		)),
+		))
+	}
+	generatedFile := plugin.NewGeneratedFile(
+		file.GeneratedFilenamePrefix+generatedFilenameExtension,
+		goImportPath,
 	)
 	generatedFile.Import(file.GoImportPath)
```
jhump, line 135: Don't we only want to do this when `!samePackage`?

I think we need to come up with a testing strategy. I need to look at the repo a little more to see how existing tests work. Hopefully there's an easy way to plugin a test for this new option.

jhump (commented):
> I'd be open to other naming patterns for this option, including making it "enum style".

I'm not sure an enum makes sense. But maybe a "suffix" string that defaults to "connect" -- then allows users to customize the location (like in case a "connect" sub-package were somehow a conflict) and then setting it to empty string would put it in the same package.

But, TBH, I think the simple bool is actually fine.

cmd/protoc-gen-connect-go/main.go
emcfarlane: Put up a testing strategy that runs the `protoc-gen-go` binary by re-compiling the `main.go` file. Wanted to bounce this off before extending. Currently it will assert that the expected files in the output but not the contents. We could add a `testdata/` directory of `.pb.go` files to check contents match if we wanted better testing of contents.

jhump: We definitely need to test contents, though I don't think "golden files" are sufficient since they are verified by human eyes. We actually need to compile them to make sure we are generating valid code: this option changes references, packages, and imports, which is a high risk of a compiler error if not done correctly.

To get as much coverage as possible with this option, we may even want to generate _all_ test files that we're currently using to test `protoc-gen-connect-go`, to a different package with this option and make sure it all compiles.

emcfarlane: Updated the testdata directory to a set of module proto files which is generated as part of `make generate`. I found having the golden files output in `testdata` nice to be able to browse the output. The test code imports these, which asserts that it is compilable and valid code, and then uses it for comparisons when running the different options.

jhump (approved):
LGTM!

PR #819, opened 2025-02-03
Fix issues with 'make generate'

jhump:
I happened to notice errors on my local machine when I would run `make generate`. I did a little digging and realize these were material errors that were happening in CI, too.

If you look at [this latest CI job](https://github.com/connectrpc/connect-go/actions/runs/13121672465/job/36608858280#step:4:71) from `main`, in the "Unit Test" step, you'll see output that repeats the following error message:
```
bash: line 1: buf: command not found
```

On my laptop, I was seeing a different error:
```
Failure: plugin protoc-gen-connect-go: exec: "protoc-gen-connect-go": executable file not found in $PATH
```

Doh! This error should have caused the `make generate` target to halt and fail, but it did not. This pull request remedies that.

There are two different issues afoot:
1. The `generate.sh` script would return success (zero exit code), even when the various commands it was running would fail. This is due to how the `-exec` flag to `find` works. So I changed it to instead use `xargs`, which will properly return a non-zero exit code if any of the processes it spawns fails.
2. The next issue was the underlying cause for the commands in `generate.sh` failing: the `PATH` variable was not properly configured. So the `make generate` target was building the dependencies but not making them available on the `PATH`.

Now that this is fixed, this also _fixes_ the generated code produced by that step so that it corresponds to the correct version of `protoc-gen-go` (i.e. the version pinned in `go.mod`).


emcfarlane (approved):

cmd/protoc-gen-connect-go/testdata/diffpackage/diffpackage.pb.go
```diff
@@ -14,7 +14,7 @@
 
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.36.0
+// 	protoc-gen-go v1.34.2
```
emcfarlane, line 17: Should we upgrade to 1.36.0?

PR #820, opened 2025-02-04
Make `AnyRequest` and `AnyResponse` nil safe interfaces

aviyam181199:
In some cases, we need to add generic behavior when using `connect`, such as logging requests and responses with proper formatting or handling infrastructure-related concerns.

Since `AnyRequest` and `AnyResponse` are interfaces, we can't simply check for nil at the top level—we also need to verify whether they represent a nil interface. This check is easy to overlook, especially when working with gRPC, where proto messages use nil-safe getters for every attribute.

To reduce redundant nil checks in business logic and prevent nil pointer dereferences, we should apply the same handling here.

### DIFF
--- a/connect.go
+++ b/connect.go
@@ -166,16 +166,25 @@ func NewRequest[T any](message *T) *Request[T] {
 // Any returns the concrete request message as an empty interface, so that
 // *Request implements the [AnyRequest] interface.
 func (r *Request[_]) Any() any {
+	if r == nil {
+		return nil
+	}
 	return r.Msg
 }
 
 // Spec returns a description of this RPC.
 func (r *Request[_]) Spec() Spec {
+	if r == nil {
+		return Spec{}
+	}
 	return r.spec
 }
 
 // Peer describes the other party for this RPC.
 func (r *Request[_]) Peer() Peer {
+	if r == nil {
+		return Peer{}
+	}
 	return r.peer
 }
 
@@ -183,6 +192,9 @@ func (r *Request[_]) Peer() Peer {
 // "Connect-" and "Grpc-" are reserved for use by the Connect and gRPC
 // protocols: applications may read them but shouldn't write them.
 func (r *Request[_]) Header() http.Header {
+	if r == nil {
+		return nil
+	}
 	if r.header == nil {
 		r.header = make(http.Header)
 	}
@@ -200,6 +212,9 @@ func (r *Request[_]) Header() http.Header {
 // if the request was never actually sent to the server (and thus no
 // determination ever made about the HTTP method).
 func (r *Request[_]) HTTPMethod() string {
+	if r == nil {
+		return ""
+	}
 	return r.method
 }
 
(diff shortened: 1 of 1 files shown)


jhump:
Why would your business logic ever need nil checks for these?

On the client, I am pretty sure you will get a nil-dereference panic if you provide a nil `*Request` pointer or if an interceptor tries to pass a nil `AnyRequest` to the next interceptor in the chain. Similarly, handlers will cause a nil-dereference panic if they return a nil response (except for signatures that return response or error, in which case a nil response is allowed if-and-only-if the error is non-nil).

The framework will never give a handler/interceptor a nil `AnyRequest` interface nor will it ever pass a nil `*Request` pointer. If you are seeing this happening, it must be a misbehaving interceptor higher up the stack. Similarly, the client will never return a nil response (except when it returns a non-nil error alongside the nil response).

These are not protobuf types, so the nil-safety of protobuf messages need not apply here. With protobuf messages, the nil safety is needed to make accessor chains less verbose in the face of nested message fields that are absent. This does not apply to `connect.Request` and `connect.Response` which have much stronger semantics and do not ever return/absent values (and they themselves should never be nil).

aviyam181199:
**Thanks for the quick response, @jhump!**  

I’d like to dig deeper into the example you provided:  
> *"except for signatures that return response or error, in which case a nil response is allowed if-and-only-if the error is non-nil."*  

Consider an interceptor responsible for logging, formatting, or manipulating responses. In such cases, it needs to account for scenarios where the response is `nil` due to an error. However, as you mentioned, it's not strictly guaranteed that a response will always be `nil` when an error is present. Since this is not an enforced constraint, an interceptor should still validate whether the response is a `nil` interface before proceeding.  

Would you suggest a better approach for handling this case?  

I agree that the motivation behind nil-safe proto getters is different, but I mentioned it as an analogy for a strategy that could also be useful here.

jhump:
I'm sorry, but that all sounds overly paranoid and overly complicated. I do _not_ think the `*Request` and `*Response` types should be amenable to use of nil pointers because that is a definite misuse.

The rules by which clients, handlers, and interceptors should abide are fairly simple and also quite idiomatic for Go code:
1. **Never** pass a nil pointer or nil interface as the value for a request or stream parameter. This goes for clients but also goes for interceptors when chaining to the next interceptor/handler in the chain.
2. **Never** return a nil pointer for a response or stream return value unless also returning a _non-nil_ error. For methods whose signature only returns a stream and cannot return an error, one must return a non-nil stream. (The stream's implementation could be to return a sentinel error from every method.) This applies to unary handlers and interceptors.
3. **Always** examine any error return value before examining a response or stream return value. If the error is not nil, assume that the response or stream is nil; **do not** examine/de-reference it. This applies to clients and interceptors.
4. **Always** examine any error return value before examining a message value, when receiving data from a stream. This is effectively the same as above: if the error is non-nil, assume the message is nil. This applies to most stream usage, including in clients, interceptors, and handlers.

In the face of bugs that violate the above rules and result in unexpected nil values, use `connect.WithRecover`, before any `connect.WithInterceptors` options, so that you can always catch and log the resulting bugs. For server handlers, you might even add a second `connect.WithRecover` option, _after_ `connect.WithInterceptors`, to catch panics in the handler and translate them into errors for your interceptor chain. The other recover function (installed before interceptors) would recover from any panics caused by the interceptor chain itself.

I suppose we could add the above rules to the docs somewhere. We likely took this sort of thing for granted since it is typical/expected of idiomatic Go. So rules of this sort may only be documented super-clearly in places where they _deviate_ from the above rules/idioms.

PR #824, opened 2025-02-06
Add RecoverInterceptor as alternative to WithRecover

### LINKED ISSUE #816: How to access `rpc.system` from `connect.WithRecover` callback
I've created a metric that counts the number of panics that occur in my handlers. I've aimed to align with the [OpenTelemetry Semantic Conventions for RPC Metrics], using the recommended attributes of `rpc.service` and `rpc.method`. For completeness, I would like to also include the required `rpc.system`, but see no obvious way to obtain this information from what is made available in [`connect.WithRecover`](https://pkg.go.dev/connectrpc.com/connect#WithRecover).

I can see in `otelconnect` that `rpc.system` is derived from `connect.AnyRequest.Peer().Protocol` (see https://github.com/connectrpc/otelconnect-go/blob/v0.7.1/interceptor.go#L101), but the peer is not included the parameters provided by `connect.WithRecover`.

Consider extending the information made available in `connect.WithRecover` to fully mirror the values accessible to `otelconnect.Interceptor`.

None at this time

**Additional context**
N/A

----

On a semi-related note, I've observed that the recover interceptor is a little awkward to compose with other interceptors. For example, I might want to add otelconnect, then recover, then validate, etc. such that I always have observability, but also have some panic protection for as many interceptors as possible. This ordering appears possible if you group interceptors by those that should be applied before and after the recover interceptor. But realising that required a bit of digging into how interceptors are chained together.

One way out of this small issue might be to make the recover interceptor public. Alternatively, one might expose the interceptor via `WithRecoverInterceptor` and `WithRecover` becomes `WithInterceptors(WithRecoverInterceptor(handle))`.

jhump:
This adds things to the handler signature (the other properties of `AnyRequest`, such as `Peer` and HTTP method) and also provides the interceptor separately, so it can be more easily combined with other interceptors. Finally, the value of ignoring `http.ErrAbortHandler` is unclear and adds a potential footgun in that a handler that panics with that value may inadvertently prevent capturing metrics or other side effects that are desirable for error-tracking in the face of a panic.

`WithRecover` remains backwards-compatible. But  it is deprecated; users should prefer the new `RecoverInterceptor` method and use it with `WithInterceptors`.

The most controversial parts of this change:
1. Using `AnyRequest` as the way to pass request properties, even for streaming RPCs. Does this seem like a bad idea? This was done to avoid adding a new exported type and to avoid a long parameter list.  One awkwardness with this approach is having a `req.Any()` method that isn't useful for client and bidi streams and is awkward to populate for server streams.
2. Can be used in clients, too, to recover from panics caused by client interceptors. (It could also technically recover from panics in the framework, though hopefully those kinds of issues don't exist!) This is more complicated for streaming calls since it means needing to wrap a stream in order recover from panics in various methods (send, receive, close...).

Thoughts?

Resolves #816.

### DIFF
--- a/recover.go
+++ b/recover.go
@@ -17,48 +17,292 @@ package connect
 import (
 	"context"
 	"net/http"
+	"sync/atomic"
 )
 
-// recoverHandlerInterceptor lets handlers trap panics, perform side effects
-// (like emitting logs or metrics), and present a friendlier error message to
-// clients.
-type recoverHandlerInterceptor struct {
-	Interceptor
+// RecoverInterceptor is an interceptor that recovers from panics. The
+// supplied function receives the context and request details.
+//
+// For streaming RPCs, req.Any() may return nil. It will always be nil
+// for client-streaming or bidi-streaming RPCs, since there could be
+// zero or even multiple request messages for such RPCs. For
+// server-streaming RPCs, it will be nil if the panic occurred before
+// the request message was received, which can happen if a panic occurs
+// in an interceptor before the RPC handler method is invoked.
+//
+// Similarly, for streaming RPCs, req.Header() may return nil. This
+// could happen in clients when the panic that is recovered occurs
+// before the stream is actually created and before request headers are
+// even allocated.
+//
+// Applications will generally want to add this interceptor first, which
+// means it will actually be the last to handle any results from the
+// RPC handler. This allows for recovering from the panics not only in
+// the handler but also in any other interceptors.
+//
+// The recovered value will never be nil. If panic was called with a nil
+// value, the recovered value will be a *[runtime.PanicNilError]. It must
+// return an error to send back to the client. If it returns nil, an
+// *Error with a code of CodeInternal will ne synthesized. The function
+// may also log the panic, emit metrics, or execute other error-handling
+// logic. The function must be safe to call concurrently.
+//
+// By default, handlers don't recover from panics. Because the standard
+// library's [http.Server] recovers from panics by default, this option
+// isn't usually necessary to prevent crashes. Instead, it helps servers
+// collect RPC-specific data during panics and send a more detailed error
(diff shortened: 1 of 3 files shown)


recover.go
```diff
@@ -17,48 +17,292 @@ package connect
...
+	return e.err
+}
+
+type recoverStreamingClientConn struct {
+	StreamingClientConn
+
+	//nolint:containedctx // must memoize the stream context to pass to recover handler
+	ctx    context.Context
+	handle func(context.Context, AnyRequest, any) error
+	req    atomic.Pointer[any]
```
emcfarlane, line 258: `atomic.Pointer` I think should be just `atomic.Value` here as you are using `any` anyway.

emcfarlane:
Passing the `AnyRequest` does feel odd for the streaming handlers. Would your solution in https://github.com/connectrpc/connect-go/pull/851#discussion_r2135765436 solve this case nicely too and we can pass `RequestInfo` to the handler?

jhump:
> Would your solution in https://github.com/connectrpc/connect-go/pull/851#discussion_r2135765436 solve this case nicely too and we can pass RequestInfo to the handler?

Oh, good idea. That would also completely eliminate any `atomic.Pointer`/`atomic.Value` shenanigans since that type doesn't include a request message. (The current signature does not get any request messages; I had only added it as an admittedly awkward component of re-using `AnyRequest`...)

I can update this PR to include a tentative `RequestInfo` interface type and then hold off merging until the shape of that interface is completely settled over in that other PR.

It might be slightly confusing for the handler to get a value that ostensibly includes response headers and trailers, but that is probably less awkward/confusing than using `AnyRequest`.

chrispine:
Do we still want to get this merged? Should I get someone on Toolchain to review?

PR #837, opened 2025-03-19
Support latest two versions of Go instead of latest three

jhump:
This changes the support posture for Go runtimes to mirror the Go SDK itself and to also mirror [recent changes](https://github.com/googleapis/google-cloud-go/commit/8a68e456d089e946bec93ffad6edb35eb0d29f93#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R23-R28) in the Google Cloud APIs for Go (starting January 1st, 2025).

The AWS APIs for Go still maintain support for three versions of Go, but they do have [a caveat](https://github.com/aws/aws-sdk-go-v2?tab=readme-ov-file#go-version-support-policy):
> **AWS reserves the right to drop support for unsupported Go versions earlier to address critical security issues.**

This caveat is topical because the reason I'm proposing changing to supporting only two versions is spurred by a security fix: [v0.36.0 of golang.org/x/net](https://github.com/connectrpc/connect-go/pull/830) requires [Go 1.23](https://github.com/golang/net/blob/v0.36.0/go.mod#L3).

I'm opening this PR to gather feedback:
* What do other maintainers think of this change?
* If we deem this change appropriate, is there anything else we should do to communicate the change to users?
* Should we instead take a softer stance, a la the AWS support policy? (In this case, we'd still document support for three versions, but we could technically go ahead and merge the above fix and _temporarily_ only support two versions.)

The impact of this change is fairly broad and will also end up requiring a similar change to seven other Go repos in the connectrpc organization, since they all depend on this one (basically all but otelconnect-go, which already is documented as supporting only two versions of Go).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,11 +12,15 @@ permissions:
   contents: read
 jobs:
   ci:
+    name: ci (go:${{ matrix.go-version.name }})
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        # when editing this list, also update steps and jobs below
-        go-version: [1.23.x, 1.24.x]
+        go-version:
+        - name: latest
+          version: 1.24.x
+        - name: previous
+          version: 1.23.x
     steps:
       - name: Checkout Code
         uses: actions/checkout@v4
@@ -25,7 +29,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v5
         with:
-          go-version: ${{ matrix.go-version }}
+          go-version: ${{ matrix.go-version.version }}
       - name: Unit Test
         run: make shorttest
       - name: Lint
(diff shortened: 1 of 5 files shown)


mattrobenolt:
I personally think this is reasonable, this is no issue for us, we try to maintain 1 version behind latest at most, we have a lot of services running 1.23.x still, but nothing older.

mattrobenolt (approved):

emcfarlane (approved):

bufdev (approved):

akshayjshah (approved):
I'm very much in favor of this change - let's do it!

To make future maintenance easier, why not use the "stable" and "oldstable" aliases instead of explicitly pinning particular Go releases? With the aliases, we won't need to do this biannual cross-repo chore.

jhump:
> To make future maintenance easier, why not use the "stable" and "oldstable" aliases instead of explicitly pinning particular Go releases? With the aliases, we won't need to do this biannual cross-repo chore.

@emcfarlane, @nicksnyder, and I discussed this. One issue is that there are actually nearly _always_ breakages when we upgrade a version of Go -- not from the new Go version itself but from `golangci-lint`. The new version of Go frequently has stuff in it (like in stdlib object files or code) that older versions of `golangci-lint` don't understand and/or don't like. So frequently we have to update the linter when we update Go versions. And updating the linter is always its own chore because new versions pretty much always include new linters that catch things in the code that we should revise/improve.

So we could use these labels, but we'd then need a nightly job so that we can be notified of red CI automatically when a new version becomes stable, instead of waiting for the first PR thereafter. And if we do this, `main` will be red for some period of time until we get the linter updated (which may have its own lag time between the Go release and the updated version of `golangci-lint` being releated) and the codebase back to green.

We decided maybe it's better to upgrade Go manually instead of having main be red for some window after a new Go version is released. So the latest commit I pushed still uses explicit Go versions, but it changes the CI job names to not include the Go version -- that way, we at least don't need to worry about adding/removing required checks when updating to the latest version of Go.

PR #841, opened 2025-03-27
Codemod to update syntax

mattrobenolt:
This is the result of running:

```
$ go run golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@latest -fix -test ./...
$ gofumpt -w .
````

Basically no substantial change except one update to use `maps.Copy`.

I'd personally recommend this as a nice habit to get into when we bump the minimum versions. I've found this picking up some good changes. It has even auto converted some calls like `strings.Split` into `strings.SplitSeq` in some of my codebases when applicable.

### DIFF
--- a/client_ext_test.go
+++ b/client_ext_test.go
@@ -715,7 +715,8 @@ type notModifiedPingServer struct {
 
 func (s *notModifiedPingServer) Ping(
 	_ context.Context,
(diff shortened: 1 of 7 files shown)


jhump (approved):
👍

Issue #848, opened 2025-05-27
Option to remove connect wrappers in generated code


anuraaga:
**Is your feature request related to a problem? Please describe.**

I understand `connect.Request` and `connect.Response` are useful types for providing access to metadata such as HTTP headers, but in practice I find myself absolutely never using such functions across many RPC servers/clients I've migrated to connect. They're essential in interceptors, but I think it's fairly common for users to not need these in RPC itself. But it gets tedious to have to call `.Msg` and feels like the framework working against me.

Another effect is not matching upstream stubs, so a migration does become both quite some work and also very difficult to do in a flag-guarded way preserving both code paths especially on the client side. If there was something awesome about calling `connect.NewRequest` and `res.Msg` on its own line after checking `err`, then it would be motivating, but it feels unnecessary.

I'm sure there must have been a lot of consideration put into this design decision to break from the upstream types, but wonder if there's been any thought otherwise on maybe allowing avoiding those types for users that don't need them?

**Describe the solution you'd like**

A code-gen option to provide client stubs and support server stubs that don't use `connect.Request` / `connect.Response`. The code-gen would probably be a simple enough place to wrap for providing to the framework which I think wouldn't need changes.

**Describe alternatives you've considered**

Continue to have extra verbosity / migration cost since it's still not a blocker. Also, I have found a generics based solution for the server side that I use both for this and other reasons, which allowed me to avoid the wrappers

https://github.com/curioswitch/go-curiostack/blob/main/server/connect.go#L18

I think a similar client version could be doable - basically instead of using stubs, the `Procedure` variables become the entry point to a single function. I might go with it for my own code but probably wouldn't recommend such a big change in paradigm for code I don't have sole ownership of.

**Additional context**

This isn't a highly important change but I thought I'd bring it up just to see what people's thoughts are on it.

anuraaga:
Thanks for the reference @jhump had checked the issues but forgot about discussions. Is it worth keeping this issue open? If there's direction on whether it's acceptable and an approach (personally I would go with an option rather than supporting both at the same time which gets more complicated, but anything would be fine), could try to help with a PR. I'm all-in on connect so would like to help make things easier if possible.

bufdev:
@anuraaga I've prototyped a solution at https://github.com/connectrpc/connect-go/pull/851 - take a look if you'd like.

PR #851, opened 2025-06-07
WIP: Prototype simple flag

bufdev:
This is re: https://github.com/connectrpc/connect-go/issues/848 https://github.com/connectrpc/connect-go/discussions/421. This is the simplest possible version of this, and is just done as a prototype for discussion.

The general concept is to effectively just model what people know already. The interface is very similar to grpc-go in this case. A final version of this will likely not be flag-based, we'll want to carefully choose how to expose this if we pursue this:

- A new plugin, such as `protoc-gen-connect-gosimple`.
- Generate to a separate package, such as `.*connectsimple`.
- Generate with a new interface name in the same package, such as `.*SimpleClient`.

Whatever we do, we do not want to v2 and will never break existing users.

### DIFF
--- a/client.go
+++ b/client.go
@@ -130,6 +130,15 @@ func (c *Client[Req, Res]) CallUnary(ctx context.Context, request *Request[Req])
 	return c.callUnary(ctx, request)
 }
 
+// CallUnarySimple calls a request-response procedure.
+func (c *Client[Req, Res]) CallUnarySimple(ctx context.Context, requestMsg *Req) (*Res, error) {
+	response, err := c.CallUnary(ctx, requestFromContext(ctx, requestMsg))
+	if response != nil {
+		return response.Msg, err
+	}
+	return nil, err
+}
+
 // CallClientStream calls a client streaming procedure.
 func (c *Client[Req, Res]) CallClientStream(ctx context.Context) *ClientStreamForClient[Req, Res] {
 	if c.err != nil {
@@ -167,6 +176,11 @@ func (c *Client[Req, Res]) CallServerStream(ctx context.Context, request *Reques
 		conn:        conn,
 		initializer: c.config.Initializer,
 	}, nil
+}
+
+// CallServerStreamSimple calls a server streaming procedure.
+func (c *Client[Req, Res]) CallServerStreamSimple(ctx context.Context, requestMsg *Req) (*ServerStreamForClient[Res], error) {
+	return c.CallServerStream(ctx, requestFromContext(ctx, requestMsg))
 }
 
 // CallBidiStream calls a bidirectional streaming procedure.
--- a/buf.gen.yaml
+++ b/buf.gen.yaml
@@ -10,5 +10,5 @@ plugins:
     opt: paths=source_relative
   - local: protoc-gen-connect-go
     out: internal/gen
-    opt: paths=source_relative
+    opt: paths=source_relative,simple=true
 clean: true
--- a/cmd/protoc-gen-connect-go/main.go
+++ b/cmd/protoc-gen-connect-go/main.go
@@ -90,6 +90,7 @@ const (
 	generatedFilenameExtension = ".connect.go"
 	defaultPackageSuffix       = "connect"
 	packageSuffixFlagName      = "package_suffix"
+	simpleFlagName             = "simple"
 
 	usage = "See https://connectrpc.com/docs/go/getting-started to learn how to use this plugin.\n\nFlags:\n  -h, --help\tPrint this help and exit.\n      --version\tPrint the version and exit."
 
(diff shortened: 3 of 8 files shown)


mfridman:
Strong +1 from me.

After a good few years and many servers/clients written for work and personal projects, my take is:

1. Unfamiliar and verbose migrating from twirp-go/grpc-go
2. Wrapping with `New{Request,Response}` is tedious, and so daily ergonomics suffer
3. Optimizes for the rare case, most code doesn't touch headers. So every call site pays the complexity cost
4. Context propagation is fine. For the rare times you need headers, passing via context is acceptable!

TL;DR - developer ergonomics matter more (esp. in this case) than avoiding the "context misuse" when that misuse is already the established patterns developers expect.

Alfus:
Strong +1 from me as well. Wrapping causes a lot of extra unneeded boilerplate code.

bufdev:
Speaking for myself, I still feel relatively strongly about generating to a separate package by default, but maybe could be convinced otherwise. I think this was a big miss on grpc-go's part, it sets up a whole host of problems w.r.t. plugin interactions. I get where your mind is though - grpc-go migration, import simplicity.

Note that with the implementation as-is in this PR now, it is just flag-based and generates to the same package as before. And this is configurable via package_suffix, including setting to empty to copy grpc-go semantics of generating to same package. I'd argue to keep it this way - two independent flags - as opposed to introducing special semantics w.r.t package if you set the simple flag.

bufdev:
A separate overall note: the implementation as-is fully allows and supports multiple protoc-gen-connect-to calls to independent packages, so for users who want both interfaces available, this is trivial. In fact, this is what this repository is now doing as of this PR: look at buf.gen.yaml, which generates to internal/gen/generics and internal/gen/simple.

mattrobenolt:
Oh yeah, I agree with the choices entirely to keep it a separate package, but more thinking in lines of expectations that users have.

This API that is being addressed here and the separate packages are the biggest real complains I've observed from discussions.

Even though we agree a separate package is better, I'm more looking at it from a lense of default expectations.

If I'm a user and I want this "simple" API, I likely am just wanting a grpc-go experience to align and be as much drop-in as possible.

I definitely don't hold a strong opinion here tho. I'm just pushing for a flip in the default behavior when opting into this simple API. But I think it's fine either way.

jhump (commented):
I'm a little worried about the fact that it generates incompatible code and requires a plugin option. This is mainly an issue for BSR generated SDKs -- we can't change the options used for the "connect-go" generated packages because of back-compat issues. Do you instead anticipate adding a "connect-go-simple" set of SDKs?

What if we generated both always, e.g. `PingServiceSimpleClient`, `PingServiceSimpleHandler`?

pkwarren:
This looks great - removing the request/response wrappers would remove a lot of boilerplate. My main feedback was related to supporting Peer/Spec and avoiding storing multiple values on the context, but @jhump covered that above: https://github.com/connectrpc/connect-go/pull/851#discussion_r2135765436.

client.go
```diff
@@ -141,6 +169,22 @@ func (c *Client[Req, Res]) CallClientStream(ctx context.Context) *ClientStreamFo
 	}
 }
 
+// CallClientStream calls a client streaming procedure in simple mode.
```
bufdev, line 172: As a reader, I don't know what simple mode is (I'm not sure what it is in this context myself). This, and the other methods for simple mode, need Godoc explaining what the differences are, likely 3-6 sentences minimum, especially as most share the same function signature.

client.go
```diff
@@ -180,7 +241,32 @@ func (c *Client[Req, Res]) CallBidiStream(ctx context.Context) *BidiStreamForCli
...
+// Request headers should be set in a [CallInfo] object inside the context using [NewClientContext]. These headers
+// are transmitted when this method is called and do not require an explicit call to [BidiStreamForClient.Send].
+//
+// Likewise, response headers and trailers should be read from the [CallInfo] object in context.
+func (c *Client[Req, Res]) CallBidiStreamSimple(ctx context.Context) (*BidiStreamForClient[Req, Res], error) {
+	stream := c.CallBidiStream(ctx)
+	if stream.err != nil {
+		return nil, stream.err
+	}
+	if err := stream.Send(nil); err != nil {
```
bufdev, line 255: Similarly, given that this is the only difference, I'm not sure if it's worth having a separate `CallBidiStreamSimple` and extra docs just to call `stream.Send` (assuming you handle `RequestHeader` no matter what)

client.go
```diff
@@ -180,7 +241,32 @@ func (c *Client[Req, Res]) CallBidiStream(ctx context.Context) *BidiStreamForCli
 	}
 }
 
+// CallBidiStreamSimple calls a bidirectional streaming procedure.
+//
+// Request headers should be set in a [CallInfo] object inside the context using [NewClientContext]. These headers
+// are transmitted when this method is called and do not require an explicit call to [BidiStreamForClient.Send].
+//
+// Likewise, response headers and trailers should be read from the [CallInfo] object in context.
+func (c *Client[Req, Res]) CallBidiStreamSimple(ctx context.Context) (*BidiStreamForClient[Req, Res], error) {
```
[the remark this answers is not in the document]
smaye81: Yep, apologies. I had this on a separate branch while I was waiting for the docs review so that it didn't over-complicate things.  Take a look at https://github.com/connectrpc/connect-go/pull/869, which should address your comment here.

bufdev:
LGTM! Let's ship it!

jrinehart-buf:
> LGTM! Let's ship it!

I'll get the docs PR up to date and ready for review.

staugaard:
When will we see this released?

PR #854, opened 2025-06-25
Add api type flag for generating an API that uses unwrapped types

smaye81:
This adds a new option to the the protoc-gen-connect-go plugin for generating client and server APIs that use unwrapped types. To use this option, set `api=simple` as a plugin option. For example:

### Usage
```diff
# buf.gen.yaml
plugins:
  - local: protoc-gen-connect-go
    out: gen
    opt:
      - paths=source_relative
+     - api=simple
```

With this option, the client and server APIs will now expect the raw request and response types rather than the Connect wrapped types. So, given the following protobuf definition, the APIs can be used as illustrated:

```proto
package foo.v1;

message FooRequest {}
message FooResponse {}

service FooService {
  rpc GetFoo(FooRequest) returns (FooResponse) {}
}
```

### Server
The server API for `GetFoo` becomes:

```go
func (f fooServer) Ping(ctx context.Context, request *foov1.FooRequest (*foov1.FooResponse, error)
```

instead of
```go
func (f fooServer) Ping(ctx context.Context, request *connect.Request[foov1.FooRequest] (*connect.Response[foov1.FooResponse], error)
```

### Client
Calling this service from a client becomes:

```go
resp, err := client.GetFoo(context.Background(), &foov1.FooRequest{})
```

instead of

```go
req := connect.NewRequest(&foov1.FooRequest{})
resp, err := client.GetFoo(ctx, req)
```

### Metadata

Using this option will require managing metadata for the request and response via Go's `Context`, which should be familiar to most users. A helper function `WithOutgoingContext` can be used which returns a context and a `CallInfo` object used for metadata. So for example, interacting with headers is done as follows:

```go
ctx, callInfo := connect.WithOutgoingContext(ctx)
callInfo.RequestHeader().Set("X-Req-Header", "header-val")
resp, err := client.GetFoo(ctx, req)
hdrVal := callInfo.ResponseHeader().Values("X-Resp-Header")
```

**This option is optional. The default behavior for the plugin is still to use wrappers for request and response types. So, users can opt-in to this generated behavior as they wish.**


bufdev:
Can you please do this on top of 851 so I can tell what changed?

PR #856, opened 2025-06-30
Implement CallInfo usage in context and add integration tests

smaye81:
This adds the usage of CallInfo in context for issuing requests with the new simple API. This builds on top of the https://github.com/connectrpc/connect-go/pull/851 which implements the simple flag for unary and server-streaming

In addition, it adds integration tests for the simple and generics API.

### DIFF
--- a/client.go
+++ b/client.go
@@ -19,7 +19,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"maps"
 	"net/http"
 	"net/url"
 	"strings"
@@ -128,23 +127,26 @@ func (c *Client[Req, Res]) CallUnary(ctx context.Context, request *Request[Req])
 	if c.err != nil {
 		return nil, c.err
 	}
-	ctx, ci := NewOutgoingContext(ctx)
-	call, ok := ci.(*callInfo)
-	if ok {
-		call.requestHeader = request.Header()
-	}
+	ctx, callInfo := newOutgoingContext(ctx)
+	callInfo.requestHeader = request.Header()
 
 	resp, err := c.callUnary(ctx, request)
 	if err != nil {
 		return nil, err
 	}
 
-	if ok {
-		call.peer = request.Peer()
-		call.spec = request.Spec()
-		call.method = request.HTTPMethod()
-		maps.Copy(call.ResponseHeader(), resp.Header())
-		maps.Copy(call.ResponseTrailer(), resp.Trailer())
+	callInfo.peer = request.Peer()
+	callInfo.spec = request.Spec()
+	callInfo.method = request.HTTPMethod()
+	if callInfo.responseHeader == nil {
+		callInfo.responseHeader = resp.Header()
+	} else {
+		mergeHeaders(callInfo.ResponseHeader(), resp.Header())
+	}
+	if callInfo.responseTrailer == nil {
+		callInfo.responseTrailer = resp.Trailer()
+	} else {
+		mergeHeaders(callInfo.ResponseTrailer(), resp.Trailer())
 	}
 
 	return resp, nil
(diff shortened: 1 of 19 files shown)


interceptor_ext_test.go
```diff
@@ -123,7 +123,7 @@ func TestOnionOrderingEndToEnd(t *testing.T) {
 	mux := http.NewServeMux()
 	mux.Handle(
 		pingv1connect.NewPingServiceHandler(
-			pingServer{},
+			pingServerGenerics{},
```
smaye81, line 126: This is because I had to create a new `pingServer` that uses the simple API also for tests. Since I figure we want `pingServer` to be the simple API, i instead renamed this one to `pingServerGenerics`.

connect_ext_test.go
```diff
@@ -165,7 +155,7 @@ func TestCallInfo(t *testing.T) {
 			assert.Equal(t, msg.GetNumber(), 1)
 			assert.Nil(t, stream.Close())
 			assert.Equal(t, stream.ResponseHeader().Values(handlerHeader), []string{headerValue})
-			assert.Equal(t, stream.ResponseTrailer().Values(handlerTrailer), []string{trailerValue})
```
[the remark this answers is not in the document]
smaye81: This is actually not working on main also. If I copy this test block to the main branch and run it, it also fails with the same error which is the trailer is not found. I was planning to investigate this to see if I'm just missing something with the setup or if this is a legit bug.

context.go
```diff
@@ -19,68 +19,246 @@ import (
...
-// SetResponseTrailer sets the response trailer within a simple handler implementation.
-func SetResponseTrailer(ctx context.Context, trailer http.Header) {
-	responseTrailerAddress, ok := ctx.Value(responseTrailerAddressContextKey{}).(*http.Header)
+func (w *responseWrapper[Res]) ResponseHeader() http.Header {
+	return w.response.Header()
+}
+
+func (w *responseWrapper[Res]) ResponseTrailer() http.Header {
+	return w.response.Trailer()
+}
```
smaye81, line 213: Think I got the placement right on these.

connect_ext_test.go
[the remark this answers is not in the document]
smaye81: Ignore this. Was a setup issue and has been fixed. Added trailer tests back in.

client.go
```diff
@@ -159,12 +187,33 @@ func (c *Client[Req, Res]) CallServerStream(ctx context.Context, request *Reques
 	if c.err != nil {
 		return nil, c.err
 	}
+	callInfo, callInfoOk := getClientCallInfoFromContext(ctx)
```
srikrsna-buf, line 190: I think this can be shifter to `newConn` to support client and bidi as well.

handler.go
```diff
@@ -169,6 +176,9 @@ func NewServerStreamHandler[Req, Res any](
 			if err != nil {
 				return err
 			}
+			ctx = newHandlerContext(ctx, &streamCallInfo{
+				conn: conn,
+			})
```
srikrsna-buf, line 181: I think we need to do this for client and bidi as well

client.go
smaye81: Maybe that should be added in https://github.com/smaye81/connect-go/pull/1, which implements client and bidi?

handler.go
smaye81: See above comment.

interceptor.go
```diff
@@ -103,3 +111,28 @@ func (c *chain) WrapStreamingHandler(next StreamingHandlerFunc) StreamingHandler
...
+		}
+		return next(ctx, spec)
+	}
+}
+
+func checkSentinel(ctx context.Context) bool {
+	callInfo, _ := ctx.Value(clientCallInfoContextKey{}).(*clientCallInfo)
+	sentinel, _ := ctx.Value(sentinelContextKey{}).(*clientCallInfo)
+	// Only verify if there's a sentinel call info to compare it to
+	return callInfo == sentinel
```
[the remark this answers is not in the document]
smaye81: Ah, not really. This was leftover from earlier code where I was checking if the value was nil in context, so I needed a type assertion on the interface. But that's no longer needed. Updated.

jhump (approved):
Looks good. Just a few minor comments.

smaye81:
@jhump incorporated all your feedback in https://github.com/connectrpc/connect-go/pull/856/commits/fa4e176fca38090b17c94d5936167fcddd82026d. Lmk what you think.

jhump (approved):

PR #872, opened 2025-08-21
Streaming full duplex h11

ii64:


emcfarlane:
Thanks for this PR. Closing for now as we need to rebase this work onto v2. Further investigation is needed on the correct solution, related issues https://github.com/connectrpc/connect-go/pull/796 and https://github.com/connectrpc/connect-go/pull/913.

PR #877, opened 2025-09-11
Remove `golang.org/x/net/http2` dependency from tests

maxbrunet:
Follow up to #873

Replace `golang.org/x/net/http2` by the standard `net/http` module for both clients and servers in all tests

The std lib implementation being more polished, it caught a few bugs like streams never closed

### DIFF
--- a/connect_ext_test.go
+++ b/connect_ext_test.go
@@ -1151,7 +1151,7 @@ func TestGRPCMissingTrailersError(t *testing.T) {
 		var connectErr *connect.Error
 		ok := errors.As(err, &connectErr)
 		assert.True(t, ok)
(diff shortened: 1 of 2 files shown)


jhump:
@maxbrunet, I updated this branch to latest `main` (which is a requirement for merging PRs in this repo), but now a test is failing (seems to be somewhat consistent since it failed in the same way in two CI jobs).

It admittedly doesn't look related to your changes, but master is green and this branch is not. Maybe it's a semantic merge conflict about an error condition that doesn't harmonize with your changes to `memhttp` or the fact that TLS is no longer being used? Let me know if you get a chance to investigate and conclude that something else is to blame other than your changes in this PR.

PR #881, opened 2025-09-18
Refactor memhttptest to work with Go 1.25 synctest

codefromthecrypt:
I noticed that things were basically already aligned for the Go 1.25 synctest package, just that a helper to shutdown the client could avoid bursted bubble panics. Feel free to tell me if this isn't desired, as I know it is random.

https://pkg.go.dev/testing/synctest@master

### DIFF
--- a/internal/memhttp/memhttptest/http_test.go
+++ b/internal/memhttp/memhttptest/http_test.go
@@ -1,0 +1,96 @@
+// Copyright 2021-2025 The Connect Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


internal/memhttp/memhttptest/http_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+package memhttptest_test
+
+import (
+	"bytes"
+	"io"
+	"net/http"
+	"strings"
+	"testing"
+	"testing/synctest"
```
jhump, line 23: We still support older version of Go where this opt-in. So I think you need to add a build tag directive above so that this file only applies when built/tested with 1.25 and above.

jhump:
@codefromthecrypt, can you take a look at the comments I left the other week? I think it would be a small revision to get CI passing and get this merged.

Issue #886, opened 2025-09-26
connect-go 1.19.0 fails to build on 32bit ARM


MDr164:
When trying to compile the new 1.19.0 version for an ARM32 target following error occurs. On the previous release this was not the case

```bash
GOOS=linux GOARCH=arm GOARM=7,softfloat go build

# connectrpc.com/connect
/home/mdr164/go/pkg/mod/connectrpc.com/connect@v1.19.0/envelope.go:378:24: math.MaxUint32 (untyped int constant 4294967295) overflows int
```

This works with any code importing connect-go as the issue is not in the generated code but rather directly in the dependency. The issue was introduced in 5e7196a96bf4b56be608d2abcd36b528c8766462 which was not present in 1.18.1

earl-warren:
Same error here.

https://codeberg.org/forgejo-integration/forgejo/actions/runs/12680#jobstep-10-15085

```
19.38 CGO_CFLAGS="-O2 -g -DSQLITE_MAX_VARIABLE_NUMBER=32766" /usr/local/go/bin/go build -trimpath  -tags 'netgo osusergo bindata timetzdata sqlite sqlite_unlock_notify' -ldflags '-linkmode external -extldflags "-static" -buildid= -s -w -X "main.ReleaseVersion=14.0-test" -X "main.MakeVersion=GNU Make 4.4.1" -X "main.Version=14.0.0-dev-2-42097bc0cb+gitea-1.22.0" -X "main.Tags=bindata timetzdata sqlite sqlite_unlock_notify" -X "main.ForgejoVersion=14.0.0-dev-2-42097bc0cb+gitea-1.22.0"' -o gitea
41.39 # connectrpc.com/connect
41.39 /go/pkg/mod/connectrpc.com/connect@v1.19.0/envelope.go:378:24: math.MaxUint32 (untyped int constant 4294967295) overflows int
99.47 make: *** [Makefile:851: static-executable] Error 1
````

tahadostifam:
I guess I've submitted this PR yesterday, but you didn't checked and even not closed it if fixed here. I wanted to remind it :))

#887

PR #888, opened 2025-09-27
Fix size check for envelope prefix (32bit build)

mikhailv:
This PR updates the size check in `makeEnvelopePrefix` to fix compilation error `connect@v1.19.0/envelope.go:378:24: math.MaxUint32 (untyped int constant 4294967295) overflows int` for 32-bit build.

Related issue: https://github.com/connectrpc/connect-go/issues/886


mikhailv:
oh, fix is already in pull requests: https://github.com/connectrpc/connect-go/pull/887

Issue #879, opened 2025-09-16
Separate handle of streaming timeout settings


catchv:
* feature request
- http.Server{ WriteTimeout: } is set as the timeout for all handlers.
- https://connectrpc.com/docs/go/deployment#timeouts-and-connection-pools
- https://github.com/golang/go/issues/16100
- http.NewResponseController is supported starting with Go 1.20.

* solve
- Introduce a timeout option when creating a New~ServiceHandler in generate.
(HandlerOption SetTimeout)

- sample code
`
func NewGreetStreamServiceHandler(svc GreetStreamServiceHandler, opts ...connect.HandlerOption) (string, http.Handler) {
	greetStreamServiceMethods := v1.File_greet_v1_greet_proto.Services().ByName("GreetStreamService").Methods()
	greetStreamServiceGreetStreamHandler := connect.NewClientStreamHandler(
		GreetStreamServiceGreetStreamProcedure,
		svc.GreetStream,
		connect.WithSchema(greetStreamServiceMethods.ByName("GreetStream")),
		connect.WithHandlerOptions(opts...),
	)
	return "/greet.v1.GreetStreamService/", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		switch r.URL.Path {
		case GreetStreamServiceGreetStreamProcedure:
			**rc := http.NewResponseController(w)**
			**rc.SetWriteDeadline(time.Now().Add(20 * time.Second))**
			greetStreamServiceGreetStreamHandler.ServeHTTP(w, r)
		default:
			http.NotFound(w, r)
		}
	})
}
`

Thanks.

catchv:
Please note that I created a middleware to use before the official reflection.

[deadline middleware](https://github.com/catchv/deadline)

PR #891, opened 2025-09-30
Enable timeout support for each service handler

R4R3D1FF:
Adds an option for read and write timeout in the handler options, that is passed to the handler config and then to the spec. Finally, ServeHTTP function of the handler reads the timeout values from the spec and enforces them using ResponseController.

This prevents slow or malicious clients from using up resources as that specific stream for the HTTP2 connection can be closed.

Deals with the following issue
https://github.com/connectrpc/connect-go/issues/879


emcfarlane (commented):
Thanks for the PR! I will bring up with the other maintainers to discuss the usability side. For the PR we need to complete testing and documentation.

huyquanha:
FYI, I've created [another PR](https://github.com/connectrpc/connect-go/pull/909) also to fix this issue, but with added protogen support, so users can set the custom timeouts using an option on the methods, in their `.proto` files. 

I've also added some tests both for the proto generation and handler deadline setting.

emcfarlane:
Thanks for the PR, and sorry for the slow follow-up. Going to close this. The same behavior can be done as a wrapper on top of the handler, without new options or `Spec` fields:

```go
func timeoutHandler(handler http.Handler, readTimeout, writeTimeout time.Duration) http.HandlerFunc {
      return func(responseWriter http.ResponseWriter, request *http.Request) {
              responseController := http.NewResponseController(responseWriter)
              now := time.Now()
              if readTimeout > 0 {
                      _ = responseController.SetReadDeadline(now.Add(readTimeout))
              }
              if writeTimeout > 0 {
                      _ = responseController.SetWriteDeadline(now.Add(writeTimeout))
              }
              handler.ServeHTTP(responseWriter, request)
      }
}
mux.Handle(path, timeoutHandler(handler, 10*time.Second, 10*time.Second))
```

This works per procedure and composes with any http.Handler, so it doesn't need to live in connect-go.

Deadlines set in `ServeHTTP` apply to the whole stream. For streaming RPCs, what we may want is per-message timeout limits in read and write that resets on every message. That has to live inside the stream's read and write paths, so it would be a separate design. Happy to continue that discussion on #879.

Issue #885, opened 2025-09-26
Response headers set in handler are lost when handler returns an error


whisper0077:
I'm very pleased with the simple flag feature! I tried it right away and found a bug.

**Describe the bug**

When an error occurs, values set in `CallInfo.ResponseHeader()` are not reflected in the response.

**To Reproduce**

The reproduction code is as follows.
https://github.com/whisper0077/connect-go/blob/main/example/main.go

1. Server Ping handler sets response header "x-custom-key: value" and returns an error if request.Number < 0
1. Client calls Ping with Number = 1 (success) and Number = -1 (error), then inspects callInfo.ResponseHeader()

- Observed behavior
For the successful call, the response header is present.
For the error call, the response header is absent.

- Expected behavior
Headers set by the handler should be sent to the client even when the handler returns an error.

```bash
$ git clone https://github.com/whisper0077/connect-go
$ cd example
$ go run main.go
2025/09/26 19:22:31 response header  ok: value
2025/09/26 19:22:31 response header err: 
```

**Environment (please complete the following information):**
- `connect-go` version or commit: `v1.19.0`
- `go version`:  `go1.25.1 linux/amd64`
- your complete `go.mod`:

```go
module connectrpc.com/connect

go 1.24.0

retract (
	v1.10.0 // module cache poisoned, use v1.10.1
	v1.9.0 // module cache poisoned, use v1.9.1
)

require (
	github.com/google/go-cmp v0.5.9
	google.golang.org/protobuf v1.36.9
)
```

**Additional context**
Related issue: #850 

In the code below, performing the merge of the response headers before returning the error appears to resolve the issue. Potential impacts on other components have not been examined.
https://github.com/connectrpc/connect-go/blob/v1.19.0/handler.go#L78-L90

emcfarlane:
Errors still propagate headers within their metadata. As a workaround you should still be able to cast as a connect error and extract the headers (`err.(*connect.Error).Meta()`). Agree though that this should play nicely with the new `CallInfo` type, looking into a fix.

PR #892, opened 2025-09-30
Fix CallInfo header/trailer propagation on error responses

emcfarlane:
This fixes CallInfo to properly propagate headers and trailers when handlers return errors, and makes error metadata accessible through CallInfo on the client side.

Headers/trailers set via `CallInfo.ResponseHeader()` and `CallInfo.ResponseTrailer()` in handlers are now merged into the connection before error responses are sent.

Fixes https://github.com/connectrpc/connect-go/issues/885


context.go
```diff
@@ -243,6 +244,36 @@ func (w *responseWrapper[Res]) ResponseTrailer() http.Header {
...
+		for k, v := range w.base.ResponseTrailer() {
+			combined[k] = v
+		}
+	}
+	if w.err != nil {
+		for k, v := range w.err.Meta() {
+			combined[k] = v
+		}
+	}
+	return combined
```
jhump, line 274: This doesn't seem right. On the client side, we already get the response headers and trailers from the underlying streaming conn and then use those to populate `err.Meta()`. Won't this duplicate metadata? Also, `err.Meta()` is not able to distinguish between headers and trailers, while the underlying connection is, so if this does _not_ duplicate metadata, I think we have a problem that headers sent by the server will get mashed together in the error metadata and then incorrectly reported by the client as trailers.

connect_ext_test.go
```diff
@@ -2968,6 +2968,328 @@ func TestSetProtocolHeaders(t *testing.T) {
...
+	// Handler sets up resopnse headers to test error propagation.
+	handler := &pluggablePingServerSimple{
+		ping: func(ctx context.Context, request *pingv1.PingRequest) (*pingv1.PingResponse, error) {
+			callInfo, ok := connect.CallInfoForHandlerContext(ctx)
+			if !ok {
+				return nil, connect.NewError(connect.CodeInternal, nil)
+			}
+			if request.GetNumber() < 0 {
+				callInfo.ResponseHeader().Set("x-custom-key", "ping-error")
+				callInfo.ResponseHeader().Set("x-callinfo-only", "from-callinfo")
```
jhump, line 2983: Seems like we might also want a `x-callinfo-and-error` key that is set from both and expected to be merged.

We should also be explicitly setting `callInfo.ResponseTrailer()` metadata (for both success and error cases).

connect_ext_test.go
```diff
@@ -2968,6 +2968,328 @@ func TestSetProtocolHeaders(t *testing.T) {
...
+		assert.Nil(t, response)
+
+		// CallInfo headers should be in ResponseHeader.
+		headerValue := callInfo.ResponseHeader().Get("x-custom-key")
+		assert.Equal(t, headerValue, "ping-error")
+		callinfoOnlyValue := callInfo.ResponseHeader().Get("x-callinfo-only")
+		assert.Equal(t, callinfoOnlyValue, "from-callinfo")
+		// Error.Meta() should be in ResponseTrailer.
+		errorMetaValue := callInfo.ResponseTrailer().Get("x-error-meta")
+		assert.Equal(t, errorMetaValue, "from-error-meta")
```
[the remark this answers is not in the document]
emcfarlane: Added asserts for empty returns in the non-expected cases.

context.go
```diff
@@ -243,6 +244,36 @@ func (w *responseWrapper[Res]) ResponseTrailer() http.Header {
...
+	combined := make(http.Header)
+	if w.base != nil {
+		for k, v := range w.base.ResponseTrailer() {
+			combined[k] = v
+		}
+	}
+	if w.err != nil {
+		for k, v := range w.err.Meta() {
+			combined[k] = v
+		}
```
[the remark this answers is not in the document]
emcfarlane: Removed

error.go
```diff
@@ -227,11 +227,9 @@ func (e *Error) AddDetail(d *ErrorDetail) {
...
 // Meta allows the error to carry additional information as key-value pairs.
 //
-// Metadata attached to errors returned by unary handlers is always sent as
-// HTTP headers, regardless of the protocol. Metadata attached to errors
-// returned by streaming handlers may be sent as HTTP headers, HTTP trailers,
-// or a block of in-body metadata, depending on the protocol in use and whether
-// or not the handler has already written messages to the stream.
+// Metadata attached to errors is returned either as HTTP headers or HTTP trailers
+// depending on the protocol, whether it is unary or streaming, and whether or
+// not the handler has already written messages to the stream.
```
jhump, line 232: I think we could be more accurate here and state that metadata attached to a handler error is accessible via Error.Metadata()` in the client, but it is also accessible via `CallInfo.ResponseTrailers` regardless of protocol.

I don't think it makes any sense to describe things about the physical transport, since that is protocol specific. Even though you widened it to allow it in HTTP headers or HTTP trailers, that isn't even accurate since streaming operations in gRPC-Web and Connect actually send them in the body.

connect_ext_test.go
```diff
@@ -2968,6 +2968,379 @@ func TestSetProtocolHeaders(t *testing.T) {
...
+			ctx, callInfo := connect.NewClientContext(t.Context())
+			response, err := client.Ping(ctx, &pingv1.PingRequest{Number: -1})
+			assert.NotNil(t, err)
+			assert.Nil(t, response)
+
+			headerValue := callInfo.ResponseHeader().Get("x-custom-key")
+			assert.Equal(t, headerValue, "ping-error")
+			assert.Equal(t, callInfo.ResponseHeader().Get("x-header-only"), "should-not-be-in-trailers")
+			assert.Equal(t, callInfo.ResponseTrailer().Get("x-trailer-only"), "should-not-be-in-headers")
+			assert.Equal(t, callInfo.ResponseHeader().Get("x-trailer-only"), "")
```
jhump, line 3136: using `http.Header.Get()` seems risky because it will succeed, even if the headers actually include more than one value. I think you need to use `Values()` and then update expectation to be a single-element slice literal.

connect_ext_test.go
```diff
@@ -2968,6 +2968,379 @@ func TestSetProtocolHeaders(t *testing.T) {
...
+			// Assert the protocol specific handling of error metadata is used.
+			switch protocol {
```
jhump, line 3138: Oof. Is it necessary to do this? If so, I think something else is wrong. I think we need to make the semantics protocol-agnostic, or else it's too error-prone.

I _think_ the likely issue is that we're not adding the trailer prefix to error metadata for connect unary errors. I think we should do that. I think we can confirm that it's backwards compatible by simply asserting that it still shows up in `err.Meta()` with the **expected** key (i.e. a key _without_ the trailer prefix).

In fact, you could likely assert in here that for unary, all headers are also present in `err.Meta()` (not just the trailers and items from the handler's `err.Meta()`).

error.go
emcfarlane: But connect-go always makes them available as the `ResponseHeader` and `ResponseTrailer` and they are _not_ currently always available in the `CallInfo.ResponseTrailers` (it depends on the protocol).

connect_ext_test.go
emcfarlane: Assert the values now.

error.go
jhump: > (it depends on the protocol)

This seems like a serious bug to me. Why does the caller have to be cognizant of the protocol? The point of multi-protocol support in this library is to provide consistent semantics, regardless of protocol.

jhump: Yeah, I'm convinced that you're right.

For Go and ES, previous code that uses `Error.Meta()` should be fine. But changing this so that Connect Unary error metadata is (semantically) in trailers instead of headers would be visible and arguably a breaking change for Kotlin and Swift, that surface metadata differently.

timostamm: With gRPC's "trailers-only" response, an RPC that produces an immediate error _may_ respond with a single terminal HEADER frame. The server might provide an API that allows setting header fields and error metadata or trailer fields separately, but the trailers-only response does not distinguish between them on the wire. In gRPC-Web, this translates to a HTTP 1.1 response with headers, and without a body. 

Because of this property of the gRPC protocol, it isn't possible to provide consistent behavior in an API for errors that exposes headers and trailers separately.

PR #895, opened 2025-10-07
Prepare for v1.19.1

emcfarlane:
Below are the draft release notes for v1.19.1.

### Bugfixes
* Fix bounds check on envelope for 32-bit archs by @emcfarlane in #887
* Fix CallInfo header/trailer propagation on error responses by @emcfarlane in #892

**Full Changelog**: https://github.com/connectrpc/connect-go/compare/v1.19.0...v1.19.1

### DIFF
--- a/connect.go
+++ b/connect.go
@@ -33,7 +33,7 @@ import (
 )
 
 // Version is the semantic version of the connect module.
-const Version = "1.20.0-dev"
+const Version = "1.19.1"
 
 // These constants are used in compile-time handshakes with connect's generated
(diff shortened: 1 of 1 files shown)


jhump:
Let's omit the commit that is just a README change since that isn't really a change to the software in this new version. I actually just added the "ignore-for-release" label to both of them, so they won't show up if you were to re-generate the release notes. I'm also editing your description to pull that item out (so you could just copy+paste from this PR description if you are okay with my edits).

PR #906, opened 2026-01-28
Fix typo in RELEASE.md

stefanvanburen:
Noticed this while borrowing the [RELEASE.md approach in vscode-buf][1], thanks!

[1]: bufbuild/vscode-buf#553


jhump (approved):
Thanks for the improvement!

PR #908, opened 2026-02-07
Skip sendMaxBytes check for EndStream control frames (#907)

### LINKED ISSUE #907: sendMaxBytes incorrectly applied to EndStream control frames in streaming RPCs
## Bug

`envelopeWriter.Write()` applies `sendMaxBytes` to **all** envelopes, including EndStream control frames. When a streaming message exceeds `sendMaxBytes`, the handler correctly gets `CodeResourceExhausted` from `stream.Send()`. But when connect-go tries to write the error into an EndStream frame via `MarshalEndStream`, the serialized error JSON itself may also exceed `sendMaxBytes`, causing the EndStream to be silently dropped. The client receives an empty HTTP 200 response with no error information.

## Code path

1. `stream.Send()` → [`envelopeWriter.Write()`](https://github.com/connectrpc/connect-go/blob/02f23a3/envelope.go#L162-L163) rejects the oversized message:
   ```go
   if w.sendMaxBytes > 0 && env.Data.Len() > w.sendMaxBytes {
       return errorf(CodeResourceExhausted, "message size %d exceeds sendMaxBytes %d", ...)
   }
   ```

2. Handler returns the error → [`connCloser.Close(err)`](https://github.com/connectrpc/connect-go/blob/02f23a3/handler.go#L333)

3. [`connectStreamingHandlerConn.Close()`](https://github.com/connectrpc/connect-go/blob/02f23a3/protocol_connect.go#L825-L830) calls `MarshalEndStream(err, ...)` — if this fails, the error is swallowed

4. [`MarshalEndStream()`](https://github.com/connectrpc/connect-go/blob/02f23a3/protocol_connect.go#L864) passes the EndStream envelope through the same `envelopeWriter.Write()`:
   ```go
   return m.Write(&envelope{Data: raw, Flags: connectFlagEnvelopeEndStream})
   ```

5. `envelopeWriter.Write()` checks `sendMaxBytes` again — the error JSON is typically larger than the limit → EndStream frame is never sent.

## connect-es does not have this issue

In `connect-es`, `endStreamSerialization` is created separately and is **not** wrapped with `limitSerialization`:

- Message serialization: [`limitSerialization(createBinarySerialization(...))`](https://github.com/connectrpc/connect-es/blob/main/packages/connect/src/protocol/serialization.ts#L86-L98) — applies `writeMaxBytes`
- EndStream serialization: [`createEndStreamSerialization()`](https://github.com/connectrpc/connect-es/blob/main/packages/connect/src/protocol-connect/handler-factory.ts#L117) — no size limit applied

washanhanzi:
Use two-stage fallback for oversized EndStream control frames

Instead of unconditionally bypassing the sendMaxBytes check for control
frames, implement a graduated fallback strategy in envelopeWriter:

  1. Try to send the full EndStream frame (error code + message + details).
  2. If it exceeds sendMaxBytes, call a reduce closure to strip error
     details in-place, re-marshal via a marshal closure, and retry.
  3. If even the minimal frame is too large (unreasonably low limit),
     force-send it anyway.

The fallback logic lives entirely in envelope.go's writeControlFrame,
which takes two closures from the protocol layer:

  - marshal: serializes the current frame state into bytes. Because it
    closes over the frame struct, it automatically reflects any mutations
    made by reduce.
  - reduce: mutates the captured frame struct to strip non-essential data
    (error details for Connect, grpc-status-details-bin for gRPC-Web).

This keeps sendMaxBytes enforcement intact for control frames while
guaranteeing that the client always receives at least a minimal error.

Fixes #907

### DIFF
--- a/connect_ext_test.go
+++ b/connect_ext_test.go
@@ -1814,6 +1814,70 @@ func TestHandlerWithSendMaxBytes(t *testing.T) {
 	})
 }
 
+func TestHandlerWithSendMaxBytesStreamEndStream(t *testing.T) {
+	// Regression test: when a streaming handler returns an error because a
+	// message exceeds sendMaxBytes, the protocol control frame carrying the
+	// error (Connect EndStream or gRPC-Web trailer frame) must not itself be
+	// rejected by the sendMaxBytes check.
+	t.Parallel()
+	const sendMaxBytes = 1 // Tiny limit so even a minimal message exceeds it.
+	testCases := []struct {
+		name             string
+		compressMinBytes int
+		clientOptions    []connect.ClientOption
+	}{
+		{
+			name:             "connect_uncompressed",
+			compressMinBytes: math.MaxInt,
+		},
+		{
+			name:             "connect_compressed",
+			compressMinBytes: 1,
+		},
+		{
+			name:             "grpcweb_uncompressed",
+			compressMinBytes: math.MaxInt,
+			clientOptions:    []connect.ClientOption{connect.WithGRPCWeb()},
+		},
+		{
+			name:             "grpcweb_compressed",
+			compressMinBytes: 1,
+			clientOptions:    []connect.ClientOption{connect.WithGRPCWeb()},
+		},
+	}
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			t.Parallel()
+			mux := http.NewServeMux()
+			mux.Handle(pingv1connect.NewPingServiceHandler(
+				&pluggablePingServer{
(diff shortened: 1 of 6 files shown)


jhump:
I don't think this is the correct approach. The "max send size" is there for a reason, to limit buffering/memory usage to read an envelope in the recipient. If the recipient has a similar max receive size, they will reject this frame.

I think a better approach is to fallback to a minimal end-stream frame if the normal end-stream frame is too large. If this minimal end-frame is still too large, then we could send it anyway since a minimal end-stream frame with a useful error message would probably be no more than 100 bytes (so if the max send and receive sizes are set that low, shame on whoever set them 😝).

washanhanzi:
Done. Also updated the PR comment. Now use two-stage fallback for oversized EndStream control frames.

Issue #356, opened 2022-09-01
Todos for upcoming Go releases


akshayjshah:
This issue tracks small cleanups we should do as we drop support for old Go releases.

## Go 1.18

- [x] Remove the `asMaxBytesError` function that uses string comparisons.

## Go 1.19

- [ ] Check for HTTP/2 error codes using `x/net/http2.StreamError` and `errors.As` (though we'll need to discuss whether we want an x/net dependency)
- [ ] Explore the new `http.ResponseController`'s capabilities and how we might expose them. ~In particular, read and write deadlines would be nice!~ (see Go 1.21 below)

## Go 1.20

- [ ] Use `http.ResponseController.EnableFullDuplex` to allow bidi streaming over HTTP/1.1.

## Go 1.21

- [ ] Explore the use of read and write deadlines on `http.ResponseController` (now that [panic bug](https://github.com/golang/go/issues/58237) is fixed in Go 1.22). Also see #604.

PR #913, opened 2026-03-12
Add HTTP/1.x bidi streaming support

emcfarlane:
This adds support for bidi streaming over HTTP/1 connections. It's an improvement over https://github.com/connectrpc/connect-go/pull/796 adding support for full-duplex streams. It makes use of the new [ResponseControler.EnableFullDuplex](https://pkg.go.dev/net/http#ResponseController.EnableFullDuplex) method to enable support. If disabled, this returns an error and we report the same error as previously. If enabled, the bidi connection is now allowed.

To test bidi streams over HTTP/1 `memhttp` is updated to use a buffered conn. WIth an unbuffered conn, both sides of the `net.Pipe` are stuck writing, and neither can read because the write is synchronous and blocking. With real TCP sockets, the kernel buffers absorb these writes and there's no deadlock. This better imitates a real conn implementation. Follows the approach taken in:  https://go-review.googlesource.com/c/go/+/740100

Closes https://github.com/connectrpc/connect-go/pull/872, related to https://github.com/connectrpc/connect-go/issues/356


envelope.go
```diff
@@ -285,15 +285,12 @@ func (r *envelopeReader) Unmarshal(message any) *Error {
...
 		// Drain the rest of the stream to ensure there is no extra data.
-		numBytes, err := discard(r.reader)
+		numBytes, _ := discard(r.reader)
 		r.bytesRead += numBytes
-		if err != nil {
-			err = wrapIfContextError(err)
-			if connErr, ok := asError(err); ok {
-				return connErr
-			}
-			return errorf(CodeInternal, "corrupt response: I/O error after end-stream message: %w", err)
```
emcfarlane, line 295: This reported the following error in the testcase `TestClientBidiOverHTTP1/cumsum_error`:
```
internal: corrupt response: I/O error after end-stream message: read tcp 127.0.0.1:61157->127.0.0.1:60942: wsarecv: An existing connection was forcibly closed by the remote host.
```
Fixed in: b5db2ac3fa40c8705967835c427386b023fba686

PR #919, opened 2026-04-17
Fix nil pointer deref in duplexHTTPCall under concurrent Send + CloseAndReceive

### LINKED ISSUE #918: Nil pointer dereference in duplexHTTPCall when Send races with CloseAndReceive on client-streaming RPC
## Summary

`duplexHTTPCall` nil-pointer-dereferences in `io.(*PipeWriter).Write` when `stream.Send` on a client-streaming RPC races with `stream.CloseAndReceive` (which reaches `CloseWrite`).

The comment on `duplexHTTPCall.makeRequest` explicitly states "This runs concurrently with Write and CloseWrite", so Send and CloseWrite are expected to be safe to call concurrently — this is a genuine bug, not API misuse.

Confirmed on `v1.19.1` and on `main` (`e299aa6`, 2026-04-17); the race-critical code is byte-for-byte identical between them.

## Stack trace

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=...]

goroutine ... [running]:
io.(*PipeWriter).Write(0x0, ...)
	io/pipe.go:161 +0x13
connectrpc.com/connect.(*envelope).WriteTo(...)
	connectrpc.com/connect/envelope.go:87 +0xc7
connectrpc.com/connect.(*duplexHTTPCall).Send(...)
	connectrpc.com/connect/duplex_http_call.go:119 +0x1f4
...
connectrpc.com/connect.(*ClientStreamForClient[...]).Send(...)
	connectrpc.com/connect/client_stream.go:77 +0x2d
<user code calling stream.Send from a goroutine>
```

## Root cause

In `duplex_http_call.go`:

```go
// Send (line ~94)
isFirst := d.requestSent.CompareAndSwap(false, true)
if isFirst {
    pipeReader, pipeWriter := io.Pipe()
    d.requestBodyWriter = pipeWriter    // ← only set in the isFirst branch
    d.request.Body = pipeReader
    d.request.GetBody = nil
    d.request.ContentLength = -1
    go d.makeRequest()
}
...
bytesWritten, err := payload.WriteTo(d.requestBodyWriter)  // ← unconditional read

// CloseWrite (line ~169)
if d.requestSent.CompareAndSwap(false, true) {
    go d.makeRequest()
    // requestBodyWriter is NOT initialised in this branch
    return nil
}
if d.requestBodyWriter != nil {
    return d.requestBodyWriter.Close()
}
return d.request.Body.Close()
```

If `CloseWrite` wins the CAS on `requestSent`, it sets `requestSent = true` and returns without initialising `requestBodyWriter`. A subsequent `Send` then observes `isFirst == false`, skips the pipe setup, and runs `payload.WriteTo(d.requestBodyWriter)` on a **nil** `*io.PipeWriter`. `(*io.PipeWriter).Write` dereferences `w.pipe`, hence the SIGSEGV at `addr=0x0`.

## Triggering pattern

The race fires whenever a client-streaming caller uses this shape:

```go
stream := client.Sum(ctx)
go func() {
    for msg := range inputChan {
        _ = stream.Send(msg)
    }
}()
resp, err := stream.CloseAndReceive() // concurrent with the goroutine's Send
```

`CloseAndReceive` calls `CloseRequest` (→ `CloseWrite`) on the main goroutine; if it reaches the CAS before the sender goroutine's first `Send` does, the bug fires deterministically on the next `Send` from the goroutine.

## Reproduction

Drop the following test into the connect-go test tree (`duplex_http_call_race_test.go`, `package connect_test`) and run `go test -run TestDuplexHTTPCallSendCloseAndReceiveNoNilDeref -race .`:

```go
// Copyright 2021-2025 The Connect Authors
//
// Licensed under the Apache License, Version 2.0 (the "License"); ...

package connect_test

import (
	"context"
	"net/http"
	"testing"
	"time"

	"connectrpc.com/connect"
	pingv1 "connectrpc.com/connect/internal/gen/connect/ping/v1"
	"connectrpc.com/connect/internal/gen/generics/connect/ping/v1/pingv1connect"
	"connectrpc.com/connect/internal/memhttp/memhttptest"
)

func TestDuplexHTTPCallSendCloseAndReceiveNoNilDeref(t *testing.T) {
	t.Parallel()

	mux := http.NewServeMux()
	mux.Handle(pingv1connect.NewPingServiceHandler(&pluggablePingServer{
		sum: func(_ context.Context, stream *connect.ClientStream[pingv1.SumRequest]) (*connect.Response[pingv1.SumResponse], error) {
			for stream.Receive() {
			}
			return connect.NewResponse(&pingv1.SumResponse{}), nil
		},
	}))
	server := memhttptest.NewServer(t, mux)
	client := pingv1connect.NewPingServiceClient(server.Client(), server.URL())

	const iterations = 5
	for range iterations {
		stream := client.Sum(t.Context())

		// Delay the first Send so CloseAndReceive reaches CloseWrite first
		// and wins the CAS on d.requestSent.
		sendDone := make(chan struct{})
		go func() {
			defer close(sendDone)
			time.Sleep(50 * time.Millisecond)
			_ = stream.Send(&pingv1.SumRequest{Number: 1}) // nil-derefs with the bug present
		}()

		if _, err := stream.CloseAndReceive(); err != nil {
			t.Fatalf("CloseAndReceive: %v", err)
		}
		<-sendDone
	}
}
```

Output on current `main`:

```
panic: runtime error: invalid memory address or nil pointer dereference
...
connectrpc.com/connect.(*duplexHTTPCall).Send(...)
	/home/.../connect-go/duplex_http_call.go:119 +0x1f4
...
connectrpc.com/connect_test.TestDuplexHTTPCallSendCloseAndReceiveNoNilDeref.func2()
	/home/.../connect-go/duplex_http_call_race_test.go:85 +0x85
FAIL	connectrpc.com/connect	0.065s
```

(The reproducer uses the existing `pluggablePingServer` from `connect_ext_test.go`.)

## Proposed fix

A fix PR is incoming. The approach: pre-create the request-body `io.Pipe` in `newDuplexHTTPCall` for client/bidi streams so `requestBodyWriter` is never observable as nil. The CAS on `requestSent` then only gates the `go d.makeRequest()` side-effect (which is idempotent w.r.t. the pipe setup).

simonferquel:
Fixes #918.

## Summary

`duplexHTTPCall.Send` used to initialise `requestBodyWriter` inside the branch of a CAS on `requestSent`. If `CloseWrite` (reached via `CloseAndReceive → CloseRequest → CloseWrite`) won the same CAS first it never initialised the pipe, so a subsequent concurrent `Send` dereferenced `nil` at `payload.WriteTo(d.requestBodyWriter)` and the process crashed with:

```
panic: runtime error: invalid memory address or nil pointer dereference
  io.(*PipeWriter).Write(...)
  connectrpc.com/connect.(*duplexHTTPCall).Send(...) duplex_http_call.go:119
  ...
  connectrpc.com/connect.(*ClientStreamForClient[...]).Send(...)
```

Since [`makeRequest`'s comment](https://github.com/connectrpc/connect-go/blob/main/duplex_http_call.go#L233) explicitly says _"This runs concurrently with Write and CloseWrite"_, concurrent `Send` + `CloseWrite` is expected to be safe; the nil-deref is a bug.

## Fix

Allocate the request-body `io.Pipe` in `newDuplexHTTPCall` for client-streaming and bidi RPCs so `requestBodyWriter` is set once, at construction, and never reassigned. The CAS on `requestSent` now only gates the `go d.makeRequest()` side-effect (which remains idempotent w.r.t. the pipe setup). `CloseWrite` can unconditionally close the pipe when it's non-nil.

Net behavior for each stream type:

| Stream type | Before | After |
|---|---|---|
| Unary | pipe never created | pipe never created (unchanged) |
| Server streaming | pipe never created | pipe never created (unchanged) |
| Client streaming | pipe created lazily in `Send`, racy | pipe created in constructor, race-free |
| Bidi | pipe created lazily in `Send`, racy | pipe created in constructor, race-free |

For client/bidi streams where no `Send` is ever called (caller only does `CloseWrite`/`CloseAndReceive`), the request body now carries an empty `io.Pipe` pair instead of `http.NoBody`; closing the write end makes the read end return `io.EOF` immediately, so the HTTP roundtrip still sends an effectively empty request body — behaviorally equivalent to the previous implementation.

## Regression test

`duplex_http_call_race_test.go` adds `TestDuplexHTTPCallSendCloseAndReceiveNoNilDeref`, which forces the bad ordering deterministically:

1. Opens a `PingService.Sum` client stream.
2. Starts a goroutine that issues a `stream.Send(...)` after a 50 ms delay.
3. Immediately calls `stream.CloseAndReceive()` on the main goroutine — which reaches `CloseWrite` and wins the CAS first.
4. Waits for the sender goroutine to finish.

Before this change the sender goroutine panics on the first iteration (verified against this tree at `HEAD~1` = `e299aa6`). With this change the test passes cleanly:

```
=== RUN   TestDuplexHTTPCallSendCloseAndReceiveNoNilDeref
--- PASS: TestDuplexHTTPCallSendCloseAndReceiveNoNilDeref (0.26s)
```

## Validation

- `go test -race -count=5 -run TestDuplexHTTPCallSendCloseAndReceiveNoNilDeref .` — passes (previously panics on iteration 0).
- `go test -race -count=1 ./...` — all packages pass, modulo a pre-existing flake in `TestConcurrentStreams` (`connect_ext_test.go:794: failed to send request: unknown: write envelope: EOF`) that I reproduced with the same frequency (~60 %) on clean `upstream/main` without any of the changes in this PR; unrelated to this fix.
- `golangci-lint run --modules-download-mode=readonly --timeout=3m0s` — clean.
- `go build ./...` across the main module and `internal/conformance` — clean.


duplex_http_call_race_test.go
```diff
@@ -0,0 +1,96 @@
...
+// Concurrent Send and CloseWrite are explicitly expected to be safe (see the
+// "This runs concurrently with Write and CloseWrite" comment on
+// duplexHTTPCall.makeRequest), so the nil-deref is a bug in duplexHTTPCall.
+//
+// This test provokes the bad ordering deterministically by delaying the
+// sender goroutine's first Send until after the main goroutine has entered
+// CloseAndReceive. With the bug present, the sender goroutine panics and
+// tears down the test binary; with the fix, the race is no longer reachable
+// and the test completes cleanly (also under -race).
+func TestDuplexHTTPCallSendCloseAndReceiveNoNilDeref(t *testing.T) {
```
emcfarlane: Move this testcase under duplex_http_call_test.go.

duplex_http_call.go
```diff
@@ -81,13 +84,30 @@ func newDuplexHTTPCall(
...
 	}
+	// For RPCs where the client sends multiple messages (client-streaming and
+	// bidi) we set up the request body pipe up-front so that requestBodyWriter
+	// is never observable as nil from any goroutine after the call has been
+	// handed out. Deferring this to the first Send previously required a CAS
+	// on requestSent to gate the setup, which could be lost to CloseWrite:
+	// CloseWrite won the CAS, skipped pipe setup, and a subsequent concurrent
+	// Send then nil-deref'd at payload.WriteTo(d.requestBodyWriter). Send and
+	// CloseWrite are explicitly expected to be safe under concurrent use (see
+	// the comment on makeRequest below).
```
emcfarlane:
```suggestion
	// Client-streaming and bidi RPCs stream the request body through an
	// io.Pipe. Set it up here so requestBodyWriter is assigned once at
	// construction and safe to read concurrently from Send and CloseWrite.
```

duplex_http_call.go
```diff
@@ -36,8 +36,11 @@ type duplexHTTPCall struct {
 	onRequestSend    func(*http.Request)
 	validateResponse func(*http.Response) *Error
 
-	// io.Pipe is used to implement the request body for client streaming calls.
-	// If the request is unary, requestBodyWriter is nil.
+	// io.Pipe is used to implement the request body for client-streaming and
+	// bidi calls. It is initialised in newDuplexHTTPCall for those stream
+	// types and is never reassigned after construction, so it is safe to read
+	// concurrently from any goroutine without further synchronisation. For
+	// unary and server-streaming RPCs it remains nil.
```
emcfarlane:
```suggestion
	// requestBodyWriter streams the request body for client-streaming and bidi
	// RPCs. Assigned once in newDuplexHTTPCall and never reassigned, so it is
	// safe to read without synchronisation. Nil for unary and server-streaming.
```

emcfarlane (approved):
Thanks for the detailed bug report and fix!

simonferquel:
Thanks for the quick feedback @emcfarlane, I integrated your suggestions.

emcfarlane (approved):

PR #923, opened 2026-04-23
Add NewInProcessHTTPClient for in-process RPCs

### LINKED ISSUE #694: Exporting `memhttp` / `memhttptest`

We are currently exploring migrating our project from the gRPC Go implementation to connect-go. Since our project uses several micro services, we have a lot of tests were we launch other microservices (in a defined state) to interact with our test target (or the other way around). For that we heavily rely on the `bufconn` package to avoid a lot of unnecessary (even though local) network traffic and network port management in the tests.

It seems that the internal packages `memhttp` / `memhttptest` offer a similar functionality. I wonder if there is an intent to publish these packages as an official testing solution?

Export the `memhttp` package.

emcfarlane:
This adds `NewInProcessClient(handler http.Handler) (HTTPClient, string)`, which returns an `HTTPClient` that dispatches requests directly to `handler`, along with a URL to pair with it. Full protocol fidelity for Connect, gRPC, and gRPC-Web, including bidi streaming.

```go
// Simple case — multi-value spread into the service constructor
client := pingv1connect.NewPingServiceClient(connect.NewInProcessHTTPClient(mux))
```
```go
// With options — unpack
hc, url := connect.NewInProcessHTTPClient(mux)
client := pingv1connect.NewPingServiceClient(hc, url, connect.WithGRPC())
```
---
Today there's no way to wire a Connect client to a handler without a network socket. Users composing multiple Connect services in the same binary have to stand up a TCP listener or reach into something like `internal/memhttp` with a custom solution.

Current implementation uses golang.org/x/net/http2 over in-memory pipes, with DisableKeepAlives to scope each ServeConn goroutine to a single request. The implementation is replaceable without breaking the signature.

This doesn't use memhttp.Server: it would bring a persistent accept-loop goroutine and a Shutdown/Close ceremony the API shouldn't expose, and its HTTP/1.1 support isn't useful here.

Adds golang.org/x/net as a direct dependency (already transitive via net/http, and already used in internal/conformance).

Closes #694.


in_process_client.go
```diff
@@ -0,0 +1,71 @@
...
+	transport := &http.Transport{
+		// DisableKeepAlives forces one connection per request, so each
+		// request's lifetime is also the lifetime of its ServeConn
+		// goroutine. No idle pool, no teardown to manage.
+		DisableKeepAlives: true,
+		DialContext: func(ctx context.Context, _, _ string) (net.Conn, error) {
+			if err := ctx.Err(); err != nil {
+				return nil, err
+			}
+			// TODO: implement buffered pipe to prevent deadlocks.
```
jhump: memhttp does not use a buffered pipe, and we've not experienced any deadlocks.

Even with a buffer, if the read side isn't reading, the write side could block forever. So I'm curious for more context on this comment: what example scenario do you foresee leading to deadlock and how would a buffer prevent it?

jhump (approved):
Just one question. Otherwise LGTM

in_process_client.go
emcfarlane: Ah your right, couldn't reproduced. Even with a large payload it's fine. The deadlock I noticed was specific to HTTP/1.1. Removed the TODO.

jhump:
One thing to consider is documenting (and probably even overriding) the peer address for calls that go through this in-memory connection. Looking at `net.Pipe`, I'm pretty sure its reported addresses mean that `req.RemoteAddr` (on the `http.Request`, that the given http handler will observe) will be simply `"pipe"`, which then ends pu in the connect `Request.Peer()` value. For the client, it gets the `Peer` from the dialed host, which in this case would always be `1.2.3.4`. If the client had middleware that tried to get the actual remote address (via `httptrace.ClientTrace`), I believe it would also report `"pipe"`.

These don't seem like very good/intuitive values. It is likely better to use a different hard-coded host in the in-process base URL to make it report something better, like maybe `http://inprocess`. Similarly, you might want to override the `RemoteAddr()` methods on the `net.Conn` instances returned by `net.Pipe` so they return an address that stringifies to `inprocess`, instead of `pipe`.

jhump:
> use a different hard-coded host in the in-process base URL to make it report something better, like maybe `http://inprocess`

Even better would likely be to use `httptrace.ClientTrace` so that a Connect client's `Peer()` sees the _actual_ remote address. Then you wouldn't need to touch this URL and could rely on the `net.Conn.RemoteAddr()` values. That would also let you return a value that is not ambiguous (i.e. a value that cannot possibly be an IP address or domain name), like `"<inprocess>"` (with angles brackets).

jhump:
FYI, those last couple of comments are not blocking. Feel free to merge this and maybe we can think about/address the remote addr concerns in a follow-up...

emcfarlane:
@jhump added an `"inprocess"` constant to return from the `net.Conn` address methods, overwrites the `"pipe"` from net.Pipe. Played a bit with `httptrace.ClientTrace` but will wait for a followup. I'll merge after you get a chance to have another look, thanks

pkwarren (approved):

emcfarlane:
Closing in favor of the implementation in v2: https://github.com/connectrpc/connect-go/pull/951

PR #927, opened 2026-05-20
Prepare for v1.20.0

oliversun9:
### Other changes
* Bump minimum supported Go version to 1.25 by @jonbodner-buf in #922
* Update Unary-Get query parameter order to match spec recommendation by @oliversun9 in #926

## New Contributors
* @jonbodner-buf made their first contribution in #922

**Full Changelog**: https://github.com/connectrpc/connect-go/compare/v1.19.2...v1.20.0

### DIFF
--- a/connect.go
+++ b/connect.go
@@ -33,7 +33,7 @@ import (
 )
 
 // Version is the semantic version of the connect module.
-const Version = "1.20.0-dev"
+const Version = "1.20.0"
 
 // These constants are used in compile-time handshakes with connect's generated
 // code.


jhump (approved):
FYI, I edited the description with my own editorial pass for the release notes. So please paste them in when cutting a release.

In particular, I removed the dep update for buf because we don't ever include those (you'll see that depend-a-bot PRs, for example, get auto-labeled and thus auto-excluded from automated release notes).

I also updated the one about the Go version bump to explicitly indicate *what* version we now require (which is really the only reason this might be a v1.20.0 instead of v1.19.3 since the one change you just landed smells kinda like a minor bug fix...).

PR #934, opened 2026-06-09
Fix issue with user-agent getting inappropriately injected into request headers

### LINKED ISSUE #933: framework ignores an explicitly empty user-agent header
In net/http, if the caller sets the "user-agent" header to an empty value explicitly (either a nil/zero-length slice of values in the `http.Header` map, or a value that is the empty string), it will skip adding its default "Go-http-client/1.1" user-agent.

But connect-go doesn't do the same thing. If I set an explicitly blank "user-agent" request header, the library will still add its own default user-agent (for example, "connect-go/1.17.0 (go1.22.4)"). I am working on a proxy and want to respect the incoming user-agent on a forwarded request. But if the incoming user-agent was missing, there's no way for me to force that property onto outbound requests that go through this library.

The issue is that protocol_grpc.go and protocol_connect.go both re-use the `getHeaderCanonical` helper function to decide if a user-agent is present. But this returns the same value regardless of whether the user-agent header is missing or is explicitly empty. These two call sites should use a different technique to probe for the "user-agent" header before deciding to inject the library's default user agent string.

jhump:
Resolves #933. If the user explicitly sets a "user-agent" header to a nil/empty value, don't inject the library's user-agent. This mirrors the handling of "user-agent" headers in the underlying net/http. So you can always opt-out of the library inserting its user-agent by setting a "user-agent" header. And if you don't want any value to be passed at all, simply insert the "user-agent" key with a nil slice or an empty value.

This augments the tests to cover new scenarios ("user-agent" value is a nil slice and "user-agent" value is a single empty-string) and also adds a case to make sure that when the incoming request has no "user-agent" header, the library does inject its own user-agent string.

### DIFF
--- a/header.go
+++ b/header.go
@@ -113,6 +113,17 @@ func getHeaderCanonical(h http.Header, key string) string {
 	return v[0]
 }
 
+func hasHeaderCanonical(hdr http.Header, key string) bool {
+	// Return true if the key exists at all. So if a user explicitly
+	// sets a nil/empty slice of values or an empty value, we still
+	// consider that to be present. This is more precise than
+	// getHeaderCanonical(...) == "" which indicates the value is
+	// not present, even when the key is present but explicitly
+	// unset/empty.
+	_, ok := hdr[key]
+	return ok
+}
+
 // setHeaderCanonical is a shortcut for Header.Set() which
 // bypasses the CanonicalMIMEHeaderKey operation when we
 // know the key is already in canonical form.
--- a/header.go
+++ b/header.go
@@ -113,15 +113,17 @@ func getHeaderCanonical(h http.Header, key string) string {
 	return v[0]
 }
 
-func hasHeaderCanonical(hdr http.Header, key string) bool {
-	// Return true if the key exists at all. So if a user explicitly
-	// sets a nil/empty slice of values or an empty value, we still
-	// consider that to be present. This is more precise than
-	// getHeaderCanonical(...) == "" which indicates the value is
-	// not present, even when the key is present but explicitly
-	// unset/empty.
-	_, ok := hdr[key]
-	return ok
+// setUserAgentIfAbsent sets the User-Agent header to the given value, but
+// only if the caller has not already provided one. We check for the key's
+// presence rather than comparing getHeaderCanonical(...) to "" so that a
+// caller can suppress the default entirely by explicitly setting a nil/empty
+// slice of values or an empty value.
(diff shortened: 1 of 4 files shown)


header.go
```diff
@@ -113,6 +113,17 @@ func getHeaderCanonical(h http.Header, key string) string {
 	return v[0]
 }
 
+func hasHeaderCanonical(hdr http.Header, key string) bool {
```
emcfarlane: This helper is a bit overkill and doc comment would be nicer on the func.

protocol_connect.go
```diff
@@ -319,7 +319,7 @@ func (c *connectClient) Peer() Peer {
 func (c *connectClient) WriteRequestHeader(streamType StreamType, header http.Header) {
 	// We know these header keys are in canonical form, so we can bypass all the
 	// checks in Header.Set.
-	if getHeaderCanonical(header, headerUserAgent) == "" {
+	if !hasHeaderCanonical(header, headerUserAgent) {
```
emcfarlane:
```suggestion
	if _, ok := header[headerUserAgent]; !ok {
```

protocol_grpc.go
```diff
@@ -237,7 +237,7 @@ func (g *grpcClient) Peer() Peer {
 func (g *grpcClient) WriteRequestHeader(_ StreamType, header http.Header) {
 	// We know these header keys are in canonical form, so we can bypass all the
 	// checks in Header.Set.
-	if getHeaderCanonical(header, headerUserAgent) == "" {
+	if !hasHeaderCanonical(header, headerUserAgent) {
```
emcfarlane:
```suggestion
	if _, ok := header[headerUserAgent]; !ok {
```

emcfarlane (approved):

header.go
jhump: Fair enough. I've just carved out the helper along a different axis, so hopefully it pays its weight now (and also is more readable, inlining the map queries). PTAL.

emcfarlane (approved):

header.go
emcfarlane: All good. think its still a bit overkill having a function for this though. In the protocol_*.go code we directly mutate headers:
```
	header[connectHeaderProtocolVersion] = []string{connectProtocolVersion}
	header[headerContentType] = []string{
		connectContentTypeForCodecName(streamType, c.Codec.Name()),
	}
```
Just a nit though.

PR #938, opened 2026-06-19
Fix data race and connection leak in duplexHTTPCall

emcfarlane:
The ctx.Done() branch added in #937 let read-side methods touch d.response without synchronizing with makeRequest (a data race), and a random select could shadow a ready response with a cancelled context, leaving CloseRead to skip closing the body (a leak). BlockUntilResponseReady now prefers responseReady, falling through to ctx.Done() only when no response is ready.

Race is captured on main: https://github.com/connectrpc/connect-go/actions/runs/27833079526/job/82374393481
This fix is for unreleased code, only exists on main.

### DIFF
--- a/duplex_http_call.go
+++ b/duplex_http_call.go
@@ -264,7 +264,7 @@ func (d *duplexHTTPCall) ResponseStatusCode() (int, error) {
 
 // ResponseHeader returns the response HTTP headers.
 func (d *duplexHTTPCall) ResponseHeader() http.Header {
-	if response := d.responseAfterReady(); response != nil {
+	if response := d.responseAfterReady(); response != nil { //nolint:bodyclose // body is closed by CloseRead
 		return response.Header
 	}
 	return make(http.Header)
@@ -272,7 +272,7 @@ func (d *duplexHTTPCall) ResponseHeader() http.Header {
 
 // ResponseTrailer returns the response HTTP trailers.
 func (d *duplexHTTPCall) ResponseTrailer() http.Header {
-	if response := d.responseAfterReady(); response != nil {
+	if response := d.responseAfterReady(); response != nil { //nolint:bodyclose // body is closed by CloseRead
 		return response.Trailer
 	}
 	return make(http.Header)
--- a/duplex_http_call.go
+++ b/duplex_http_call.go
@@ -245,6 +245,7 @@ func (d *duplexHTTPCall) Read(data []byte) (int, error) {
 }
 
 func (d *duplexHTTPCall) CloseRead() error {
+	// Ignore the error: an error response still has a body that must be closed.
 	response, _ := d.BlockUntilResponseReady()
 	if response == nil {
 		return nil // No response, so nothing to close.
@@ -265,6 +266,7 @@ func (d *duplexHTTPCall) ResponseStatusCode() (int, error) {
 
 // ResponseHeader returns the response HTTP headers.
 func (d *duplexHTTPCall) ResponseHeader() http.Header {
+	// Ignore the error: an error response still carries headers to forward.
 	if response, _ := d.BlockUntilResponseReady(); response != nil {
 		return response.Header
 	}
@@ -273,6 +275,7 @@ func (d *duplexHTTPCall) ResponseHeader() http.Header {
 
 // ResponseTrailer returns the response HTTP trailers.
 func (d *duplexHTTPCall) ResponseTrailer() http.Header {
+	// Ignore the error: an error response still carries trailers to forward.
 	if response, _ := d.BlockUntilResponseReady(); response != nil {
 		return response.Trailer
 	}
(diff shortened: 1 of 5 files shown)


duplex_http_call.go
```diff
@@ -245,11 +245,11 @@ func (d *duplexHTTPCall) Read(data []byte) (int, error) {
 }
 
 func (d *duplexHTTPCall) CloseRead() error {
-	_ = d.BlockUntilResponseReady()
-	if d.response == nil {
-		return nil
+	response := d.responseAfterReady()
+	if response == nil {
+		return nil // Response failed, close is successful.
```
jhump: Maybe this means "close is unnecessary", since we haven't closed anything yet?

duplex_http_call.go
```diff
@@ -297,6 +295,21 @@ func (d *duplexHTTPCall) BlockUntilResponseReady() error {
 	}
 }
 
+// responseAfterReady blocks until the response is ready, then returns it, or
+// nil if the context was cancelled before the response arrived. A nil result
+// means makeRequest may still be writing d.response, so it must not be read.
+// An error-status response is returned like any other, so callers can close
+// its body and read its headers.
+func (d *duplexHTTPCall) responseAfterReady() *http.Response {
+	_ = d.BlockUntilResponseReady()
```
jhump: This is a bit confusing.

IMO, `BlockUntilResponseReady` should return both the response and the error -- that would be more straight-forward and would fix the race condition. To fix the connection leak, have it always check `d.responseReady` at the end -- in other words, in the case for `ctx.Done()`, have it do what this function does and do a non-blocking read of `d.responseReady` and return `d.response, d.responseErr` if it's ready (despite the context being canceled).

Even though `BlockUntilResponseReady` is technically exported, it's on an unexported type that is never handed out to external callers, so it should be fine/safe to change its signature.

duplex_http_call.go
emcfarlane: Reworded.

duplex_http_call.go
emcfarlane: Changed the signature.

duplex_http_call.go
```diff
@@ -245,9 +245,9 @@ func (d *duplexHTTPCall) Read(data []byte) (int, error) {
 }
 
 func (d *duplexHTTPCall) CloseRead() error {
-	response := d.responseAfterReady()
+	response, _ := d.BlockUntilResponseReady()
```
jhump: For consistency with other checks and with most Go code, instead of ignoring the error return and only checking `response == nil`, capture the error return as `err` and then check `err != nil`. Some of the checks do that, and then some use this style.

duplex_http_call.go
```diff
@@ -256,23 +256,24 @@ func (d *duplexHTTPCall) CloseRead() error {
...
 		return 0, err
 	}
-	return d.response.StatusCode, nil
+	return response.StatusCode, nil
 }
 
 // ResponseHeader returns the response HTTP headers.
 func (d *duplexHTTPCall) ResponseHeader() http.Header {
-	if response := d.responseAfterReady(); response != nil { //nolint:bodyclose // body is closed by CloseRead
+	if response, _ := d.BlockUntilResponseReady(); response != nil {
```
jhump: Here, too

duplex_http_call.go
```diff
@@ -256,23 +256,24 @@ func (d *duplexHTTPCall) CloseRead() error {
...
 		return response.Header
 	}
 	return make(http.Header)
 }
 
 // ResponseTrailer returns the response HTTP trailers.
 func (d *duplexHTTPCall) ResponseTrailer() http.Header {
-	if response := d.responseAfterReady(); response != nil { //nolint:bodyclose // body is closed by CloseRead
+	if response, _ := d.BlockUntilResponseReady(); response != nil {
```
jhump: And here

jhump (approved):
Just a nit on consistency for how you're checking the values returned from `BlockUntilResponseReady`. Otherwise LGTM

duplex_http_call.go
emcfarlane: It's intentional. We need to close the body if the response is given back, even if an error is returned. An error can be from response validation.

duplex_http_call.go
emcfarlane: Same, error response we still want to forward whatever headers we received.

duplex_http_call.go
jhump: Under what conditions would this return a non-nil response _and_ a non-nil error? TBH, I think there are none. We only return `d.response` along with `d.responseErr`, which I thought came from the transport. The transport won't return both non-nil -- you get either get a response or an error if it could not create a response (IIUC). If there _are_ such conditions that truly warrant this formulation, I think this definitely warrants a comment here and the other call sites below as it is a bit counter-intuitive to read and slightly out of the norm in Go code.

emcfarlane: Not many cases, but these ones: https://github.com/connectrpc/connect-go/blob/621486895950dbbd6820f5790abc8ab076a66ca8/duplex_http_call.go#L347-L366

jhump: Hmm. Do we ever actually _use_ `d.response` in these cases where we also set `d.responseErr`? Seems like it would be better for this code to only publish a non-nil `d.response` when it writes a nil `d.responseErr` (and this function could then handle cleaning up the response, by closing the body). Returning both non-nil, especially if the body is never actually used when err is non-nil, is confusing and pushing clean-up responsibility to distant code paths, which is error/leak-prone. 🤷

PR #946, opened 2026-07-13
refactor: Replace sort package with slices package

dongjiang1989:
## Summary
- Replace all usages of the `sort` package with the modern `slices` package (Go 1.21+)
- Use `slices.Sort` instead of `sort.Strings` for cleaner and more concise code
- Enable `depguard` linter to prevent future use of the deprecated `sort` package

## Changes
- `protocol.go`: Replace `sort.Strings` with `slices.Sort` (2 occurrences)
- `connect_ext_test.go`: Replace `sort.Strings` with `slices.Sort` (2 occurrences)
- `.golangci.yml`: Enable `depguard` linter with rule to deny `sort` package imports

## Test plan
- [x] `make lint` passes with 0 issues
- [x] `make test` passes
- [x] `go test -race ./...` passes with no race conditions

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -3,7 +3,6 @@ linters:
   default: all
   disable:
     - cyclop           # covered by gocyclo
-    - depguard         # unnecessary for small libraries
     - funcorder        # consider enabling in the future
     - funlen           # rely on code review to limit function length
     - gocognit         # dubious "cognitive overhead" quantification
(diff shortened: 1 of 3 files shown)


emcfarlane (commented):
All replaces are for `sort.Strings` which delegates to `slices.Sort` already: https://cs.opensource.google/go/go/+/refs/tags/go1.26.5:src/sort/sort.go;l=181

emcfarlane:
Thanks for taking the time to put the PR together! We do want to modernize these calls but as stated above this already delegates. In general, we try to avoid PRs that only touch style/idiom. Appreciate the interest in contributing. Closing for now.

PR #953, opened 2026-07-30
V2 implement eager stream creation

emcfarlane:
> Until now v2 kept v1's deferred dispatch: a client or bidi stream sends nothing until the first Send. That kept migration one-to-one and kept all four RPC shapes consistent. This PR is raised for direct feedback from maintainers.

This PR proposes a behavior change. Client-streaming and bidi RPCs send the request when the stream is created, before the first Send.

###  Changes

v2 has no stream header method, so the natural port of v1 generics code moves `stream.RequestHeader()` writes to `info.RequestHeader()` in the same position, after the stream is created. v2 as implemented defers the `CallInfo` capture to the first `Send` so that this port works unchanged:

```go
// v1 generics
stream := client.CumSum(ctx)
stream.RequestHeader().Set("X-Tenant", "acme")
// the natural v2 port, works today, but wouldn't with this PR
ctx, info := connect.NewClientContext(ctx)
stream, err := client.CumSum(ctx)
info.RequestHeader().Set("X-Tenant", "acme")
```

This PR takes a migration cost in exchange for eager dispatch. Header writes after the call must move above it, and connect-go-v2-migrate warns on each one:

```go
// after this PR
ctx, info := connect.NewClientContext(ctx)
info.RequestHeader().Set("X-Tenant", "acme")
stream, err := client.CumSum(ctx)
```

This matches the v1 simple mode contract: headers are captured and the request is sent when the method is called. The server sees the RPC without waiting for a client message, and a canceled context fails at creation instead of on the first `Send`.

### Summary

Removes the `SendHeaders()` method from `ClientStream`. The headers are captured on stream creation. `ServerStream` keeps it to send response headers without a message.

Unary and server-streaming RPCs still send on `Send` inside the call.  Buffering the single request message preserves Content-Length, retries via GetBody, and the GET fallback. Callers cannot observe the difference.

The tool `connect-go-v2-migrate` warns on `Send(nil)` and on header writes after stream creation.

### DIFF
--- a/cmd/connect-go-v2-migrate/rewrite.go
+++ b/cmd/connect-go-v2-migrate/rewrite.go
@@ -58,6 +58,8 @@ const (
 	ruleInterceptorMigration = "interceptor_migration"
 	ruleStreamParamType      = "stream_param_type"
 	ruleStreamParamAmbiguous = "stream_param_ambiguous"
+	ruleStreamSendNil        = "stream_send_nil"
+	ruleRequestHeaderTiming  = "request_header_timing"
 	ruleEcosystemMigration   = "ecosystem_migration"
 	ruleBufgenReinstall      = "bufgen_reinstall_plugin"
 	ruleBufgenGoMod          = "bufgen_update_go_mod"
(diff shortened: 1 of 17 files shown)


emcfarlane:
Closing due to no feedback. Behaviour as is, is consistent and fine and reduces the chance of migration to v2 causing any issues with headers.

PR #957, opened 2026-08-17
Only send a 304 when the request was a GET

NotAFlightRisk:
The 304 branch in `connectUnaryHandlerConn.Close` works out whether it's answering a GET by counting query parameters, and `Peer.Query` gets filled in for every method. So any POST that happens to carry a query string gets the conditional-GET treatment. Same handler returning the same `NewNotModifiedError`, one extra param on the URL:

```
POST /svc/Method          ->  500  application/json  {"code":"unknown","message":"not modified"}
POST /svc/Method?foo=bar  ->  304  no body
```

The message and any error details go with the body, and a connect-go client turns the second one into `unknown: 304 Not Modified`. Thats not hard to land on - a base URL with a param baked into it, or a proxy tacking on cache-busting ones.

`mergeResponseHeader` a few lines down already checks `request.Method`, and the client side only honours a 304 for GETs, so Im fairly confident the method is what the query count was standing in for.

The test bolts the query string on in a wrapper handler rather than through the client, beacuse a connect-go client won't put one on a POST by itself. `TestGetNotModified` still covers the GET side.

### DIFF
--- a/protocol_connect.go
+++ b/protocol_connect.go
@@ -728,7 +728,7 @@ func (hc *connectUnaryHandlerConn) Close(err error) error {
 		hc.mergeResponseHeader(err)
 		// If the handler received a GET request and the resource hasn't changed,
 		// return a 304.
-		if len(hc.peer.Query) > 0 && IsNotModifiedError(err) {
+		if hc.request.Method == http.MethodGet && IsNotModifiedError(err) {
 			hc.responseWriter.WriteHeader(http.StatusNotModified)
 			return hc.request.Body.Close()
 		}
--- a/client_ext_test.go
+++ b/client_ext_test.go
@@ -202,6 +202,25 @@ func TestGetNotModified(t *testing.T) {
 	assert.Equal(t, http.MethodGet, unaryReq.HTTPMethod())
 }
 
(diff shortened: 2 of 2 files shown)


protocol_connect.go
```diff
@@ -728,7 +728,7 @@ func (hc *connectUnaryHandlerConn) Close(err error) error {
 		hc.mergeResponseHeader(err)
 		// If the handler received a GET request and the resource hasn't changed,
 		// return a 304.
-		if len(hc.peer.Query) > 0 && IsNotModifiedError(err) {
+		if hc.request.Method == http.MethodGet && IsNotModifiedError(err) {
```
emcfarlane, line 731: This makes sense, I'm not sure why the request method wasn't checked. How did you run into this issue though?

emcfarlane (approved):
Thanks!

protocol_connect.go
NotAFlightRisk: was just poking around 😅

But no, I was debugging why one of our POST handlers were coming back to the client as unknown: 304 instead of the actual error message. the base URL we build clients from has a query param on it (for tracing).

PR #959, opened 2026-08-25
Fix panic in stream Spec and Peer when client construction failed

NotAFlightRisk:
`Spec()` and `Peer()` on a client stream panic with a nil pointer dereference when the client itself failed to build. `NewClient` holds that error back for the first call rather than returning it, so a URL with no scheme, or an option it doesnt like, still gets you a stream. `CallClientStream` and `CallBidiStream` return it on its own. No error beside it, so there's nothing to check first. The simple variants do return an error, but the stream they give you alongside panics just the same - their `stream == nil` guard isn't the state those calls produce.

Every other method on both types has checked that error since #265. `Spec` and `Peer` arrived in #364 three months later and missed it. `RequestHeader` on the same struct already answers this exact case with an empty `http.Header{}`, so empty `Spec{}` and `Peer{}` it is.

The assertions went in `TestNewClient_InitFailure` as well as the unit tests, since thats the one that builds a client through the public API - both of its subtests segfault on main without the guard. The other way round would be dropping an `errStreamingClientConn` into the four error paths in `client.go`, the way the interceptor chain does, but that turns `Conn()` on a failed client from nil into a live sentinel and the existing tests assert its nil.

### DIFF
--- a/client_ext_test.go
+++ b/client_ext_test.go
@@ -69,6 +69,8 @@ func TestNewClient_InitFailure(t *testing.T) {
 	t.Run("bidi", func(t *testing.T) {
 		t.Parallel()
 		bidiStream := client.CumSum(t.Context())
+		assert.Equal(t, bidiStream.Spec(), connect.Spec{})
+		assert.Equal(t, bidiStream.Peer(), connect.Peer{})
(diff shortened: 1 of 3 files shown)


emcfarlane (approved):
Thanks!

PR #961, opened 2026-08-25
Set finite default for ReadMaxBytes

emcfarlane:
Set a default `ReadMaxBytes` of 4 MiB for both clients and handlers. Both previously defaulted to no limit. 

The limit applies per message. Messages over the limit fail with `CodeResourceExhausted`.

This is a behavior change for anyone relying on unbounded reads. `WithReadMaxBytes(0)` restores the previous behavior, and existing `WithReadMaxBytes(n)` calls are unaffected. Handlers that want to bound the total size of a request stream rather than individual messages should still use `http.MaxBytesHandler`.

### DIFF
--- a/client.go
+++ b/client.go
@@ -334,6 +334,7 @@ func newClientConfig(rawURL string, options []ClientOption) (*clientConfig, *Err
 		Procedure:        protoPath,
 		CompressionPools: make(map[string]*compressionPool),
 		BufferPool:       newBufferPool(),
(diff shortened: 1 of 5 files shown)


emcfarlane:
Closing, as I believe it to be too disruptive. This default will be set in v2 and the migrate tool will set the `WithReadMaxBytes` option to 0 to maintain the current behavior.

PR #966, opened 2026-08-28
Create release binary for protoc-gen-connect-go

### LINKED ISSUE #695: Include prebuilt executables in releases
In order to `buf generate`, my environment must have `go` installed so that I can first install `protoc-gen-connect-go`.

Include prebuilt executables with releases.

Install go and use `go install...` :smile: 

**Additional context**
This is certainly a convenience request. My current situation is trying to create more efficient CI jobs. I can download [`protoc-gen-go`](https://github.com/protocolbuffers/protobuf-go/releases/tag/v1.32.0) and [`buf`](https://github.com/bufbuild/buf/releases/tag/v1.29.0) without needing to bloat the build environment by installing go and associated tools. It would be nice to be able to simply download an executable for `protoc-gen-connect-go` as well.

emcfarlane:
This adds release artifacts and updates the RELEASE.md to validate.

Builds `protoc-gen-connect-go-vX.Y.Z-{Darwin,Linux,Windows}-{x86_64,arm64/aarch64}` archives, each containing the binary, LICENSE and checksum.

Closes https://github.com/connectrpc/connect-go/issues/695

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,25 @@
+name: release
+on:
+  release:
+    types: [published]
+permissions:
+  contents: write
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    # Release events fire for every tag, limit to target vX.Y.Z.
+    if: ${{ startsWith(github.event.release.tag_name, 'v') }}
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v7
+        with:
+          fetch-depth: 1
+      - name: Install Go
+        uses: actions/setup-go@v7
+        with:
+          go-version: 1.27.x
+          check-latest: true
+      - name: Release
+        env:
+          GITHUB_TOKEN: ${{ github.TOKEN }}
+        run: make release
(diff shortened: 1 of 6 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,25 @@
+name: release
+on:
+  release:
+    types: [published]
+permissions:
+  contents: write
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    # Release events fire for every tag, limit to target vX.Y.Z.
```
anuraaga, line 10: I suspect releases are always published only for the v tags. On the flip side, if a release was published for some reason for a different version, shouldn't it get binaries in the same way?

anuraaga (approved):

.github/workflows/release.yaml
emcfarlane: We will do releases for the migration tool in v2. This will be a tag like `cmd/connect-go-v2-migrate/v1.0.0` which we don't want protoc and its not necessary to have a binary for it. So this protects against that, but I think it will always have to be a semver release otherwise. Can relax this restriction later if needed.

pkwarren (approved):

PR #967, opened 2026-09-01
V2 configure codecs and compressors as ordered lists

emcfarlane:
This PR moves codec and compressor configuration from additive options to authoritative slices (`WithCodecs` and `WithCompressors`). This resolves a few historical API limitations and lets the caller state the preference order directly.

Compression negotiation is unchanged. The server still responds with the client's first supported entry, matching v1. The recent spec update https://github.com/connectrpc/connectrpc.com/pull/405 would let the server choose the response encoding instead, which matters because browsers hardcode `Accept-Encoding: gzip, deflate, br, zstd` and can never negotiate past `gzip` under the current rule. That spec change isn't ready for v2, so it's left out here. The ordered slices are the piece it needs, so it can land on top later without another API break.

Ordering was the underlying problem. Additive options made the preference order implicit in registration order, and the most recently registered compressor was the most preferred, which reads backwards. `WithCompressors` and `WithCodecs` now replace the defaults, and the order you provide in the slice is exactly what gets advertised. It applies to both codecs and compressors for consistency and matches the pattern in [connect-py](https://github.com/connectrpc/connect-py).

This simplifies the option API surface. Removes `WithAcceptCompression`, registered compressors are now always advertised, and removes `WithNoCompression`, an empty `WithCompressors()`  disables compressors.

Migration tool has been updated to warn on `connect.WithCodec`, `WithCompression`, and `WithAcceptCompression` options. These aren't mechanical translations and require review. For example, the conformance tests set a custom JSON codec for a stricter encoding and has been updated.

### Examples 

Compression preference in order:
```go
// Prefer brotli, fall back to gzip. Sends Accept-Encoding: br,gzip
connecthttp.WithCompressors(brotli, connectgzip.New())
```

Removing the default JSON codec was previously impossible:
```go
// Serve binary Protobuf only.
connecthttp.Mount(mux, srv, connecthttp.WithCodecs(connectproto.NewBinaryCodec()))
```

Client prefer the first registered codec, so looses the need to always set these two options:
```go
// before
connecthttp.WithCodec(jsonCodec), connecthttp.WithSendCodec(jsonCodec.Name())
// after
connecthttp.WithCodecs(jsonCodec)
```

### DIFF
--- a/connecthttp/option.go
+++ b/connecthttp/option.go
@@ -51,73 +51,62 @@ func WithProtoJSON() Option {
 	return WithSendCodec(connect.CodecNameJSON)
 }
 
-// WithSendCodec selects the codec (by name) used for outgoing client requests.
-// The codec must also be registered via [WithCodec]. Defaults to
-// [connect.CodecNameProto]. [Mount] ignores this option.
+// WithSendCodec configures a client to use the specified codec by name for
+// outgoing requests. The named codec must be registered via [WithCodecs].
+//
+// If this option is not used, clients default to using the first codec
+// provided to [WithCodecs]. [Mount] ignore this option.
 func WithSendCodec(name string) Option {
 	return sendCodecOption(name)
 }
 
-// WithCodec registers a [connect.Codec]. Clients may use it for outgoing
-// requests. Servers accept it on inbound requests. Multiple WithCodec options
-// merge. Defaults to
-// [connectrpc.com/connect/v2/connectproto.BinaryCodec] and
-// [connectrpc.com/connect/v2/connectproto.JSONCodec].
-func WithCodec(codec connect.Codec) Option {
-	return codecsOption([]connect.Codec{codec})
+// WithCodecs configures the supported codecs for a client or handler. It
+// replaces any previously configured codecs, including the defaults.
+//
+// Handlers match incoming requests to codecs by name based on the content
+// type, so order only matters for clients. By default, a client sends
+// requests using the first codec in this list unless [WithSendCodec]
+// specifies otherwise. If multiple codecs share the same name, only the
+// first is kept.
+//
+// By default, the Protobuf binary and JSON codecs from
+// [connectrpc.com/connect/v2/connectproto] are registered, with binary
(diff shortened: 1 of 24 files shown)


connecthttp/option.go
```diff
@@ -51,73 +51,62 @@ func WithProtoJSON() Option {
...
 }
 
-// WithSendCodec selects the codec (by name) used for outgoing client requests.
-// The codec must also be registered via [WithCodec]. Defaults to
-// [connect.CodecNameProto]. [Mount] ignores this option.
+// WithSendCodec configures a client to use the specified codec by name for
+// outgoing requests. The named codec must be registered via [WithCodecs].
+//
+// If this option is not used, clients default to using the first codec
+// provided to [WithCodecs]. [Mount] ignore this option.
```
pkwarren:
```suggestion
// provided to [WithCodecs]. [Mount] ignores this option.
```

docs/v2-guide.md
```diff
@@ -384,11 +384,16 @@ connecthttp.WithSendMaxBytes(2048)
...
+connecthttp.WithCodecs(connectproto.NewBinaryCodec(), connectproto.NewJSONCodec())
+connecthttp.WithCompressors(connectgzip.New())
 ```
 
-By default `connecthttp` registers the protobuf binary and JSON codecs from `connectproto` and sends with the binary codec, so a transport built with no options is ready to use.
+By default, `connecthttp` is ready to use without configuration. It registers the gzip compressor alongside the Protobuf binary and JSON codecs from `connectproto`, defaulting to binary for outgoing requests.
+
+When customizing these, `WithCodecs` and `WithCompressors` replace the default sets entirely, so you must list everything you need in a single call. Order matters for both, but in different ways:
+
+* **Compressors:** Listed in order of preference (most preferred first). Clients and handlers advertise this order in `Accept-Encoding`, and a handler responds using the first encoding the client listed that it also supports.
```
pkwarren: Should we mention `WithSendCompression` here since we mention `WithSendCodec` below?

pkwarren (approved):

PR #971, opened 2026-09-05
fix: do not apply sendMaxBytes to EndStream control frames

### LINKED ISSUE #907: sendMaxBytes incorrectly applied to EndStream control frames in streaming RPCs

Tyagiquamar:
## Problem
`envelopeWriter.Write()` applies `sendMaxBytes` to all envelopes, including `connectFlagEnvelopeEndStream` control frames. When a streaming message exceeds `sendMaxBytes`, `stream.Send()` returns `CodeResourceExhausted`. However, when `connect-go` attempts to write the error into an EndStream frame via `MarshalEndStream`, the serialized error JSON itself may exceed `sendMaxBytes`, causing the EndStream frame to be silently dropped. Consequently, the client receives an empty HTTP 200 response without error details.

## Root Cause
`envelopeWriter.Write()` checked `w.sendMaxBytes > 0 && env.Data.Len() > w.sendMaxBytes` without checking whether the envelope is an `EndStream` control frame (`connectFlagEnvelopeEndStream`).

## Fix
In `envelopeWriter.Write()`, if `env.IsSet(connectFlagEnvelopeEndStream)` is true, write the envelope directly (`return w.write(env)`), bypassing `sendMaxBytes` limit checks and compression. Added unit test `TestEnvelopeWriterEndStreamExceedsSendMaxBytes` in `envelope_test.go`.

Fixes #907

### DIFF
--- a/envelope.go
+++ b/envelope.go
@@ -156,6 +156,9 @@ func (w *envelopeWriter) Marshal(message any) *Error {
 // Write writes the enveloped message, compressing as necessary. It doesn't
 // retain any references to the supplied envelope or its underlying data.
 func (w *envelopeWriter) Write(env *envelope) *Error {
+	if env.IsSet(connectFlagEnvelopeEndStream) {
+		return w.write(env)
+	}
 	if env.IsSet(flagEnvelopeCompressed) ||
 		w.compressionPool == nil ||
 		env.Data.Len() < w.compressMinBytes {
--- a/envelope_test.go
+++ b/envelope_test.go
@@ -118,3 +118,25 @@ func (b byteByByteReader) Read(data []byte) (int, error) {
 	data[0] = next
 	return 1, nil
 }
+
+func TestEnvelopeWriterEndStreamExceedsSendMaxBytes(t *testing.T) {
+	t.Parallel()
+	dst := &bytes.Buffer{}
+	wtr := envelopeWriter{
+		sender:       writeSender{writer: dst},
+		sendMaxBytes: 10,
+	}
+	largePayload := bytes.Repeat([]byte("a"), 100)
+	normalEnv := &envelope{Data: bytes.NewBuffer(largePayload)}
+	err := wtr.Write(normalEnv)
+	assert.NotNil(t, err)
+	assert.Equal(t, err.Code(), CodeResourceExhausted)
+
+	endStreamEnv := &envelope{
+		Data:  bytes.NewBuffer(largePayload),
+		Flags: connectFlagEnvelopeEndStream,
+	}
+	err = wtr.Write(endStreamEnv)
+	assert.Nil(t, err)
+}
+


emcfarlane:
Hi @Tyagiquamar, see this https://github.com/connectrpc/connect-go/pull/908#issuecomment-3967325804 for the approach.

Tyagiquamar:
Thanks for pointing me to the guidance on #908, @emcfarlane.

I've replaced the unconditional EndStream bypass with the two-stage fallback you described:

1. `sendMaxBytes` is enforced on all envelopes, including EndStream control frames.
2. If the full EndStream exceeds the limit, `MarshalEndStream` strips metadata/details and sends a minimal error frame indicating the end-stream message exceeded `sendMaxBytes`, while preserving the original error code.
3. If even that minimal frame exceeds the limit, it is force-sent as a last resort.

The same pattern is applied to gRPC-Web trailer frames via `MarshalWebTrailers`. The fallback logic lives in `writeControlFrame` (generic retry/force-send) with protocol-specific `marshal`/`reduce` closures in the Connect and gRPC-Web marshalers.

Added unit tests for enforcement and fallback behavior, plus a `CountUp` streaming regression test that reproduces #907 at the protocol boundary (connect and gRPC-Web, compressed and uncompressed). All tests pass (`go test ./...`, `go vet ./...`).

emcfarlane:
Thanks for the PR, and sorry for the runaround on the rework. Closing this in favor of #908, which was open first and does the same thing, so the credit should go to the original author.

PR #982, opened 2026-09-17
Reject a negative Connect-Timeout-Ms instead of expiring the context

NotAFlightRisk:
The Connect handler already rejects a timeout header that's non-numeric or longer than 10 characters, but there's no guard after `strconv.ParseInt` for a negative one. `Connect-Timeout-Ms: -1` therefore reaches `context.WithTimeout`, which builds a context that's already past its deadline.

The symptom depends on the handler. Against `main` at `a807df58` the stock ping handler still answers `200 {"number":"42"}` for `-1`, because it never looks at the context. A handler that does look - a database call, an outbound request - fails on that first call, so the caller gets an error out of the services own code instead of a `400 invalid_argument` naming the bad header.

Looks like a missed guard rather than intended behaviour. Three parsers doing the same job all reject a negative:

1. `connecthttp/protocol_grpc.go` - `grpcParseTimeout` checks `if err != nil || num < 0`.
2. connect-es - `packages/connect/src/protocol-connect/parse-timeout.ts` matches `/^\d{1,10}$/` before `parseInt`.
3. The conformance reference server - `internal/app/referenceserver/checks.go` checks `if err != nil || intVal < 0`.

`TestInvalidHeaderTimeout` already covers `10s` and `12345678901`, so the negative was the gap. The new subtest gets 200 without the guard, 400 with it.

`-0` and `+5` still get through, because `ParseInt` takes them and neither is negative. Thats what `grpcParseTimeout` and the reference server do too. Enforcing the spec's literal "1 to 10 digits" would also reject `0`, which connect-es tests as valid, so tighter grammar is a seperate conversation.

The client now gets `400 {"code":"invalid_argument","message":"parse timeout: \"-1\" is negative"}`.


emcfarlane (approved):
Thanks, it does look like a missed guard. The timeout should be 0 or greater https://connectrpc.com/docs/protocol/#unary-request-response-rpcs

